Issue #18, opened 2019-08-24
Book Styling - Print


hollasch:
Create CSS styling for the print variant of the book. This should yield a good solution for PDF book generation.

This is not included in the  v2 - web release.

hollasch:
Fixed in PR #177

PR #177, opened 2019-09-05
book print styling

hollasch:
Added @media print CSS queries for styling the printed version of the
book from the rendered HTML.

This change includes markup of the book content to hint which chunks
belong together on the same page. Look for `<div class='together'>`.

This includes minor fixes to book content.


hollasch:
Resolves #18

PR #181, opened 2019-09-07
Fix diffuse (should use random point on unit sphere, not inside it)

aras-p:
Should fix issue #145, see also https://aras-p.info/blog/2018/03/31/Daily-Pathtracer-Part-4-Fixes-Mitsuba/

Issue #179, opened 2019-09-05
Images: Images in books have artifacts


trevordblack:
Many of the pictures in the html files have obvious signs of artifacts.

There are compression artifacts, maximization artifacts.

Near as I can tell, none of these artifacts produce a wrong-looking or misleading image, but the image quality is occasionally noticeably poor.

hollasch:
Agreed. I've always planned on revisiting the images, but never got around to creating an issue; thanks.

Need to determine desired general image resolution. Some images (like surface acne) might show better as blown-up PNGs.

trevordblack:
We are aware of the problem.
This (#179). #155, and #145 are all probably going to make the next release (2.1.0 or 2.0.1).
We are in no position to merge anything for the next few weeks.

hollasch:
Aras P fixed most of these issues for book1. See PR #230.

trevordblack:
The images should probably be generated at some large texture (e.g. 1000x500, 2000x1000) and some high sample count (e.g. 100+). The src outlined in the book has people printing at lower resolutions and sample counts. Might need to add a slight blurb explaining the increased fidelity.

hollasch:
Current plan for rendered images:

1. Initial PPM image will be 256×256 for easier reading of the output PPM file.
2. General render size to be 384×216, up from 200×100. Still tolerable for performance, easier to view the result. It also yields a 16:9 aspect ratio instead of the previous 2:1.
3. Image parameters now work with an explicit aspect ratio, which drives the vertical resolution. This makes it easier for readers to quickly change the output image size according to their immediate needs.
4. The virtual image window will be computed automatically, without a bunch of hard-coded arithmetic values for various components.

hollasch:
Also note that we want to present actual renders (rather than high-quality ones), so the reader can compare their results against our resulting images.

hollasch:
This has forked out into individual issues.

PR #182, opened 2019-09-07
Update Book 1 screenshot images without severe JPG artifacts, and with fixed diffuse

aras-p:
As per #179 - the screenshots are really low quality JPGs. I've regenerated all the screenshots in Book 1 as PNGs.

Ran the PNGs through pngquant+Zopfli to reduce file sizes, but as expected they are a bit larger. All the images add ~180 kilobytes to previous book size.

Also the fix for "incorrect" diffuse (#181) is reflected in the screenshots. Resulting AO effect is slightly less intense than before, but more correct.

Also in one place fixed source to compile and the color in the source to match the color of the sphere in the screenshot.

Images in Book 2 and 3 are "not too bad"; they are larger and JPG artifacts aren't as visible there. Also they are much slower to render, and I'm lazy... :)


hollasch:
Ok, v2.0.0 is out the door, now we can focus on v3. Changes in this PR will be reviewed for that target.

@aras-p — pull requests should be preceded by an existing issue, so we can plan around releases and assignments. Also, one PR ↔ one issue. This PR needs to be made against the latest version now, and should also be separated into individual PRs, with references to their associated issue. Let me know if you want to tackle this. If not, I'll rework this PR into individual ones, and would request your review.

aras-p:
I can rework this, but in a few days from now.

hollasch:
Now superseded by PR #230.

Issue #14, opened 2019-08-24
Sanity pass for code blocks.


hollasch:
- [x] In One Weekend
  * [x] chapter 1
  * [x] chapter 2
  * [x] chapter 3
  * [x] chapter 4
  * [x] chapter 5
  * [x] chapter 6
  * [x] chapter 7
  * [x] chapter 8
  * [x] chapter 9
  * [x] chapter 10
  * [x] chapter 11
  * [x] chapter 12
- [x] The Next Week
  * [x] chapter 1
  * [x] chapter 2
  * [x] chapter 3
  * [x] chapter 4
  * [x] chapter 5
  * [x] chapter 6
  * [x] chapter 7
  * [x] chapter 8
  * [x] chapter 9
- [x] The Rest Of Your Life
  * [x] chapter 1
  * [x] chapter 2
  * [x] chapter 3
  * [x] chapter 4
  * [x] chapter 5
  * [x] chapter 6
  * [x] chapter 7
  * [x] chapter 8
  * [x] chapter 9

trevordblack:
For instances of "float" and "double" and other types in the text, do we want them to be inlined, i.e.
`float`s and `double`s.

hollasch:
I tend to use “single precision” and “double precision”, or “32-bit floats” and “64-bit floats” in prose unless I'm talking specifically about the types. If the types are important, then yeah, generally I try to mark them up as code: `float`/`double`. I'm not super strict on this, as sometimes my prose looks like a clown car of type changes, so sometimes I'll drop the inline code style just so the text reads easier (not usually).

So generally, use code styling, but with an escape hatch when you need it.

lorihollasch:
General: should be hittable, not hitable

Book 1:
Ch 2
•	“top part of my vec3 class”:
o	Reorder include files to match src conventions
•	“lots of vector operations”:
o	dot: unclear (spaces needed?)
o	function signatures: inconsistent spacing, eg: ){ versus ) {
o	two spaces after e[x] in some methods
o	order methods consistently as + - * /
o	unbold unit_vector
•	“change our main”:
o	Reorder include files to match src conventions
o	Put spaces

Ch 3
•	“add antialiasing later”:
o	Reorder include files to match src conventions
o	Remove *world and depth parameters from color

Ch 5
•	“motion blur at some point”:
o	Put hit_record and { on same line
o	Make spacing in 2 temp calculations the same
•	“And a list of objects”:
[...]

PR #183, opened 2019-09-08
Add styling for code highlighting

### LINKED ISSUE #11: Need solution to highlighted code lines
One solution would be to enable line numbers in code blocks, and then have the text reference the appropriate lines.

Another solution would be to introduce some form of CSS/HTML markup to denote highlighted regions. The downside of this is that you couldn't just import code blocks into the book in the future (which I think I'd like to do for other reasons). Probably better to just leave the code pure.

For example, see the last code block in Ray Tracing: The Next Week, chapter 4 (this used code comments). See the first code block in Ray Tracing: The Rest of Your Life, chapter 7 (new chapter 8). This one uses background highlighting.

----

After discussion, we'll just use code comments to call out changes. We need to make these changes for all three books.

hollasch:
This adds styles for deleted lines and for highlighted (or new) lines.

Note that switching sub-styles in the middle of a code block will
swallow any leading blank line before, so in some cases you'll see two
blank lines in the tilde block where there would normally be one. This
will still render with a single blank line.

In addition to addressing existing cases, I've also found other cases
where highlighting illustrates incremental changes from one version to
the next, particularly in _Ray Tracing In One Weekend_ (book 1).

See issue #14
Resolves #11

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -320,10 +320,12 @@ Now we can change our main to use this:
         std::cout << "P3\n" << nx << " " << ny << "\n255\n";
         for (int j = ny-1; j >= 0; j--) {
             for (int i = 0; i < nx; i++) {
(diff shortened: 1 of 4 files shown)


hollasch:
Merging. Feel free to review/comment/object at any time.

PR #184, opened 2019-09-09
Fix typo: hitable should be hittable

hollasch:
Sweeping change, so should be done before v2 web release.

This is part of the sanity pass work outlined in issue #14.

PR #205, opened 2019-09-22
Fix hittable filename typo

bennetthardwick:
`InOneWeekend` fails to compile, complaining:
```
hittable_list.h: No such file or directory
```

Looks like there was some left over typos from #184.

Thanks!


trevordblack:
Add it to the changelog

bennetthardwick:
@trevordblack I've updated the change log to include this PR and #184. Perhaps you could update contributing guidelines so new contributors know to update this file with their changes?

trevordblack:
The "update the CHANGELOG" was a comment for @ronaldfw

This is an opinion we might change in the future, but at present, we don't
want users to update the guides (CHANGELOG, readmes)

bennetthardwick:
Ah okay no worries, I've reverted my changes.

ronaldfw:
@trevordblack Thanks for the reminder, will update the changelog once this is merged.

PR #221, opened 2019-10-19
Float to double

hollasch:
Finally, warning-free on Windows. We're now using `double` pretty much everywhere. Also, I prefer using `auto` when possible (the type is derived from the expression), as this makes it less work for readers to convert to `float` if they prefer.


hollasch:
GAH! Accidentally did a PR to master.

PR #222, opened 2019-10-19
Float to double

hollasch:
Finally, warning-free on Windows. We're now using double pretty much everywhere. Also, I prefer using auto when possible (the type is derived from the expression), as this makes it less work for readers to convert to float if they prefer.


src/InOneWeekend/material.h
```diff
@@ -19,17 +19,17 @@
 struct hit_record;
 
 
-float schlick(float cosine, float ref_idx) {
-    float r0 = (1-ref_idx) / (1+ref_idx);
+double schlick(double cosine, double ref_idx) {
+    double r0 = (1-ref_idx) / (1+ref_idx);
```
hollasch: Oops. Shoulda been `auto`.

hollasch: Fixed in subsequent commits. Also forgot the books!

hollasch:
Chucking this onto the development branch. Post-reviews welcome.

PR #230, opened 2019-10-21
Update book1 images, plus fixed diffuse

hollasch:
This is copied from PR #182, targeting the development branch. Original
commentary:

Ran the PNGs through pngquant+Zopfli to reduce file sizes, but as
expected they are a bit larger. All the images add ~180 kilobytes to
previous book size.

Also the fix for "incorrect" diffuse (#181) is reflected in the
screenshots. Resulting AO effect is slightly less intense than before,
but more correct.

Also in one place fixed source to compile and the color in the source to
match the color of the sphere in the screenshot.

Images in Book 2 and 3 are "not too bad"; they are larger and JPG
artifacts aren't as visible there. Also they are much slower to render,
and I'm lazy... :)


hollasch:
This is a hand copy of PR #182, targeting the development branch.

PR #231, opened 2019-10-22
Add degrees_to_radians() utility function

### LINKED ISSUE #217: Revisit opportunities for defined constants
There are many places where we use, for example, `180.0 / pi` to convert between radians and degrees, or `pi/4.0`. When sensible, replace these with constants or utility functions defined in `rtweekend.h`.

hollasch:
Resolves #217


hollasch:
Gah. Wrong target branch.

Issue #146, opened 2019-07-15
Add code commentary for sphere discriminant


hollasch:
The code for computing the ray-sphere intersection first appears incorrect, because a factor of two is introduced elsewhere. See commentary for issue RayTracing/raytracing.github.io#113.

Add code commentary to explain this easily-misunderstood line.

hollasch:
Resolved in #233

Issue #138, opened 2018-09-15
Confusing code over the two roots of ray-sphere intersection.


alexnardini:
The code from line 42 (inluded) until line 49 (included) in the "sphere.h" is not needed? if it does can you please explain why? ( It does look to me as just a repetition of what has been already calculated.)

```
        temp = (-b + sqrt(discriminant)) / a;
        if (temp < t_max && temp > t_min) {
            rec.t = temp;
            rec.p = r.point_at_parameter(rec.t);
            rec.normal = (rec.p - center) / radius;
            rec.mat_ptr = mat_ptr;
            return true;
}
```

hollasch:
Resolved in #233, development branch.

PR #233, opened 2019-10-24
Clarify sphere intersect + slight perf improvement

### LINKED ISSUE #113: incorrect discriminant formula
https://github.com/petershirley/raytracinginoneweekend/blob/0abfd4c3e0d6fcb932d087832706ac968675a72e/sphere.h#L32

`    float discriminant = b*b - a*c;`

should be

`    float discriminant = b*b - 4*a*c;`

hollasch:
This change updates both the books and the code.

The sphere intersection code used the fact that the b (in the quadratic
formula) had a factor of two in it. This yielded a simplification in the
formula that hinted at, but not explained. Further, the variable in the
new formula used the unfortunate name `b`, when it was really b/2.

This change adds a deriviation of the simplification, and changes the
code to use a more honest name, and also adds some very tiny efficiency
improvements.

Resolves #113


hollasch:
Oops, #113 is already closed. However, there are two other near dupes: #138 and #146.

PR #235, opened 2019-10-24
Replace instances of FLT_MAX with infinity

### LINKED ISSUE #216: Replace instances of FLT_MAX with infinity
In all cases (probably), FLT_MAX should really be `infinity`, defined in `rtweekend.h`.

hollasch:
IEEE infinity is your friend.

Resolves #216


hollasch:
Thanks Aras!

PR #252, opened 2019-11-06
Mem alloc simple

hollasch:
*DRAFT*
This is a sketch of revised memory management, using simple `new`/`delete` operators, and a scheme where objects and materials have a single clear owner that frees memory on destruction.

If a single material instance is used for more than one `hittable`, it will end up in a double-free on destruction. If we went with this approach, we would need to note this limitation in the text. We _could_ deploy additional machinery to guard against sharing (like doing our own ref counting), but that's going way off track, I think, particularly in the face of better mechanisms like `std::shared_ptr`.

### DIFF
--- a/src/InOneWeekend/hittable.h
+++ b/src/InOneWeekend/hittable.h
@@ -26,15 +26,7 @@ struct hit_record {
 
 class hittable {
     public:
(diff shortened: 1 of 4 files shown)


hollasch:
Abandoning.

PR #254, opened 2019-11-06
Mem alloc shared

hollasch:
DRAFT
This is a sketch of revised memory management, using std::shared_ptr, allowing material and object (hittable) instances to be shared.

This approach has the downside of being more strongly a C++ solution, which might be a stumbling block for readers who have only a passing familiarity with C++. It's also more verbose and requires a bit more setup. However, it's robust and safe to use in a variety of ways. Finally one argument against this is that readers familiar with modern C++ memory management are free to do this on their own, just as they're encourage to write the entire program themselves.

With this approach, we should include a brief section explaining how this works for those readers who might need a quick guide.

### DIFF
--- a/src/InOneWeekend/hittable.h
+++ b/src/InOneWeekend/hittable.h
@@ -16,17 +16,22 @@
 
 
 class material;
+class hittable;
+
+using material_ptr = std::shared_ptr<material>;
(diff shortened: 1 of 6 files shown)


hollasch:
Abandoning.

PR #255, opened 2019-11-06
Use Markdeep's new chapter indirection syntax

hollasch:
This allows you to refer to a chapter by its title, instead of by
number, which makes it easier to add/delete/move chapters without
worrying about maintaining proper chapter number references.

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -192,10 +192,10 @@ definitely want to stratify.
 One Dimensional MC Integration
 ====================================================================================================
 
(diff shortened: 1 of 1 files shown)


hollasch:
Morgan pointed out that there's an alternate chapter reference format. You could use
```
... we could have framed chapter [A Simple Monte Carlo Program] in an integral ...
```
which would generate
> ... we could have framed chapter 2 in an integral ...

with "2" hyperlinked to the chapter.

The alternate format looks like this:
```
Recall from the Generating Random Directions chapter that to sample ...
```
which generates
> Recall from the Generating Random Directions chapter that to sample ...

with "Generating Random Directions" hyperlinked.

Issue #155, opened 2019-08-27
Picking directions on the upper hemisphere


vchizhov:
This is more of a suggestion that should provide a better intuition for people learning ray tracing for the first time. As I see it, currently it is not quite clear why the scattering directions need to be picked on a unit sphere offset along the normal (even though that provides a cosine weighted hemisphere distribution) - the geometric and algebraic derivations are both non-trivial, and possibly non-intuitive (this is clearly illustrated by the fact that picking directions inside the ball yields a cos^3 distribution, while picking points on the sphere yields the desired cos distribution).
My suggestion is to directly generate points in or on the upper hemisphere, which is a lot more intuitive and can be directly related to the rendering equation and light scattering (even informally). This can be done without inverse transform sampling - by generating point in the unit ball (or on the unit sphere) as was already done, and reflecting all vectors in the lower hemisphere to be in the upper hemisphere, with respect to the normal (requires a dot check per vector). Note that this will induce a uniform distribution (so higher variance, and a normalization factor of 2pi and not pi), and will require a cos(theta) multiplication in the recursion (which will be a good point to introduce Lambert's law which will help motivate the name of the material too).

hollasch:
Thoughts, @petershirley ?

petershirley:
I've been pondering this and not converging.     Here are some issues:

1) "diffuse" does not need to be Lambertian
2) But Lambertian has some advantages
3) The methods of generating rays on the unit shell is indeed unintuitive

Position 1: the simplest way to generate Lambertian directions should be used
Position 2: the simplest way to generate *any* kind of diffuse rays should be used as VC suggests

I am pretty comfortable with both those arguments-- they both have a point.   Anyone have a strong opinion?

vchizhov:
I wouldn't say that what I propose is necessarily simpler to implement. Sampling points within the unit ball, normalizing those to get points on the unit sphere, and offsetting those along the scattering normal doesn't seem more complex, than sampling points in the unit ball, and selectively reflecting those around the scattering plane (using a dot product).

Theoretically, however, motivating the former approach is a lot harder, and possibly not very intuitive. Algebraically it requires notions of probability theory and knowledge on the properties of the Dirac delta: https://github.com/vchizhov/Derivations/blob/master/Probability%20density%20functions%20of%20the%20projected%20offset%20disk%2C%20circle%2C%20ball%2C%20and%20sphere.pdf

[...]

Issue #137, opened 2018-04-22
Chapter 7, 8, 9 floating point intersection error


vchizhov:
There is a small issue with the materials introduced in chapters 7, 8 and 9 (lambertian, metal, dielectric). Due to floating point error, it may happen that the intersection of a ray with some geometry (in this case a sphere) may be "on the wrong side" (let the ideal intersection be at t0, however due to precision error let one get t1: t1>t0, then `r_in.point_at_parameter(t1)` would be "on the wrong side"). That means that when we try to scatter a ray, it will start inside of the sphere resulting in self-intersection and a black pixel contribution in the final image (akin to shadow acne). This is valid for both lambertian and metal materials, it is especially obvious for perfect mirror materials (`fuzz==0.0`) when they are introduced in chapter 8 since we get black pixels where we're not supposed to. To mitigate this issue, one can define epsilon = 0.01 (depending on the scale of your geometry this value may have to be smaller or larger) and when returning the scattered ray, make it have origin not `rec.p`, but rather `rec.p + epsilon*normal`. Which leads me to another issue - if an object is intersected "from inside", the wrong normal is used for lambertian scattering and reflection (fortunately the right normal is used for refraction). To fix this, use this normal for lambertian scattering, reflection and refraction: `normal = dot(r_in,rec.normal)<0.0f? rec.normal : -rec.normal`. For refraction the scattered ray should rather start from `rec.p - epsilon*normal` (we want to be on the opposite side, to avoid self-intersection).
Another thing that may look like a mistake is: `schlick(cosine, ref_idx)` since based on the definition it should be `schlick(cosine, ni_over_nt)`. However, in this case it is lucky that `schlick(cosine, 1.0/ni_over_nt) == schlick(cosine, ni_over_nt)` since `((1-ni_over_nt)/(1+ni_over_nt))^2 == ((1-1/ni_over_nt)/(1+1/ni_over_nt))^2`.

trevordblack:
For clarity, this issue is just that the scattering ray should be moved form the surface `epsilon*+-Normal`?

vchizhov:
@trevordblack Not only. There's the issue of incorrect normals being used for metals and lambertian materials, especially for intersection on the inside. I have fixed those in one of the commits to my fork.

trevordblack:
Finally looped back around on this.

1. If you believe it is a good idea to add epsilon offsets, create a separate issue.
[...]

PR #258, opened 2019-11-13
Tdb/diffuse

hollasch:
Created PR from Trevor's branch to facilitate review + discussion.


petershirley:
Nice job!   I don't see any issues.

mrmcsoftware:
Maybe I'm missing something, but there seems to be a discrepancy:  in line 1318 it's random_in_hemisphere (IN) and on line 1337 it's random_on_hemisphere (ON).  It seems to me it should be on_hemisphere, but also that the in_hemisphere code should use random_on_unit_sphere instead (and thus changing name to random_on_hemisphere), though I don't know the history mentioned about the "first raytracing papers", so I could be wrong.

trevordblack:
Yep. I just now noticed the same thing.
I wasn't 100% with the `random_**_unit_sphere`

mrmcsoftware:
Thanks.  I just tried both random_in_hemisphere and random_on_hemisphere in the line 1337 role (in my own raytracer) - both produced the same results (as far as I could tell from less than ideal number of samples).  I guess this is because the normal (rec.normal) is not being added, so it doesn't matter if the random vector is normalized.  So, just need to make 1337 and 1318 match, the rest is dealer's choice, so to speak.

trevordblack:
Correct

books/RayTracingInOneWeekend.html
```diff
@@ -1068,6 +1068,8 @@
 ideal diffuse surfaces. (I used to do it as a lazy hack that approximates mathematically ideal
 Lambertian.)
 
+(Reader Vassillen Chizhov proved that the lazy hack is indeed just a lazy hack and is inaccurate. The correct representation of ideal Lambertian isn't much more work and is presented at the end of the chapter.)
```
hollasch: Wrap Markdeep lines to 100 chars/line.

books/RayTracingInOneWeekend.html
```diff
@@ -1073,29 +1073,29 @@
 the chapter.)
 
 <div class='together'>
-Pick a random point $s$ from the unit radius sphere that is tangent to the hitpoint, and send a ray
-from the hitpoint $\mathbf{p}$ to the random point $\mathbf{S}$. That sphere has center
-$\mathbf{p} + \vec{\mathbf{N}}$:
+Pick a random point s from the unit radius sphere that is tangent to the hitpoint, and send a ray
```
hollasch: Since `s` is a mathematical variable, use `$s$`.

hollasch: This wording is still confusing. The unit radius sphere is tangent to the _surface_, at the hit point `$s$`.

hollasch: “hit point” is two words. Fix here and elsewhere (five instances, I believe).

books/RayTracingInOneWeekend.html
```diff
@@ -1073,29 +1073,29 @@
...
 
 </div>
 
 <div class='together'>
-We also need a way to pick a random point in a unit radius sphere centered at the origin. One way to
-do that is to pick a random angle $a$ on the $\mathbf{Z}$ plane and a random $z$ coordinate in
-$[-1,+1]$ to construct a vector with some trigonometry:
+We also need a way to pick a random point in a unit radius sphere centered at the origin. We’ll use
+what is usually the easiest algorithm: a rejection method. First, we pick a random point in the unit
+cube where x, y, and z all range from -1 to +1. We reject this point and try again if the point is
```
hollasch: First, ~we~ pick a random point ...
~We~ Reject this point and try again if ...

books/RayTracingInOneWeekend.html
```diff
@@ -1073,29 +1073,29 @@
...
+outside the sphere. A do/while construct is perfect for that:
```
hollasch: do-while construct

books/RayTracingInOneWeekend.html
```diff
@@ -1232,10 +1232,44 @@
...
+                return vec3(0,0,0);
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
+            vec3 target = rec.p + rec.normal + random_on_unit_sphere();
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+            return 0.5 * ray_color(ray(rec.p, target - rec.p), world, depth-1);
+        }
+
+        vec3 unit_direction = unit_vector(r.direction());
+        auto t = 0.5*(unit_direction.y() + 1.0);
+        return (1.0-t)*vec3(1.0, 1.0, 1.0) + t*vec3(0.5, 0.7, 1.0);
```
hollasch: Might consider this more succinct line:
```
return (1-t)*vec3(1,1,1) + t*vec3(0.5, 0.7, 1.0);
```
Tempting to add a `lerp()` utility function for lines like this.

books/RayTracingInOneWeekend.html
```diff
@@ -1230,7 +1230,7 @@
 The rejection method presented here produces random points in the unit ball offset along the surface
 normal. This corresponds to picking directions on the hemisphere with high probability close to the
 normal, and a lower probability of scattering rays at grazing angles. The distribution present
-scales by the $cos^2 (\phi)$ where $\phi$ is the angle from the normal. This is useful since light
+scales by the $cos^3 (\phi)$ where $\phi$ is the angle from the normal. This is useful since light
```
hollasch: `$\cos^3(\phi)$`

books/RayTracingInOneWeekend.html
```diff
@@ -1229,10 +1229,16 @@
 <div class='together'>
 The rejection method presented here produces random points in the unit ball offset along the surface
 normal. This corresponds to picking directions on the hemisphere with high probability close to the
-normal, and a lower probability of scattering rays at grazing angles. The distribution present scales by the $cos^2 (\phi)$ where $\phi$ is the angle from the normal. This is useful since light arriving at shallow angles spreads over a larger area, and thus has a lower contribution to the final color.
+normal, and a lower probability of scattering rays at grazing angles. The distribution present
+scales by the $cos^2 (\phi)$ where $\phi$ is the angle from the normal. This is useful since light
```
hollasch: `$\cos^2(\phi)$` — trig functions are special terms in LaTeX, so get a backslash prefix.

books/RayTracingInOneWeekend.html
```diff
@@ -1229,10 +1229,16 @@
...
+arriving at shallow angles spreads over a larger area, and thus has a lower contribution to the
+final color.
 
 However, we are interested in a Lambertian distribution, which has a distribution of $cos (\phi)$.
```
hollasch: `$\cos(\phi)$`

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+    [Listing [ray-color-unit-sphere]: <kbd>[main.cc]</kbd> ray_color() with replacement diffuse]
 </div>
 
 <div class='together'>
 After rendering we get a similar image:
 
-  ![Figure [correct-lambertian-sphere]: Correct rendering of a diffuse sphere](../images/img.correct-lambertian.jpg)
+  ![Correct rendering of Lambertian spheres]: ](../images/img.correct-lambertian.png)
 
 It's hard to tell what the difference between these two diffuse methods are, given that our scene of
```
hollasch: It's hard to tell ~what~ the difference between these two diffuse methods ~are~, given ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
-two spheres is so simple. You should be able to notice two important visual differences:
+two spheres is so simple. But, you should be able to notice two important visual differences:
```
hollasch: ... so simple, but you should ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+
+  1. The shadows are less pronounced after the change
+  2. Both spheres are lighter in appearance after the change
+
+Both of these changes are due to the more even scattering of the light rays, less rays are
```
hollasch: , ~less~ fewer rays are ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+scattering toward the normal. This means that for diffuse objects they will appear _lighter_
```
hollasch: ... for diffuse objects, they ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+because more light bounces toward the camera. For the shadows, less of the light bounces
```
hollasch: ... less ~of the~ light bounces ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+straight-up, so the parts of the larger sphere directly underneath the smaller sphere are brighter.
+</div>
+
+<div class='together'>
+The initial hack presented in this book went a long time before it was proven to be an incorrect
```
hollasch: ... in this book ~went~ lasted a long time ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+approximation of ideal Lambertian diffuse. A big reason that the error persisted for so long is
+that can be difficult to:
+
+  1. Mathematically prove that the probability distribution is incorrect
+  2. Intuitively explain why a $cos (\phi)$ distribution is desirable (and what it looks like it)
```
hollasch: `$\cos(\phi)$`

hollasch: (and what it ~looks like it~ would look like).

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+
+Not a lot of common, everyday objects are perfectly diffuse, so our visual intuition of how these
+objects behave under light can be poorly formed.
+</div>
+
+<div class='together'>
+In the interest of learning, we are including an intuitive and easy to understand diffuse method.
+For the two methods above we had a random vector, first of random length and then of unit length,
+offset from the hitpoint by the normal. It may not be immediately obvious why the vectors should be
```
hollasch: ~hitpoint~ hit point

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+displaced by the normal.
+
+A more intuitive approach is to have an equal probability of scattering in any direction. The
```
hollasch: Perhaps “... a uniform scatter direction.”?

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+distribution is truly uniform for all angles away from the hitpoint, with no dependence on the
```
hollasch: ~hitpoint~ hit point

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+angle from the normal. Many of the first raytracing papers used this diffuse method (before
+adopting Lambertian diffuse).
+
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+    vec3 random_in_hemisphere(const vec3& normal) {
+        vec3 in_unit_sphere = random_in_unit_sphere();
+        if(dot(in_unit_sphere, normal) > 0.0) // In the same hemisphere as the normal
```
hollasch: Space after `if`.

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [ray-color-hemisphere]: <kbd>[main.cc]</kbd> ray_color() with hemispherical scattering]
 
+Gives us the following image:
+
+  ![Rendering of diffuse spheres with hemispherical scattering](../images/fig.rand-hemispherical.png)
+</div>
+
+<div class='together'>
+The rendering scenes will become more complicated over the course of the book. You are encouraged to
```
hollasch: ~The rendering~ Scenes will become ...

books/RayTracingInOneWeekend.html
```diff
@@ -1265,20 +1271,91 @@
...
+switch between the different diffuse renderers presented here. Most scenes of interest will contain
+a disproportionate amount of diffuse materials. Understanding how changing your diffuse method
+changes your scene lighting is a valuable insight to gain.
```
hollasch: ~Understanding how changing your diffuse method changes your scene lighting is a valuable insight to gain.~
You can gain valuable insight by understanding the effect of different diffuse methods on the lighting of the scene.

books/RayTracingInOneWeekend.html
hollasch: Actually, since there are _two_ spheres tangent to the surface, we should further clarify that we want the one on the side that contains the ray origin → sign(Rd ⋅ N) = sign((p - Sc) ⋅ N)

A better diagram would be immensly useful here. I've created #263 to capture this.

trevordblack:
@aras-p This pull request is in response to the discussion had on #155

books/RayTracingInOneWeekend.html
trevordblack: There are exactly two unit spheres tangent to the surface at hit point $p$ but there are an infinite number of unit spheres tangent to the hit point alone. Wording will need be massaged.

aras-p:
> @aras-p This pull request is in response to the discussion had on #155

Right, I guess my point is that parts of it are redundant or incorrect by now -- e.g. the "diffuse images will be different between this & that approach", since the images have been fixed to be "actually correct" since the initial book release.

The whole idea of "point inside the sphere" vs "point on the sphere" and the discussion around it is useful to have, but the "the book images have point inside the sphere which is slightly wrong" bit is incorrect by now.

trevordblack:
Within a margin of error, all of the images in the book will need to be regenerated for the v3 release.
They'll need the correct diffuse, higher sample count, and made LARGE.

The change in diffuse will need to be propagated to the latter books. Someone (probably me) will need to parse through RestOfYourLife and qualify how the new diffuse changes the code there. #155 has a decent summary of those changes needed.

So at present, I'm not too worried about image discrepancies.

books/RayTracingInOneWeekend.html
```diff
@@ -1073,18 +1073,19 @@
 the chapter.)
 
 <div class='together'>
-Pick a random point s from the unit radius sphere that is tangent to the hitpoint, and send a ray
-from the hitpoint $p$ to the random point $s$. That sphere has center $(p + N)$:
+There are two unit radius spheres tangent to the hit point $p$ of a surface. These two spheres
+have a center of $(p + N)$ and $(p - N)$, where $N$ is the normal of the surface. The sphere with a
+center at $(p - N)$ is considered _inside_ the surface, whereas the sphere with center $(p + N)$ is considered _outside_ the surface. Pick a random point $s$ from the unit radius sphere outside the surface and send a ray from the hit point $p$ to the random point $s$:
```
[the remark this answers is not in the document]
trevordblack: _RIGHT._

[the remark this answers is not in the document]
trevordblack: I wasn't abundantly conscious in this passage cause I figured it would be rewritten in #137.
But, I completely forget to be _right_ in the first place.

Issue #92, opened 2018-11-26
The perlin_interp function not returning correct values?


40163650:
When I followed the tutorial in the book, I found that just before the turbulence is added I encountered some issues with my scene not rendering correctly.

I expected it to look like this:
![screenshot_1_expected](https://user-images.githubusercontent.com/12065481/48987023-f3652e80-f112-11e8-82cf-0a1d10cc07e6.png)
What I actually got was this:
![screenshot_1_actual](https://user-images.githubusercontent.com/12065481/48987032-00821d80-f113-11e8-9ecb-a9ae5708ae7f.png)

After continuing the tutorial to the turbulence section, I noticed that fabs() was used on the accumulated total. I wondered if this would fix my issue with the perlin_interp function (i.e. noise.noise() not noise.turb()), so I used it at the end there.

It did get rid of the black boxes but the result was also still not quite what was intended:
![screenshot_1](https://user-images.githubusercontent.com/12065481/48987116-d3823a80-f113-11e8-9456-ddf266fe197a.png)

Perhaps this is because I used a scale value of 1, but setting it higher leads to results more similar to using turbulence.

hollasch:
Fixed in #261, development branch.

PR #261, opened 2019-11-15
Fixed the perlin_interp bug. Cleaned up src

trevordblack:
There are 2 solutions presented in #92 and both solve the underlying issue.

The problem is such:

1. The `perlin_interp` function will produce negative values
2. These negative values become the luminance for the surface
3. Negative luminance isn't phy realizable, AND are getting passed to the `sqrt()` of the gamma
4. The outputs are garbled.

However, an understanding implicit in the writing of the source in the text is not made explicit.
Specifically, negative perlin values is acceptable because the `turb` value sums over multiple `perlin_interp` calls, and then takes the absolute value.

So the output of `turb` is always positive.

Both of the solutions present in #92 solve the issue of `NaN`s when caling `perlin_interp` alone, but remove most of the visual complexity of a `turb` call due to removing the negatives.

The only way to remove the `NaN`s and maintain visual complexity in `turb` is to cast perlin values back to 0...1 in the `noise` function

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -1326,7 +1326,9 @@ And the interpolation becomes a bit more complicated:
 </div>
 
 <div class='together'>
-The output of the perlin interpretation can return negative values. Negative light isn't physically realizable (and doesn't make a lot of sense). Plus, negative values of light will be passed to the `sqrt()` function of our gamma function and get turned into `NaN`s. We will cast the perlin output back to between 0 and 1.
+The output of the perlin interpretation can return negative values. These negative values will be
+passed to the `sqrt()` function of our gamma function and get turned into `NaN`s. We will cast the
+perlin output back to between 0 and 1.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class noise_texture : public texture {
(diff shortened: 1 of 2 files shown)


src/TheNextWeek/texture.h
```diff
@@ -58,7 +58,8 @@ class noise_texture : public texture {
         virtual vec3 value(double u, double v, const vec3& p) const {
             // return vec3(1,1,1)*0.5*(1 + noise.turb(scale * p));
             // return vec3(1,1,1)*noise.turb(scale * p);
-            return vec3(1,1,1)*0.5*(1 + sin(scale*p.x() + 5*noise.turb(scale*p)));
+            return vec3(1,1,1)*
+                0.5 * (1 + sin(scale*p.z() + 10*noise.turb(p)));
```
hollasch: Now that we've standardized on 96-characters per line, you can put this on a single line.

books/RayTracingTheNextWeek.html
```diff
@@ -1325,6 +1325,26 @@
     [Listing [perlin-interp]: <kbd>[perlin.h]</kbd> Perlin interpolation function so far]
 </div>
 
+<div class='together'>
+The output of the perlin interpretation can return negative values. Negative light isn't physically realizable (and doesn't make a lot of sense). Plus, negative values of light will be passed to the `sqrt()` function of our gamma function and get turned into `NaN`s. We will cast the perlin output back to between 0 and 1.
```
hollasch: Wrap Markdeep to 100 chars/line. Also, darkbulbs are awesome and actually useful, but yes, a problem here due to the square root.

PR #273, opened 2019-12-01
Hollasch/update guidelines

hollasch:
This PR targets master. On pull, we'll merge to development as well.


hollasch:
Fumbled. Will re-create.

Issue #97, opened 2019-07-25
Change to BVH Hit code


trevordblack:
Following Peter Shirley's blog post in March,
http://psgraphics.blogspot.com/2019/03/making-your-bvh-faster.html
Part 2
If the picture is a dead link:
https://pbs.twimg.com/media/D1bKEwiVYAARNl4.png:large

it is worth changing the BVH code to

```
bool hit(const ray& r, float tmin, float tmax, hit_record& rec) const
{
  if(box.hit(r, tmin, tmax, rec)) {
    if(left->hit(r, tmin, tmax, rec)) {
      right->hit(r, tmin, rec.t, rec);
      return true;
    } else {
      return right->hit(r, tmin, tmax, rec);
    }
  }
  return false;
}
```

I haven't personally experimented with the timings of the implementation, so I can't verify the following, but
According to the post it is both simpler and faster.
In general, for learning texts, changes for speed are rarely warranted, but I believe that changing the BVH code here would be simpler for the student to understand and reason about.

I am willing to make the change myself.

Trevor  David Black

trevordblack:
I wasn't seeing a speedup from this on the Cornell box. This shouldn't be too surprising. The cornell box has very few primitives.

I went to go render the scene from the end of TheNextWeek and got an out-of-memory pointer error.

So, after I figure that out, I can figure out if this is a performance speedup.

If it isn't, we might need to seriously investigate if this makes sense.

Also, slight correction for the code:

```
// tmax and tmin become t_max and t_min
// line 3: removal of rec in function call
bool hit(const ray& r, float t_min, float t_max, hit_record& rec) const
{
  if(box.hit(r, t_min, t_max)) {
    if(left->hit(r, t_min, t_max, rec)) {
      right->hit(r, t_min, rec.t, rec);
      return true;
    } else {
      return right->hit(r, t_min, t_max, rec);
    }
  }
  return false;
}
```

trevordblack:
Throws out-of-memory pointer cause it looks for the earthmap.jpg texture in the current directory. Might need to change that to `../images/earthmap.jpg`.

Anyway, got it working, can investigate BVH speedup

trevordblack:
Well _Shit_
Without bvh change:
```
lucifer@morningstar:~/writing/raytracing.github.io/build$ time ./theNextWeek > restweek2.ppm
Scanlines remaining: 0   
Done.

real	12m8.605s
user	12m8.487s
sys	0m0.028s
```

[...]

PR #288, opened 2019-12-03
Tdb/bvh hit change

trevordblack:
See issue #97


hollasch:
Delete and redo into development instead of master. If I had a nickel for every time I've done this…

PR #291, opened 2019-12-03
Changed to simpler bvh hit code. Resolves #97

### LINKED ISSUE #97: Change to BVH Hit code

trevordblack:
**sigh**

Pushed this to be merged with `master` instead of `development`.

Correcting that here.

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -714,15 +714,13 @@ check the children and sort out any details:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     bool bvh_node::hit(const ray& r, double t_min, double t_max, hit_record& rec) const {
-        if(box.hit(r, t_min, t_max)) {
-            if(left->hit(r, t_min, t_max, rec)) {
-                right->hit(r, t_min, rec.t, rec);
-                return true;
-            } else {
-                return right->hit(r, t_min, t_max, rec);
-            }
-        }
-        return false;
+        if (!box.hit(r, t_min, t_max))
+            return false;
+
+        bool hit_left = left->hit(r, t_min, t_max, rec);
+        bool hit_right = right->hit(r, t_min, hit_left ? rec.t : t_max, rec);
+
+        return hit_left || hit_right;
     }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     [Listing [bvh-hit]: <kbd>[bvh.h]</kbd> Bounding volume hierarchy hit function]
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -714,29 +714,15 @@ check the children and sort out any details:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     bool bvh_node::hit(const ray& r, double t_min, double t_max, hit_record& rec) const {
-        if (box.hit(r, t_min, t_max)) {
-            hit_record left_rec, right_rec;
-            bool hit_left = left->hit(r, t_min, t_max, left_rec);
-            bool hit_right = right->hit(r, t_min, t_max, right_rec);
-            if (hit_left && hit_right) {
-                if (left_rec.t < right_rec.t)
-                    rec = left_rec;
-                else
-                    rec = right_rec;
+        if(box.hit(r, t_min, t_max)) {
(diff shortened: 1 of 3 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -714,29 +714,15 @@
...
-                    rec = right_rec;
-                return true;
-            }
-            else if (hit_left) {
-                rec = left_rec;
-                return true;
-            }
-            else if (hit_right) {
-                rec = right_rec;
+        if(box.hit(r, t_min, t_max)) {
```
hollasch: Need space after `if`.

books/RayTracingTheNextWeek.html
```diff
@@ -714,29 +714,15 @@
...
+                right->hit(r, t_min, rec.t, rec);
                 return true;
+            } else {
+                return right->hit(r, t_min, t_max, rec);
             }
-            else
-                return false;
         }
-        else return false;
+        return false;
```
hollasch: Space after keywords (`if`).

Also, this can be simplified further:

    if (!box.hit(r, t_min, t_max)) return false;

    bool hit_left  = left->hit (r, t_min, t_max, rec);
    bool hit_right = right->hit(r, t_min, t_max, rec);

    return hit_left || hit_right;

src/TheNextWeek/bvh.h
```diff
@@ -34,30 +34,15 @@ bool bvh_node::bounding_box(double t0, double t1, aabb& output_box) const {
...
-        }
-        else if (hit_right) {
-            rec = right_rec;
-            return true;
-        }
-        else
-            return false;
     }
-    else return false;
+    return false;
```
hollasch: Similar comment as above.

src/TheRestOfYourLife/bvh.h
```diff
@@ -34,30 +34,15 @@ bool bvh_node::bounding_box(double t0, double t1, aabb& output_box) const {
...
-        }
-        else if (hit_right) {
-            rec = right_rec;
-            return true;
-        }
-        else
-            return false;
     }
-    else return false;
+    return false;
```
hollasch: Similar comment as above.

trevordblack:
_Not Exactly_.

the rec is incorrect with your solution. If both left and right

trevordblack:
thenextweek
thenextweek_wrong

trevordblack:
On further investigation, it would need to be something like:

```
if (!box.hit(r, t_min, t_max)) return false;

bool hit_left  = left->hit (r, t_min, t_max, rec);
bool hit_right = right->hit(r, t_min, (hit_left) ? rec.t : t_max, rec);

return hit_left || hit_right;
```

hollasch:
Well, that was me being sloppy. I had assumed that the `hit()` function was accounting for the `rec` values. Perhaps it should?

trevordblack:
I dug in a bit.

I think that `hit()` should _not_ account for `rec` values. It definitely could.

At present **lots** of objects inherit `hit()` from `hittable` and all of those would need to be rewritten to account for such a change.

To add to that, at present, the philosophy of the code is that all hittable objects need to
1. Figure out if they intersect
2. And if that intersect is between t_min and t_max

We would need to add
3. Also check against rec.

But, the big reason I don't think this is a great idea is that you should treat all `rec`s going into `hittable` as new objects. Anything that inherits `hittable` _should_ be able to assume that it's input record is entirely default.

If it isn't, _hoooooooo booy_. What can end up happening is that for the `hittable`s that don't diligently go through and overwrite all values in `rec` (which will be all of them) you can have state from other rays leak into this ray, or data from other `hittable`s. It's just a **bad** show.


We can make it so that `hittable` needs to account for `rec`'s values. But then you're opening yourself up to unimaginably horrific bugs (I've made this mistake before)

hollasch:
Ok, then next discussion. If the hit record is assumed to be fresh, why aren't we just returning it by value? Do we forbid callers from re-using the hit record? If not, then all hit methods must ensure that every value is set in the event of a hit. Or we could return a fresh hit record, instead of getting a hit record as an out parameter.

Probably overthinking it as this point. Let's go with your suggested change to my suggested change.

trevordblack:
why not

```
if(!box.hit(r, t_min, t_max)) return false;

if(left->hit(r, t_min, t_max, rec)) {
    right->hit(r, t_min, rec.t, rec);
    return true;
} 

return right->hit(r, t_min, t_max, rec);
```

hollasch:
That'll work.

hollasch:
Went with your earlier alternate solution.

PR #279, opened 2019-12-02
Tdb/the rest of your life edits

trevordblack:
As a means of procrastinating on other edits that I'm working on, I decided to read back through TheRestOfYourLife.

This is an omni-bus PR for all of the changes I found in chapters 1-6.
Ideally, lots of these would be within unique Issues and PRs, but I pushed for velocity over cleanliness.

Don't be kind.
I'll loop back around and create more PRs for other problems I perceive in the later chapters.

PR #293, opened 2019-12-05
Header sort

hollasch:
Lots of header simplification and reorganization.


trevordblack:
1. Move `ffmax` and `ffmin` somewhere else
2. Maybe move all `random_.*` into their own random.h header
3. Move surface_texture.h inside texture.h
4. Maybe collect all material.h into common

What's the cut-off for what makes it into common and what doesn't?
That should probably codified somewhere.

trevordblack:
As much as possible, I don't want this collation to infect the text.

hollasch:
Created https://github.com/RayTracing/raytracing.github.io/issues/294 for the random functions.

hollasch:
Created #295 for the `surface_texture.h` → `texture.h` move.

hollasch:
I looked at `material.h`, but that one's a hairball.

hollasch:
Holding on this PR to merge in pending PR #279 .

PR #297, opened 2019-12-09
Random

### LINKED ISSUE #285: Work on `vec3` class
There are many `vec3` utility functions that belong in `vec3.h`, but instead are defined elsewhere. For example, utilities that generate random points on or in the unit sphere. There are also a couple of random vec3 functions that are better located in the header, and more common patterns that could use small utility functions to simplify the code.

### LINKED ISSUE #294: Move random functions into their own header
They're scattered across various files.

hollasch:
Resolves #285 
Resolves #294


hollasch:
Superseded by #300.

PR #308, opened 2019-12-12
Fix issues with white-to-blue gradient code

### LINKED ISSUE #306: Chapter 4 Blending depending on up/downess of the y coordinate for InOneWeekend
In Chapter 4 of **In One Weekend**, it states `The ray_color(ray) function linearly blends white and blue depending on the up/downess of the y coordinate.`, but I was seeing more white near the center of the picture. I assumed it should be the same color for each row because it was described as a `simple gradient`. I realized the statement misleading because the length of the ray cast from the camera to the pixel changes as x changes which changes the normalized vector component y of the direction vector. A note in the text stating as such would have cleared this up, simply stating it is affected by the length of the ray cast and driven off the normalized y component.

hollasch:
Resolves #306


hollasch:
Nope. Nope nope nope NOPE.

I though this was just for a test image, but instead, it's the background gradient used for all images from that point on. Changing this means re-rendering all of the following images.

Abandoning this change.

PR #311, opened 2019-12-16
Tdb/normals and dielectrics

trevordblack:
How I envision normals and dielectrics will work going forward.
I was very diligent with InOneWeekend, but not thorough enough for the other two books.

There might be daemons I let through.

I'm at the point in writing where I absolutely hate what I've created.
Need cooler heads.

Don't be kind.


trevordblack:
Oop. Forgot to do aarect.h

trevordblack:
In TheNextWeek Chapter 7 [Rectangles and Lights], a class `flip_normals` is introduced to invert the normals for a hittable. This class is introduced because the naive solution for `**_rect` produces blackness instead of showing 3/5 rects.

My suspicion, and I'm not sure how to prove this, is that the code `lambertian::scatter`:

```
        virtual bool scatter(
            const ray& r_in, const hit_record& rec, vec3& attenuation, ray& scattered
        ) const {
            vec3 target = rec.p + rec.normal + random_unit_vector();
            scattered = ray(rec.p, target-rec.p, r_in.time());
            attenuation = albedo->value(rec.u, rec.v, rec.p);
            return true;
        }
```
does the following thing:
When a ray hits the backside of a lambertian, that ray is scattered. But because the normals point on the other side of the surface, the ray is transmitted through the surface before being scattered. So hitting the back side of a lambertian material acts diffuse transmission, rather than diffuse reflection.

I'm not entirely sure how to robustly test this, but that is my suspicion.

I discovered this when I realized that under the new normal system (where normals always point against the ray) all rects will be visible, _unless_ their normals are flipped.

So, it appears as if `flip_normal` erroneously forces diffuse transmission.
Under the new system, the de facto replacement of the `flip_normal` class is a `flip_face` class. The only mateial at present that accounts for `face`edness is the `dielectric` material, and we don't have a scene that combines aarect as glass.

If we want to motivate the introduction of `flip_face` we'll need to do that. We could change one of the cornell box rectangles to glass.

This has become a massive rewrite problem.

trevordblack:
I need to rewrite TheRestOfYourLife with the new system, both text and src.

It's not a lot of work, and I'll swing back around to it.

hollasch:
Lots of busy on my end — I will review when I can.

trevordblack:
Talk about "Foot Guns into Children's Toys"

I used `git rebase -i` to collapse the 20 commits into 1 commit. That tool is insane.

I opened another branch because I'm not sure if I did it correctly, and there's a nonzero chance I destroyed all of my changes.

I'm closing this PR, go to the other one.

hollasch:
Generally I use `git rebase -i` to edit time. This allows me to cluster ideas in logical chunks, and to fold intermediate (but ultimately unimportant) expressions into a clear change. I try to preserve the granularity of the change, keeping clear increments of the development.

Whole-effort squashing is _occasionally_ useful, but used aggressively destroys the history of logical progression, and that's the whole reason for keeping versions.

Regarding your concern about potential destruction, my practice is to start in a feature branch (say `frotzjam`). If I've already pushed it at some point, I do this:

    git switch -c frotzjam-2
    git rebase -i development

and then push `frotzjam-2` when I'm ready for the next review. That way I have all versions of my development until I'm ready to pick one and integrate it.

Issue #305, opened 2019-12-12
Implement a formal `scene` object


hollasch:
Instead of a static fixed array of object pointers, implement a simple `scene` class that manages the objects in the scene. In addition, this could include scene properties like background color and such.

hollasch:
From #158:

I was thinking that the scene would be the new hittable_list, but now I'm thinking that a scene manages the following:

A background color
A hittable_list of objects
A camera
A sample function, which takes u,v coordinates and maximum ray depth
Also, the hittable_list class should be refactored to use a std::vector, with the following methods:

clear() to clear the list of all objects.
add() to add a hittable* to the list.

trevordblack:
maximum ray depth meaning recursion depth? Not, t_max?

`sample` needs the camera, the list of objects, a `u`, a `v` and a recursion depth.

I don't love putting the `sample` function inside of the `scene` class. I would personally prefer the `scene` to be a bucket of dumb bits, as pure a `struct` as possible.

Do we want to move `time0` and `time1` out of `camera` and into `scene`?
Having the shutter times maintained somewhere with larger scope makes sense to me, but is more work and therefore I don't think is a good idea.

But, this works well for a new user.

trevordblack:
I think the most logical place to introduce these changes in the text is at the start of Chapter 13 [Where Next].

The chapter acts as coda, collecting all of the preceding chapters into one final scene.
It might make sense to actually collect all of the data objects into one class at the same time.

trevordblack:
The final scene in TheNextWeek has a lit up background, and, out of the box, has the wrong colors for the two textured spheres

mrmcsoftware:
I commented in #158 the following, but perhaps it wasn't seen because of #305 superseding? #158.  So, I'll comment here.

-------------

[...]

Issue #158, opened 2019-09-01
Many scenes in final code render as black image unless one changes the background color


ronaldfw:
I just noticed this when playing around with the code. There are several example scenes in the main.cc file that look like one can just uncomment them to render them. But when doing so, one just gets a black image. The reason is that these scenes do not have lights and the background color was set to black (0, 0, 0) at some point (Chapter 6 to be precise). They render fine (assuming good lookfrom / lookat calls) when the background color is set to an actual color.

I think the code is fine as is if it is supposed to represent "here's what you should have at the end of the book", but a couple things we could consider:
- Lighter change: Add comments to indicate that one needs to change background color and lookfrom/lookat calls.
- Heavier change: Background color (and arguably lookfrom/lookat arguments) should be part of the scene.

trevordblack:
Lighter change.

It's industry standard to have the background be an ambient light.

I'm of the opinion that we should either be simple for teaching purposes,
or industry standard.

Following that, we should either simplify, or model the background as a
child of the `light` class.

trevordblack:
I'm going to investigate (and possibly implement) what the changes would look like if we modeled the background as a child of the `light` class

hollasch:
Adding an explicit background color to a `scene` object would be a tidy solution to this issue. See #305.

hollasch:
There's another problem here: the random "world"s that are constructed are distinct from the camera. When you fix the background for these, you often get a view that doesn't make sense for that scene.

I was thinking that the `scene` would be the new `hittable_list`, but now I'm thinking that a `scene` manages the following:

- A background color
- A `hittable_list` of objects
- A camera
- A `sample` function, which takes u,v coordinates and maximum ray depth

Also, the `hittable_list` class should be refactored to use a `std::vector`, with the following methods:

- `clear()` to clear the list of all objects.
- `add()` to add a `hittable*` to the list.

mrmcsoftware:
[...]

Issue #147, opened 2019-07-24
General report of memory leaks in code. Review source for missing deletes.


hollasch:

hollasch:
See the `hittable-vector` branch for a sketch of how `hittable_list` might be improved with a `std::vector` of objects, rather than using a straight array.

trevordblack:
Yep.

This is absolutely a drop-in improvement over the existing system.
I lament the continued deprecation of raw arrays, but the entire c++ ecosystem is moving away from C.

However.
We're still leaking memory.
The materials are not being properly cleared out with this solution.

It's a great first step.

hollasch:
Right, there's a whole next phase in addressing this with the rest of the scene structure. Coding up a simple RAII approach with `new`/`delete` is simple and straight-forward, but unfortunately doesn't properly handle objects that are referenced from multiple places. For example, if materials are shared among several `hittable` instances. I _think_ (but have not verified) that the text employs no shared instances, but it's certainly reasonable to be able to do so.

So there are two different approaches:

1. Deploy the simple RAII `new`/`delete` approach, with warnings (and assertions?) that instances may not be shared.

2. Deploy a standard C++ `shared_ptr` solution. It's a bit more verbose, but has the additional advantage that it advertises the ability to share instances among several objects.

I believe I'm going to code up option (2) to see how it plays out. If we don't like it, we can try (1) and somehow try to address the shared reference problem.

trevordblack:
Found in Chapter 8, InOneWeekend:

> One question is whether we can mix and match shapes and materials (so we assign a sphere a material) or if it’s put together so the geometry and material are tightly bound (that could be useful for procedural objects where the geometry and material are linked). We’ll go with separate — which is usual in most renderers — but do be aware of the limitation.


We must not forget to amend this, whichever way we go

hollasch:
Peeled out the `scene` object work from this issue: #305.

hollasch:
Now that `hittable_list` uses a `std::vector`, migrating the codebase to use `std::shared_ptr` was fairly straight-forward, with only a few hiccups. See PR #324. This should resolve the issue of memory leaks.

hollasch:
Done.

PR #319, opened 2020-01-05
Hitvec.2

hollasch:
Sorry about this change; it's pretty much a squash of my changes, after chasing a bunch of dead ends. Should have been more incremental and disciplined. Still, I'm happy with the final results.

The primary change in this is that we're now using `hittable_list` with a `std::vector<hittable*>`, which alleviates a bunch of the awkward memory allocation. This opened up a number of opportunities to simplify code.

This version still leaks memory, but would moving to shared pointers from this base would be straight-forward.

Related/motivating issues: #305, #158, #147

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -617,8 +617,7 @@ $\vec{\mathbf{N}}$:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
         if (discriminant < 0) {
             return -1.0;
-        }
-        else {
+        } else {
             return (-b - sqrt(discriminant) ) / (2.0*a);
         }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
@@ -683,8 +682,7 @@ Using these observations, we can now simplify the sphere-intersection code to th
 
     if (discriminant < 0) {
         return -1.0;
-    }
-    else {
+    } else {
         return (-half_b - sqrt(discriminant) ) / a;
     }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -792,27 +790,35 @@ And a list of objects:
     #define HITTABLE_LIST_H
 
     #include "hittable.h"
+    #include <vector>
 
     class hittable_list: public hittable {
         public:
             hittable_list() {}
-            hittable_list(hittable **l, int n) {list = l; list_size = n; }
+            hittable_list(hittable* object) { add(object); }
+
+            void clear()               { objects.clear(); }
+            void add(hittable* object) { objects.push_back(object); }
+
             virtual bool hit(const ray& r, double tmin, double tmax, hit_record& rec) const;
-            hittable **list;
-            int list_size;
+
+        public:
+            std::vector<hittable*> objects;
     };
 
     bool hittable_list::hit(const ray& r, double t_min, double t_max, hit_record& rec) const {
         hit_record temp_rec;
         bool hit_anything = false;
-        double closest_so_far = t_max;
-        for (int i = 0; i < list_size; i++) {
-            if (list[i]->hit(r, t_min, closest_so_far, temp_rec)) {
+        auto closest_so_far = t_max;
+
+        for (auto object : objects) {
+            if (object->hit(r, t_min, closest_so_far, temp_rec)) {
                 hit_anything = true;
                 closest_so_far = temp_rec.t;
                 rec = temp_rec;
             }
         }
+
         return hit_anything;
     }
 
@@ -820,6 +826,11 @@ And a list of objects:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     [Listing [hittable-list-initial]: <kbd>[hittable_list.h]</kbd> The hittable_list class]
 </div>
+
+If you're unfamiliar with C++'s `std::vector`, this is a generic array-like collection of an
+arbitrary type. Above, we use a collection of pointers to `hittable`. `std::vector` automatically
+grows as more values are added: `objects.push_back(object)` adds a value to the end of the
+`std::vector` member variable `objects`.
 
 <div class='together'></div>
 We need some math constants that we conveniently define in their own header file. For now we only
@@ -876,10 +887,10 @@ And the new main:
     #include <iostream>
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-    vec3 ray_color(const ray& r, hittable *world) {
+    vec3 ray_color(const ray& r, hittable& world) {
         hit_record rec;
-        if (world->hit(r, 0.0, infinity, rec)) {
-            return 0.5*vec3(rec.normal.x()+1, rec.normal.y()+1, rec.normal.z()+1);
+        if (world.hit(r, 0.0, infinity, rec)) {
+            return 0.5 * (rec.normal + vec3(1,1,1));
         }
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
@@ -893,7 +904,9 @@ And the new main:
     int main() {
         int nx = 200;
         int ny = 100;
+
         std::cout << "P3\n" << nx << ' ' << ny << "\n255\n";
+
         vec3 lower_left_corner(-2.0, -1.0, -1.0);
         vec3 horizontal(4.0, 0.0, 0.0);
         vec3 vertical(0.0, 2.0, 0.0);
@@ -901,10 +914,9 @@ And the new main:
 
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-        hittable *list[2];
-        list[0] = new sphere(vec3(0,0,-1), 0.5);
-        list[1] = new sphere(vec3(0,-100.5,-1), 100);
-        hittable *world = new hittable_list(list,2);
+        hittable_list world;
+        world.add(new sphere(vec3(0,0,-1), 0.5));
+        world.add(new sphere(vec3(0,-100.5,-1), 100));
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
 
         for (int j = ny-1; j >= 0; --j) {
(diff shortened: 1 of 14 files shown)


src/TheNextWeek/main.cc
```diff
@@ -20,272 +20,307 @@
...
     hit_record rec;
-    if (depth <= 0 || !world->hit(r, 0.001, infinity, rec))
+
+    // If we've exceeded the ray bounce limit, no more light is gathered.
+    if (depth <= 0)
         return vec3(0,0,0);
 
+    // If the ray hits nothing, return the background color.
+    if (!world.hit(r, 0.001, infinity, rec))
+        return vec3(0.5,0.5,0.5);
```
trevordblack: I understand why this change exists (having grey background).
I just don't love it, and having a scene-specific background was a big reason to create a dedicated `scene` class (for me).

At least as an intermediate step, create a `scene_background` function that is `inline`d and returns vec3(0.5)

src/TheNextWeek/hittable_list.h
```diff
@@ -11,34 +11,56 @@
...
-    if (list_size < 1) return false;
+    if (objects.empty()) return false;
 
     aabb temp_box;
-    bool first_true = list[0]->bounding_box(t0, t1, temp_box);
+    bool first_true = objects[0]->bounding_box(t0, t1, temp_box);
 
     if (!first_true)
         return false;
     else
```
trevordblack: `else` is unnecessary here. Remove.

src/TheRestOfYourLife/hittable_list.h
```diff
@@ -11,50 +11,58 @@
...
-    if (list_size < 1) return false;
+    if (objects.empty()) return false;
 
     aabb temp_box;
-    bool first_true = list[0]->bounding_box(t0, t1, temp_box);
+    bool first_true = objects[0]->bounding_box(t0, t1, temp_box);
 
     if (!first_true)
         return false;
     else
```
trevordblack: `else` is unneccessary here.
Remove

src/TheRestOfYourLife/hittable_list.h
```diff
@@ -63,19 +71,23 @@ bool hittable_list::bounding_box(double t0, double t1, aabb& output_box) const {
...
+    for (auto object : objects)
+        sum += weight * object->pdf_value(o, v);
+
+    return sum;
 }
 
+
+vec3 hittable_list::random(const vec3 &o) const {
+    auto int_size = static_cast<int>(objects.size());
+    auto index = static_cast<size_t>(random_int(0, int_size-1));
```
trevordblack: why the cast back to `size_t`?

src/TheRestOfYourLife/main.cc
```diff
@@ -16,24 +16,31 @@
...
     hit_record rec;
-    if (depth <= 0 || !world->hit(r, 0.001, infinity, rec))
+
+    // If we've exceeded the ray bounce limit, no more light is gathered.
+    if (depth <= 0)
         return vec3(0,0,0);
 
+    // If the ray hits nothing, return the background color.
+    if (!world.hit(r, 0.001, infinity, rec))
+        return vec3(0,1,1);
```
trevordblack: Consider creating a `background_color` function

src/TheNextWeek/bvh.h
```diff
@@ -13,12 +13,20 @@
...
 #include "common/rtweekend.h"
 #include "hittable.h"
+#include <algorithm>
 
 
 class bvh_node : public hittable  {
     public:
-        bvh_node() {}
-        bvh_node(hittable **l, int n, double time0, double time1);
+        bvh_node() = delete;
```
trevordblack: That's going to be a Yikes from me.

I think that deleting the default constructor is a good idea.
I just hate the committee's choice of syntax.
I think it's best to gloss over this detail, and lead it to the user to grokk what's happening

src/TheNextWeek/bvh.h
```diff
@@ -13,12 +13,20 @@
...
+
+        bvh_node::bvh_node(hittable_list& list, double time0, double time1)
```
trevordblack: This threw an error for me in clang
`bvh_node::bvh_node` -> `bvh_node`

It's complaining about the scope

src/TheNextWeek/bvh.h
trevordblack: Ehhh, I guess the syntax is fine.

I guess the `= delete;` activated my "this is outside our usual c++ calls" lizard brain.

src/TheNextWeek/bvh.h
```diff
@@ -28,95 +36,77 @@ class bvh_node : public hittable  {
...
-}
-
-int box_y_compare (const void * a, const void * b)
-{
-    aabb box_left, box_right;
-    hittable *ah = *(hittable**)a;
-    hittable *bh = *(hittable**)b;
+bool box_x_compare (const hittable* a, const hittable* b) { return box_compare(a, b, 0); }
+bool box_y_compare (const hittable* a, const hittable* b) { return box_compare(a, b, 1); }
+bool box_z_compare (const hittable* a, const hittable* b) { return box_compare(a, b, 2); }
```
trevordblack: hops through 2 function redirections.

Consider inlining `box_compare`

src/TheNextWeek/bvh.h
```diff
@@ -28,95 +36,77 @@ class bvh_node : public hittable  {
...
-    if (box_left.min().y() - box_right.min().y() < 0.0)
-        return -1;
-    else
-        return 1;
-}
+bvh_node::bvh_node(
+    std::vector<hittable*>& objects, size_t start, size_t end, double time0, double time1
+) {
+    int axis = random_int(0,2);
+    auto comparator = (axis == 0) ? box_x_compare
```
trevordblack: Consider replacing with a switch-case or if-elseif

trevordblack: ... instead of the ternary command

src/TheNextWeek/bvh.h
trevordblack: ... Actually, don't consider it. _Just do it_

Considering that std::sort is a **significant** part of our runtime, unless the compiler is automatically inlining the function, there's a good 2-5% just on the floor here.

and the
```
if (!a->bounding_box(0,0, box_a) || !b->bounding_box(0,0, box_b))
        std::cerr << "No bounding box in bvh_node constructor.\n";
```
is easily another drag on perf (but I don't recommend removing it)

src/TheRestOfYourLife/bvh.h
```diff
@@ -13,12 +13,20 @@
...
+#include <algorithm>
 
 
 class bvh_node : public hittable  {
     public:
-        bvh_node() {}
-        bvh_node(hittable **l, int n, double time0, double time1);
+        bvh_node() = delete;
+
+        bvh_node::bvh_node(hittable_list &list, double time0, double time1)
```
trevordblack: throws error in clang

src/TheRestOfYourLife/bvh.h
```diff
@@ -28,131 +36,82 @@ class bvh_node : public hittable  {
...
 
-int box_y_compare (const void * a, const void * b)
-{
-    aabb box_left, box_right;
-    hittable *ah = *(hittable**)a;
-    hittable *bh = *(hittable**)b;
+bvh_node::bvh_node(
+    std::vector<hittable*> &objects, size_t start, size_t end, double time0, double time1
+) {
+    if (end <= start) {
```
trevordblack: Should this be copied over to _TheNextWeek_?

books/RayTracingInOneWeekend.html
```diff
@@ -2307,38 +2323,45 @@
...
+
+        objects.add(new sphere(vec3(0,-1000,0), 1000, new lambertian(vec3(0.5, 0.5, 0.5))));
+
         int i = 1;
         for (int a = -11; a < 11; a++) {
             for (int b = -11; b < 11; b++) {
                 auto choose_mat = random_double();
                 vec3 center(a + 0.9*random_double(), 0.2, b + 0.9*random_double());
-                if ((center-vec3(4,0.2,0)).length() > 0.9) {
+                if ((center - vec3(4, .2, 0)).length() > 0.9) {
```
trevordblack: `vec3(4, 0.2, 0))`

books/RayTracingTheNextWeek.html
```diff
@@ -641,20 +642,23 @@
...
-        if (list_size < 1) return false;
+        if (objects.empty()) return false;
+
         aabb temp_box;
-        bool first_true = list[0]->bounding_box(t0, t1, temp_box);
+        bool first_true = objects[0]->bounding_box(t0, t1, temp_box);
+
         if (!first_true)
             return false;
         else
```
trevordblack: unnecessary else

books/RayTracingTheNextWeek.html
```diff
@@ -756,35 +766,38 @@
...
+        } else if (object_span == 2) {
+            if (comparator(objects[start], objects[start+1])) {
+                left = objects[start];
+                right = objects[start+1];
+            } else {
+                left = objects[start+1];
+                right = objects[start];
+            }
+        } else {
+            std::sort(objects.begin() + start, objects.begin() + end, comparator);
```
trevordblack: Need to add `#include <algorithm>` for the `std::sort`

books/RayTracingTheNextWeek.html
```diff
@@ -1825,21 +1858,21 @@
...
+        auto white = new lambertian(new constant_texture(vec3(0.73, 0.73, 0.73)));
+        auto green = new lambertian(new constant_texture(vec3(0.12, 0.45, 0.15)));
+        auto light = new diffuse_light(new constant_texture(vec3(15, 15, 15)));
 
-        return new hittable_list(list,i);
+        objects.add(new yz_rect(0, 555, 0, 555, 555, green));
+        objects.add(new yz_rect(0, 555, 0, 555, 0, red));
+        objects.add(new xz_rect(213, 343, 227, 332, 554, light));
+        objects.add(new xz_rect(0, 555, 0, 555, 0, white));
+        objects.add(new xy_rect(0, 555, 0, 555, 555, white));
```
trevordblack: The sharing of materials between the objects in this scene is the #1 reason that rewriting the memory deletions is so hard
:((

src/TheNextWeek/main.cc
hollasch: 1. Make the change easy.
2. Make the easy change.

First step is to have code for a background color. Next step is to use it in the `ray_color` function. Final step is to get the background color from the scene/render-params. This is just the first iteration.

src/TheNextWeek/hittable_list.h
hollasch: Will fix.

src/TheRestOfYourLife/hittable_list.h
hollasch: Done.

src/TheRestOfYourLife/hittable_list.h
hollasch: Just being explicit about the conversion from signed to unsigned. But it looks like my compiler doesn't care either, so I'll remove the cast.

src/TheRestOfYourLife/main.cc
hollasch: Future attractions.

src/TheNextWeek/bvh.h
hollasch: I was being pedantic about disallowing the construction of unspecified BVH nodes, but I can definitely see how I'm just adding a cognitive speed bump in this context. Would totally do this in production code, but will simplify.

hollasch: I'm just going to pull the parameterless ctor altogether, and ignore the fact that C++ will autogenerate the default empty constructor.

src/TheNextWeek/bvh.h
hollasch: All gone.

src/TheNextWeek/bvh.h
hollasch: Eh. Ok.

src/TheNextWeek/bvh.h
hollasch: So, you want me to use modern C++ here? `decltype` or `std::function<>`?  :D

hollasch: Hmmm. The two alternatives are

```
    decltype(box_x_compare) *comparator;
    switch (random_int(0,2)) {
        default:
        case 0: comparator = box_x_compare; break;
        case 1: comparator = box_y_compare; break;
        case 2: comparator = box_z_compare; break;
    }
```
or
```
    #include <functional>
    ...
    std::function<bool(const hittable*, const hittable*)> comparator;
    switch (random_int(0,2)) {
        default:
        case 0: comparator = box_x_compare; break;
        case 1: comparator = box_y_compare; break;
        case 2: comparator = box_z_compare; break;
    }
```
Gonna stick with ternary; I believe it's much easier for a general audience to grok.

src/TheRestOfYourLife/bvh.h
hollasch: All gone.

src/TheRestOfYourLife/bvh.h
hollasch: _Oif_. Excellent catch.

books/RayTracingInOneWeekend.html
hollasch: Will do.

books/RayTracingTheNextWeek.html
hollasch: Done.

src/TheNextWeek/bvh.h
hollasch: Actually, I can hoist this sanity check out as a prepass on the unsorted `std::vector` before invoking the sort function.

books/RayTracingTheNextWeek.html
hollasch: This change is largely about moving the dynamic list to using `std::vector`, and thus having defined destructor logic. It currently has bare `hittable` pointers. For the next rev, we can convert this to `std::vector<std::shared_ptr<hittable>>`, and then ensure that object/material creation returns a shared pointer.

src/TheNextWeek/bvh.h
hollasch: Oh, never mind; it's not a sanity check, it actually computes the aabb (over and over; sigh).

I actually did start out on a cached aabb implementation, but that ended up in the weeds pretty quickly. Rolled back for a more incremental approach.

hollasch:
Block of 7 commits above incorporate review feedback.

PR #324, opened 2020-01-15
Use shared_ptr for materials and objects

hollasch:
For src/inOneWeekend code only.


hollasch:
Up for consideration. Do we want to do this? It eliminates the memory leaks, particularly in multi-frame scenarios. The cost is additional C++ jargon. We'll have to go over this in the books, but it shouldn't be too hard.

trevordblack:
I give a tentative approval to the addition of `shared_ptr`

petershirley:
I think it's a good idea.

hollasch:
Superseded by #328

PR #328, opened 2020-01-19
Memory management 2

hollasch:
Rebased form of `memory-management` branch incorporating Trevor's latest changes.

This is the completed change to using `shared_ptr` to address existing memory leaks.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -915,7 +915,11 @@ We add a list of objects:
     #define HITTABLE_LIST_H
 
     #include "hittable.h"
+    #include <memory>
     #include <vector>
+
+    using std::shared_ptr;
+    using std::make_shared;
 
     class hittable_list: public hittable {
         public:
@@ -954,8 +958,8 @@ We add a list of objects:
 
 ## Some New C++ Features
 
-This code uses two C++ features that may trip you up if you're not normally a C++ programmer:
-`vector` and `shared_ptr`.
+The `hittable_list` class code uses two C++ features that may trip you up if you're not normally a
+C++ programmer: `vector` and `shared_ptr`.
 
 `shared_ptr<type>` is a pointer to some allocated type, with reference-counting semantics.
 Every time you assign its value to another shared pointer (usually with a simple assignment), the
@@ -975,16 +979,24 @@ Typically, a shared pointer is first initialized with a newly-allocated object, 
 arguments `(1, true)`. It returns a `shared_ptr<thing>`. This can be simplified as in the second
 line with the `auto` type declaration, because the type is sufficiently defined by the return type
 of `make_shared<thing2>`.
+</div>
 
 We'll use shared pointers in our code, because it allows multiple objects to use a common object
 (for example, a bunch of spheres that all use the same material), and because it makes memory
 management automatic and easier to reason about.
-</div>
+
+`std::shared_ptr` is included with the `<memory>` header.
 
 The second C++ feature you may be unfamiliar with is `std::vector`. This is a generic array-like
 collection of an arbitrary type. Above, we use a collection of pointers to `hittable`. `std::vector`
 automatically grows as more values are added: `objects.push_back(object)` adds a value to the end of
 the `std::vector` member variable `objects`.
+
+`std::vector` is included with the `<vector>` header.
+
+Finally, the `using` statements in listing [hittable-list-initial] tell the compiler that we'll be
+getting `shared_ptr` and `make_shared` from the `std` library, so we don't need to prefex these with
+`std::` every time we reference them.
 
 ## Common Constants and Utility Functions
 
@@ -999,8 +1011,16 @@ file.
     #ifndef RTWEEKEND_H
     #define RTWEEKEND_H
 
-    #include <limits>
     #include <cmath>
+    #include <cstdlib>
+    #include <limits>
+    #include <memory>
+
+
+    // Usings
+
+    using std::shared_ptr;
+    using std::make_shared;
 
     // Constants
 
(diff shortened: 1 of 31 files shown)


hollasch:
Resolves #147

books/RayTracingInOneWeekend.html
```diff
@@ -920,15 +920,15 @@
     class hittable_list: public hittable {
         public:
             hittable_list() {}
-            hittable_list(hittable* object) { add(object); }
+            hittable_list(shared_ptr<hittable> object) { add(object); }
```
trevordblack: This caught my eye.
It's a little strange to have a list class that has a valued-constructor of the first element.

But going through the usage in the code base, it's pretty reasonable.
It's only used in places where the list only has a single value.

trevordblack: Needs
```
#include <memory>
```

books/RayTracingInOneWeekend.html
```diff
@@ -952,10 +952,41 @@
     [Listing [hittable-list-initial]: <kbd>[hittable_list.h]</kbd> The hittable_list class]
 </div>
 
-If you're unfamiliar with C++'s `std::vector`, this is a generic array-like collection of an
-arbitrary type. Above, we use a collection of pointers to `hittable`. `std::vector` automatically
-grows as more values are added: `objects.push_back(object)` adds a value to the end of the
-`std::vector` member variable `objects`.
+## Some New C++ Features
```
trevordblack: Should probably qualify that `shared_ptr` is in the `<memory>` header

books/RayTracingTheNextWeek.html
```diff
@@ -906,14 +923,14 @@
 where you used to have
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-    new lambertian(vec3(0.5, 0.5, 0.5)))
+        ...make_shared<lambertian>(vec3(0.5, 0.5, 0.5))
```
trevordblack: 4 spaces instead of 3

books/RayTracingTheNextWeek.html
```diff
@@ -1045,41 +1068,46 @@
...
+            static double* ranfloat;
+            static int* perm_x;
+            static int* perm_y;
+            static int* perm_z;
     };
 
     static double* perlin_generate() {
-        auto p = new double[256];
-        for (int i = 0; i < 256; ++i)
+        auto p = new double[perlin::point_count];
```
trevordblack: Are we not moving `perlin` over to `vector`s and `shared_ptr`s?

books/RayTracingTheNextWeek.html
```diff
@@ -1305,10 +1346,13 @@
...
                 return perlin_interp(c, u, v, w);
             }
-            static vec3 *ranvec;
-            static int *perm_x;
-            static int *perm_y;
-            static int *perm_z;
+
+        public:
+            static const int point_count = 256;
+            static vec3* ranvec;
```
trevordblack: Are we not moving `perlin` over to `vector`s and `shared_ptr`s?

src/common/rtweekend.h
```diff
@@ -12,8 +12,14 @@
 #include <cstdlib>
 #include <limits>
 #include <cmath>
+#include <memory>
 
 
+// Usings
+
+using std::shared_ptr;
```
trevordblack: These `using`s need to be explicitly written out in the text. Otherwise, everyone is going to be asking why `shared_ptr` and `make_shared` is undefined

src/common/texture.h
```diff
@@ -93,6 +100,7 @@ class image_texture : public texture {
             return vec3(r, g, b);
         }
 
+    public:
         unsigned char *data;
```
trevordblack: `vector`s and `shared_ptr`s?

books/RayTracingInOneWeekend.html
hollasch: Memory is a project-wide include, via `rtweekend.h`. (We include cstdlib, limits, cmath and memory.)

books/RayTracingInOneWeekend.html
hollasch: Will do.

books/RayTracingTheNextWeek.html
hollasch: Good catch. Will fix.

src/common/rtweekend.h
hollasch: Worse than that, `hittable_list` is presented just before the `rtweekend.h` common header. Will rework.

books/RayTracingTheNextWeek.html
hollasch: No, there's no need. See comment above.

books/RayTracingTheNextWeek.html
hollasch: No need. These basis values are never shared out, but are only used to compute Perlin values. In this case, a static array is sufficient (and simple), and has the lifetime we need (across many uses & frames if need be). It just needs to be initialize once at the beginning of the program.

Still, this caught a fragment of old code, which allocated the buffer. The new code just declares a static array.

hollasch: Will update.

books/RayTracingInOneWeekend.html
hollasch: Ah, but `rtweekend.h` is presented _after_ `hittable_list`. Sigh. I'll rework the text.

books/RayTracingTheNextWeek.html
```diff
@@ -1069,47 +1080,44 @@
...
-    };
-
-    static double* perlin_generate() {
-        static double p[perlin::point_count];
+            double* ranfloat;
+            int* perm_x;
+            int* perm_y;
+            int* perm_z;
+
+            void init() {
```
hollasch: This code moves into the constructor in the following commit.

PR #330, opened 2020-01-22
Scene

hollasch:
In the end, the changes needed to fix the various images in _The Next Week_ were fairly straight-forward, and a formal `scene` class was not needed. This is particularly true because only the second book's code has multiple scenes included in it.

Instead, I opted for a simple switch statement that sets up the viewing parameters and background color in addition to the prior geometry creation.

Leading up to this, I took the liberty to perform some clarifying refactorings as well.

### DIFF
--- a/src/TheRestOfYourLife/main.cc
+++ b/src/TheRestOfYourLife/main.cc
@@ -91,29 +91,28 @@ hittable_list cornell_box(camera& cam, double aspect) {
 
 
 int main() {
-    int nx = 600;
-    int ny = 600;
+    int image_width = 600;
+    int image_height = 600;
     int num_samples = 100;
     int max_depth = 50;
 
-    std::cout << "P3\n" << nx << ' ' << ny << "\n255\n";
+    std::cout << "P3\n" << image_width << ' ' << image_height << "\n255\n";
 
     camera cam;
-    auto aspect = double(ny) / double(nx);
-
-    auto world = cornell_box(cam, aspect);
+    auto aspect_ratio = double(image_width) / image_height;
+    auto world = cornell_box(cam, aspect_ratio);
 
     auto lights = make_shared<hittable_list>();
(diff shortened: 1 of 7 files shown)


src/TheRestOfYourLife/main.cc
```diff
@@ -91,33 +91,32 @@ hittable_list cornell_box(camera& cam, double aspect) {
...
+        for (int i = 0; i < image_width; ++i) {
             vec3 color;
-            for (int s = 0; s < num_samples; ++s) {
-                auto u = (i + random_double()) / nx;
-                auto v = (j + random_double()) / ny;
+            for (int s = 0; s < samples_per_pixel; ++s) {
+                auto u = (i + random_double()) / image_width;
+                auto v = (j + random_double()) / image_height;
                 ray r = cam.get_ray(u, v);
                 color += ray_color(r, world, lights, max_depth);
```
trevordblack: Needs `background_color`.

hollasch:
I did find the original background color of (0.7, 0.8, 1.0), but that was quite blue-shifted compared to the images in the book. I ultimately went for the less saturated blueish-white color, which seems to match the images better.

src/TheRestOfYourLife/main.cc
hollasch: Done.

trevordblack:
Everything looks good to me.
This turned into a bit more of a headache than you were probably imagining.

Thanks for the follow-through.

hollasch:
Actually, much _less_ of a headache. I was planning on creating a new `scene` class, perhaps with a `render()` or `ray_color()` method, or perhaps a `scene` class and a `renderer` class, coupled with some data tables for each scene, blah blah blah.

It would probably be the correct thing to do for a released program, but for this project, it would be gratuitous overhead.

PR #332, opened 2020-01-23
Refactoring

### LINKED ISSUE #331: Final v3 refactor pass
Final pass of all code for opportunistic refactoring, with no change in behavior.

There are many places where the code could be simplified a bit, for example, inverting conditional tests, taking advantage of early returns, improving naming, and so forth.

This is a general bucket item for the pass prior to final v3 release.

hollasch:
A bunch of refactoring and clean up I'd been meaning to do.

Resolves #331.


books/RayTracingTheRestOfYourLife.html
```diff
@@ -1215,17 +1217,15 @@
...
 
 
     void onb::build_from_w(const vec3& n) {
         axis[2] = unit_vector(n);
-        vec3 a;
-        if (fabs(w().x()) > 0.9)
-            a = vec3(0, 1, 0);
-        else
-            a = vec3(1, 0, 0);
+        vec3 a = (fabs(w().x()) > 0.9) ? vec3(0,1,0) : vec3(1,0,0);
```
trevordblack: Lot going on here...

src/TheRestOfYourLife/onb.h
```diff
@@ -18,24 +18,31 @@ class onb
...
 
 
 void onb::build_from_w(const vec3& n) {
     axis[2] = unit_vector(n);
-    vec3 a;
-    if (fabs(w().x()) > 0.9)
-        a = vec3(0, 1, 0);
-    else
-        a = vec3(1, 0, 0);
+    vec3 a = (fabs(w().x()) > 0.9) ? vec3(0,1,0) : vec3(1,0,0);
```
trevordblack: lot going on here...

CMakeLists.txt
```diff
@@ -63,7 +63,6 @@ set ( SOURCE_REST_OF_YOUR_LIFE
   src/TheRestOfYourLife/hittable.h
   src/TheRestOfYourLife/hittable_list.h
   src/TheRestOfYourLife/material.h
-  src/TheRestOfYourLife/moving_sphere.h
```
trevordblack: Why you do this?

trevordblack:
Why the deletion of moving_sphere in RestOfYourLife?

Otherwise, Excellent changes.

Does it build?

CMakeLists.txt
hollasch: It's dead code; unused.

src/TheRestOfYourLife/onb.h
hollasch: Is that an objection? I only mildly prefer the ternary here.

In a related vein, I have thought of defining global constants for things like `origin`, `unit_x`, `unit_y`, and `unit_z`.

hollasch:
Yup, I built it. Didn't run the timing gauntlet, though. Doing that now.

Being unable to run all 10 scene types for theNextWeek bugs me, though. Perhaps I could add that as the single command-line parameter. Was considering rich command-line parameterization, but in the end decided it was a lot of added complexity for little (expository) benefit.

trevordblack:
No command-line parameterization. It makes a little sense for the code, but is actively awful for the text.

Slight objection to the ternary. It's just a little hard to parse, is all. Don't do global constants. Global is already a c++ construct that is being shied away from in other langs. Plus, maintaining globals in the text would be a pain.

CMakeLists.txt
trevordblack: It seems like a waste to remove it..

hollasch:
On limited, scene-selecting command-line parameterization, we don't need to document this in the book — the multi-scene code is already missing from the book, and I think that's fine. Before we had big chunks of dead code plus a bunch of commented-out lines. I'm now thinking of just having an integer option. If present, use it to select the scene.

The advantage to this is that contributors can run a sanity test before submitting a pull request (which we can require). Without this, we have to set up and test every change ourselves manually.

hollasch:
We have a performance regression, but apparently not in this branch. I'll track it down separately. All tests pass with this branch.

PR #337, opened 2020-01-26
constant_medium: refactor + additional text

### LINKED ISSUE #186: constant_medium::hit fails with concave boundary shapes
The `constant_medium` class allows _any_ other shape to be used as the boundary of the medium, but its `hit` method will only work correctly when the boundary shape is convex. For non-convex shapes it can return a miss in cases where it should return a hit. Specifically, if extending the ray backwards would enter & exit the volume, then the current `hit` method will fail to find any places where extending the ray _forwards_ would enter the volume.

I don't think this breaks anything in the book because the book only seems to have convex shapes, but it might be worth adding a warning about it or something? It could catch people out if they add support for loading 3D models and try to use one as a volume, for example.

hollasch:
- Opportunistic refactoring of constant_medium.h, updated in book also.
- Set up main.cc to allow for custom samples_per_pixel and ray depth per
  scene.
- Add explanatory text that the constant_medium boundary shape must be
  convex.

Resolves #186

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2430,7 +2430,7 @@ density $C$ and the boundary. I’ll use another hittable for the boundary. The 
     class constant_medium : public hittable {
         public:
             constant_medium(shared_ptr<hittable> b, double d, shared_ptr<texture> a)
-                : boundary(b), density(d)
+                : boundary(b), neg_inv_density(-1/d)
             {
                 phase_function = make_shared<isotropic>(a);
             }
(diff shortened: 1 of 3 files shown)


trevordblack:
Looking over chapter 9 now, the `isotropic` material produces a `scattered` ray of
```
scattered = ray(rec.p, random_in_unit_sphere()));
```
which is of non-unit length. So I figured we might want to make it a unit length direction. I remembered that I wrote a unit-length random function: `random_on_unit_length`. Which, since it's gone, must have been removed sometime. Almost certainly by me. Anyway. Not a high value.

trevordblack:
I think `leave that as an exercise for the reader` is an adequate "fix" for the convex issue.
The refactor also leads (at least me) to a simple implementation of the concave fix.

hollasch:
`random_on_unit_sphere()` → `random_unit_vector()`

hollasch:
I don't _think_ we need to make the bounce direction unit length, as it's just a new ray direction. If it ultimately needs to be normalized, that would happen on next use anyway.

PR #341, opened 2020-02-12
Simplify lambertian scatter direction calculation

### LINKED ISSUE #340: Unnecessary calculation in lambertian::scatter()
The `lambertian::scatter()` method (`material.h`) adds and then immediately subtracts the hit point:

    vec3 target = rec.p + rec.normal + random_unit_vector();
    scattered = ray(rec.p, target-rec.p, r_in.time());

hollasch:
Resolves #340

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -251,9 +251,9 @@ Be sure that in the materials you have the scattered rays be at the time of the 
             virtual bool scatter(
                 const ray& r_in, const hit_record& rec, vec3& attenuation, ray& scattered
             ) const {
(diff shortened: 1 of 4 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -2459,7 +2459,7 @@
             virtual bool scatter(
                 const ray& r_in, const hit_record& rec, vec3& attenuation, ray& scattered
             ) const  {
-                scattered = ray(rec.p, random_in_unit_sphere());
+                scattered = ray(rec.p, random_in_unit_sphere(), r_in.time());
```
trevordblack: Nice Catch!

PR #357, opened 2020-02-19
sphere intersection explaination reduction

dafhi:
hacked and slashed till it read easily

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -483,23 +483,15 @@ Adding a Sphere
 <div class='together'>
 Let’s add a single object to our ray tracer. People often use spheres in ray tracers because
 calculating whether a ray hits a sphere is pretty straightforward. Recall that the equation for a
-sphere centered at the origin of radius $R$ is $x^2 + y^2 + z^2 = R^2$. The way you can read that
-equation is “for any $(x, y, z)$, if $x^2 + y^2 + z^2 = R^2$ then $(x,y,z)$ is on the sphere, and
-otherwise it is not”. It gets uglier if the sphere center is at
-$(\mathbf{c}_x, \mathbf{c}_y, \mathbf{c}_z)$:
+sphere centered at the origin is $x^2 + y^2 + z^2 = R^2$.
+It gets uglier if the center is at $\mathbf{C}$:
 
-  $$ (x-\mathbf{c}_x)^2 + (y-\mathbf{c}_y)^2 + (z-\mathbf{c}_z)^2 = R^2 $$
+  $ (x-\mathbf{c}_x)^2 + (y-\mathbf{c}_y)^2 + (z-\mathbf{c}_z)^2 = R^2 $
 </div>
 
 <div class='together'>
 In graphics, you almost always want your formulas to be in terms of vectors so all the x/y/z stuff
-is under the hood in the `vec3` class. You might note that the vector from center
-$\mathbf{c} = (\mathbf{c}_x,\mathbf{c}_y,\mathbf{c}_z)$ to point $\mathbf{P} = (x,y,z)$ is
-$(\mathbf{p} - \mathbf{c})$, and therefore
-
-  $$ (\mathbf{p} - \mathbf{c}) \cdot (\mathbf{p} - \mathbf{c})
-     = (x-\mathbf{c}_x)^2 + (y-\mathbf{c}_y)^2 + (z-\mathbf{c}_z)^2
-  $$
+is under the hood in the `vec3` class.
 </div>
 
 <div class='together'>
@@ -508,11 +500,7 @@ So the equation of the sphere in vector form is:
   $$ (\mathbf{p} - \mathbf{c}) \cdot (\mathbf{p} - \mathbf{c}) = R^2 $$
 </div>
 
-<div class='together'>
-We can read this as “any point $\mathbf{p}$ that satisfies this equation is on the sphere”. We want
-to know if our ray $p(t) = \mathbf{a} + t\vec{\mathbf{b}}$ ever hits the sphere anywhere. If it does
-hit the sphere, there is some $t$ for which $p(t)$ satisfies the sphere equation. So we are looking
-for any $t$ where this is true:
+<div class='together'>We are looking for any $t$ where this is true:
 
   $$ (p(t) - \mathbf{c})\cdot(p(t) - \mathbf{c}) = R^2 $$
 
@@ -523,17 +511,7 @@ or expanding the full form of the ray $p(t)$:
 </div>
 
 <div class='together'>
-The rules of vector algebra are all that we would want here, and if we expand that equation and
-move all the terms to the left hand side we get:
-
-  $$ t^2 \vec{\mathbf{b}}\cdot\vec{\mathbf{b}}
-     + 2t \vec{\mathbf{b}} \cdot \vec{(\mathbf{a}-\mathbf{c})}
-     + \vec{(\mathbf{a}-\mathbf{c})} \cdot \vec{(\mathbf{a}-\mathbf{c})} - R^2 = 0
-  $$
-</div>
-
-<div class='together'>
-The vectors and $R$ in that equation are all constant and known. The unknown is $t$, and the
+The vectors and $R$ are constant and known. The unknown is $t$, and the
 equation is a quadratic, like you probably saw in your high school math class. You can solve for $t$
 and there is a square root part that is either positive (meaning two real solutions), negative
 (meaning no real solutions), or zero (meaning one real solution). In graphics, the algebra almost
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -511,6 +511,16 @@ or expanding the full form of the ray $p(t)$:
 </div>
 
 <div class='together'>
+The rules of vector algebra are all that we would want here, and if we expand that equation and
+move all the terms to the left hand side we get:
+
+  $$ t^2 \vec{\mathbf{b}}\cdot\vec{\mathbf{b}}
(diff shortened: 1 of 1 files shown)


dafhi:
Didn't realize I cut too much.  The inclusion of full written-out equation is of great benefit.  
t2b⃗ ⋅b⃗ +2tb⃗ ⋅(a−c)→+(a−c)→⋅(a−c)→−R2=0

If nobody objects within the next few days I'll delete this pull request and re-upload

trevordblack:
It's good to bring up issues with clarity in the text.

On the one hand I'd say try to use a light touch when you want to make changes in the body of the book.
But on the other, your experience of the words is much more authentic than any of us (who've poured over it, edited it, multiple times).

In most cases we try to discretize every change into an issue. However, for big changes, meaning code architecture changes or changes to the meaning in the text. We usually end up creating one PR that encompasses includes every change, and then PR that for discussion.

Our preference so far has been to add detail to the text, not remove it, so simplifying it isn't something we've really put enough polish into. It's something we're overdue for.

We're due to release v3 pretty soon.
I have an issue assigned to me that is just going through and refactoring which will be after v3. You have the choice of pushing off text changes until after v3, or trying to increase readability now.

Up to you.
The issues you're bringing up have been fantastic.

dafhi:
Oh, thank you!  It looks like I may be able to complete a single-file request via website.  I'll make the change now and give that a shot or just re-upload to my branch

dafhi:
Ok, done!
should contain: t2b⃗ ⋅b⃗ +2tb⃗ ⋅(a−c)→+(a−c)→⋅(a−c)→−R2=0

books/RayTracingInOneWeekend.html
```diff
@@ -483,23 +483,15 @@
...
 Let’s add a single object to our ray tracer. People often use spheres in ray tracers because
 calculating whether a ray hits a sphere is pretty straightforward. Recall that the equation for a
-sphere centered at the origin of radius $R$ is $x^2 + y^2 + z^2 = R^2$. The way you can read that
-equation is “for any $(x, y, z)$, if $x^2 + y^2 + z^2 = R^2$ then $(x,y,z)$ is on the sphere, and
-otherwise it is not”. It gets uglier if the sphere center is at
-$(\mathbf{c}_x, \mathbf{c}_y, \mathbf{c}_z)$:
+sphere centered at the origin is $x^2 + y^2 + z^2 = R^2$.
+It gets uglier if the center is at $\mathbf{C}$:
 
-  $$ (x-\mathbf{c}_x)^2 + (y-\mathbf{c}_y)^2 + (z-\mathbf{c}_z)^2 = R^2 $$
```
hollasch: This needs to use double dollar signs — it's a block equation, rather than inline.

books/RayTracingInOneWeekend.html
```diff
@@ -483,23 +483,15 @@
 <div class='together'>
 Let’s add a single object to our ray tracer. People often use spheres in ray tracers because
 calculating whether a ray hits a sphere is pretty straightforward. Recall that the equation for a
-sphere centered at the origin of radius $R$ is $x^2 + y^2 + z^2 = R^2$. The way you can read that
-equation is “for any $(x, y, z)$, if $x^2 + y^2 + z^2 = R^2$ then $(x,y,z)$ is on the sphere, and
-otherwise it is not”. It gets uglier if the sphere center is at
-$(\mathbf{c}_x, \mathbf{c}_y, \mathbf{c}_z)$:
+sphere centered at the origin is $x^2 + y^2 + z^2 = R^2$.
+It gets uglier if the center is at $\mathbf{C}$:
```
hollasch: Lines should wrap at 100 columns.

books/RayTracingInOneWeekend.html
```diff
@@ -483,23 +483,15 @@
 <div class='together'>
 Let’s add a single object to our ray tracer. People often use spheres in ray tracers because
 calculating whether a ray hits a sphere is pretty straightforward. Recall that the equation for a
-sphere centered at the origin of radius $R$ is $x^2 + y^2 + z^2 = R^2$. The way you can read that
-equation is “for any $(x, y, z)$, if $x^2 + y^2 + z^2 = R^2$ then $(x,y,z)$ is on the sphere, and
-otherwise it is not”. It gets uglier if the sphere center is at
-$(\mathbf{c}_x, \mathbf{c}_y, \mathbf{c}_z)$:
+sphere centered at the origin is $x^2 + y^2 + z^2 = R^2$.
```
hollasch: What's the rationale for dropping the explanation of `R`?

hollasch: I think the commentary about interpreting the value of the left-hand side is to set up understanding for later code where we distinguish between points _inside_ the sphere and points _on_ the sphere.

books/RayTracingInOneWeekend.html
```diff
@@ -483,23 +483,15 @@
...
 <div class='together'>
 In graphics, you almost always want your formulas to be in terms of vectors so all the x/y/z stuff
-is under the hood in the `vec3` class. You might note that the vector from center
-$\mathbf{c} = (\mathbf{c}_x,\mathbf{c}_y,\mathbf{c}_z)$ to point $\mathbf{P} = (x,y,z)$ is
-$(\mathbf{p} - \mathbf{c})$, and therefore
-
-  $$ (\mathbf{p} - \mathbf{c}) \cdot (\mathbf{p} - \mathbf{c})
-     = (x-\mathbf{c}_x)^2 + (y-\mathbf{c}_y)^2 + (z-\mathbf{c}_z)^2
-  $$
+is under the hood in the `vec3` class.
```
hollasch: If this derivation is removed, the equation at line 500 is left as an exercise for the reader. I don't see a reason to pull this explanation.

books/RayTracingInOneWeekend.html
```diff
@@ -523,17 +511,7 @@
...
-
-  $$ t^2 \vec{\mathbf{b}}\cdot\vec{\mathbf{b}}
-     + 2t \vec{\mathbf{b}} \cdot \vec{(\mathbf{a}-\mathbf{c})}
-     + \vec{(\mathbf{a}-\mathbf{c})} \cdot \vec{(\mathbf{a}-\mathbf{c})} - R^2 = 0
-  $$
-</div>
-
-<div class='together'>
-The vectors and $R$ in that equation are all constant and known. The unknown is $t$, and the
+The vectors and $R$ are constant and known. The unknown is $t$, and the
```
hollasch: In general, the original text sets up the reader for understanding the mathematics behind ray intersection with a surface defined by an implicit equation. In this situation it picks the simple case of ray-sphere intersection, but provides a basis by which a reader could also derive the math for ellipsoids, cones, torii and other surfaces. Removing the derivation leaves the reader with only the final equations for a sphere.

I believe the text is weakened by removing these explanations.

dafhi: sure, okay.  i'm not a math whiz, especially with 3d

books/RayTracingInOneWeekend.html
dafhi: i'll give this a shot maybe later today.

dafhi: In high school, one of my English teachers discussed the topic of, I can't remember the term that was used, but the idea was about efficiency.
"An example of [inefficiency] is 'dog puppies.'"

This idea stuck with me.  If I've proofread something I typed enough times, imo it's a pretty easy read.

Efficiency isn't all there is in life, of course.  I had a "logic"-interested friend in college who asked our philosophy professor "Isn't the goal of language to be efficient?"

"The primary goal of a language should be to be beautiful."

Efficiency tends to have priority in my life.

My thoughts on the first paragraph:

"Let's add a single object to our ray tracer.  People often use spheres in ray tracers because .."
Nice.

"Recall that the equation for a sphere centered at the origin of radius R is x2 + y2 + z2 = R2."
OK he just said R twice.  Let's axe "of Radius R"

"The way you can read that equation is 'for any (x,y,z), if x2 + y2 + z2 = R2'.."
Another thing said twice.

"..then (x,y,z)"
hmm

"..is on the sphere, and otherwise it is not."
binary logic.  the other half is implied.  axe that and keep reading

"It gets uglier if the sphere center is at (cx,cy,cz)."
level of understanding:  instant

After reading the second paragraph, I saw that (cx,cy,xz) became represented as 'c' so I choose to 'c' (cx,cy,cz) in first paragraph.

At this point I look again at "The way you can read that equation .. otherwise it is not."
Look like dragon's breath.  x,y & z are already implied from the rather simple and beautiful looking equations that follow.  "Axe the whole sentence."

hollasch: > "Recall that the equation for a sphere centered at the origin of radius R is x2 + y2 + z2 = R2."
OK he just said R twice. Let's axe "of Radius R"

If I give you the equation n₁sinθ₁ = n₂sinθ₂, it's worthless unless I then tell you what n₁, θ₁, n₂ and θ₂ all _mean_. It's not repetition to say “with incident angle θ₁”.

I agree with the comment about not eliding the “otherwise false” phrases.

dafhi: if one took high school algebra there's a ridiculously slim chance of the need to explain what R represents

dafhi: yeah i'm a hot-head.  sorry, i'm trying.  with what you bring up, and 'non double-speak' one could say x2 + y2 + z2 = Radius2

books/RayTracingInOneWeekend.html
dafhi: Looks janky imo.  Your call

trevordblack:
I think this PR was addressed by #371
Should this PR be closed out?

dafhi:
I thought of something.  Don't close this yet

dafhi:
sorry wrong thread on my prev deleted comment

PR #367, opened 2020-02-22
Improve color [0,1]->[0,255] mapping

### LINKED ISSUE #359: Quick note about Random() * 255.999
[updated:  old ceiling was 2]

Coming from my background in 2d programming
 
rnd*255.999 will yield a bias.

  an illustration:

Range with ceiling of 3 will yield

0 to 0.499:  [integer result:  0]  1/6 chance
.5 to 1.499: [integer result:  1]  
1.5 to 2.499: [integer result:  2]
2.5 to 2.999: [integer result:  3]  1/6 chance

Corrected approach:  rnd*256 - 0.5

hollasch:
Resolves #359

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -1046,12 +1046,6 @@ file.
     inline double ffmin(double a, double b) { return a <= b ? a : b; }
     inline double ffmax(double a, double b) { return a >= b ? a : b; }
 
-    inline double clamp(double x, double min, double max) {
-        if (x < min) return min;
-        if (x > max) return max;
-        return x;
-    }
-
     // Common Headers
 
     #include "common/ray.h"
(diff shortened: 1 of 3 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -1046,12 +1046,6 @@
     inline double ffmin(double a, double b) { return a <= b ? a : b; }
     inline double ffmax(double a, double b) { return a >= b ? a : b; }
 
-    inline double clamp(double x, double min, double max) {
```
trevordblack: Do we replacing this with std::clamp?
If so, we need to include <algorithm>, and _maybe_ std::clamp?

hollasch: For now, we only require C++11, and `std::clamp` is C++17.

As a bonus, this also has the advantage of being more transparent to the reader.

PR #382, opened 2020-02-23
Prefix common header filenames with common/

### LINKED ISSUE #347: Book One, newly-typed vec3.h into "common"
Dev.

Creating a color gradient image (Listing 6)
source code looks like it wants newly-typed "vec3.h" into the common folder

hollasch:
Added explanatory text for the src/common/ directory, and made explicit
the location of headers in the common source directory.

Resolves #347

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -204,8 +204,13 @@ class `vec3` for colors, locations, directions, offsets, whatever. Some people d
 because it doesn’t prevent you from doing something silly, like adding a color to a location. They
 have a good point, but we’re going to always take the “less code” route when not obviously wrong.
 
+Since `vec3` is such a fundamental class, it will be common to all of our ray tracing project source
+code. If you look at the accompanying source code for this book, you'll see a directory under `src/`
+for each book. In addition, you'll see a directory named `src/common/`. That's where you'll find all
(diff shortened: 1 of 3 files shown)


trevordblack:
This gets a hard no from me on this one.

I hate the idea of asking the reader to put up infrastructure for books 2 and 3 when they've barely started book 1.
I understand that it's discordant that we have the repo with a common folder and the books without one. But adding the common folder is a change that I've hated from the beginning. I was okay with it because it would make our lives easier without impacting the end-user.

I'm really not happy with this change.

Sorry.

dafhi:
Just saw the request.

I'm with Kevin on this one.  

How about

".. class, we will introduce new functionality to it throughout the book.  If you do decide to type in most of the code, in the .. *ahem* .. unlikely event that you made a boo boo, you can always check the /src/common folder for assistance.  [feels like more could be said]

dafhi:
I'll have another look.  Been super busy

dafhi:
I like /common for 2 reasons.

1.  top-down thinking
2.  efficiency

Snafus can occur where changes break previous Listings.  For this reason I recommend optional params in some of the headers.

The 'common' reference simply indicate to the reader an important header file for all Listings in one book

hollasch:
See PR #387. Let readers search for each header, but set up CMake to properly pull from project + common directories without requiring path specification.

PR #387, opened 2020-03-01
Add src/common to include directories

hollasch:
This adds the src/common directory to the include directories, so
compilers will search there for headers in addition to the target source
directories.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -319,7 +319,7 @@ tastes. The second part of the header file contains vector utility functions:
 Now we can change our main to use this:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-    #include "common/vec3.h"
+    #include "vec3.h"
 
     #include <iostream>
 
@@ -370,7 +370,7 @@ The function $p(t)$ in more verbose code form I call `ray::at(t)`:
     #ifndef RAY_H
     #define RAY_H
 
-    #include "common/vec3.h"
+    #include "vec3.h"
 
     class ray {
         public:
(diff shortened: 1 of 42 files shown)


trevordblack:
This is in lieu of #347  ?

hollasch:
Inspired by it, I suppose. It's really fixing a weakness in the original `CMakeLists.txt` that based includes from the `src/` directory, along with the default path of finding headers in the same directory as the source file. This change basically says that in addition to the source directory, also look in `src/common` for all targets. Which means that you no longer have to explicitly include something from `common/`.

dafhi:
git compare cracks me.  i'll download locally and check it out

books/RayTracingInOneWeekend.html
```diff
@@ -319,7 +319,7 @@
 Now we can change our main to use this:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-    #include "common/vec3.h"
+    #include "vec3.h"
```
dafhi: overall happy with this change, though i fully support a common folder structure, which could be illustrated by a more experienced programmer than myself.

books/RayTracingInOneWeekend.html
```diff
@@ -1064,19 +1064,20 @@
...
 
-    #include "common/ray.h"
-    #include "common/vec3.h"
+    #include "ray.h"
+    #include "vec3.h"
 
     #endif
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [rtweekend-initial]: <kbd>[common/rtweekend.h]</kbd> The rtweekend.h common header]
+    [Listing [rtweekend-initial]: <kbd>[rtweekend.h]</kbd> The rtweekend.h common header]
```
dafhi: rtweekend seems like a solid candidate for leaving in common

hollasch: No headers move in this change. It simply eliminates the need to prefix headers with the 'common/' path. It also allows programmers to put the headers all in one source directory, or to use the source code as-is with some headers in common. Either way, the compiler will locate the header.

PR #388, opened 2020-03-05
Fix 384

hollasch:
Rephrase paragraph with overloaded use of the word “object”.


hollasch:
Gah! Aimed at `master`, pulled trigger.

PR #402, opened 2020-03-18
InOneWeekend: update ray_color() to match source

### LINKED ISSUE #391: Difference in src and text inline src
In InOneWeekend, the inline src for `ray_color` is different from the src.

hollasch:
Resolves #391


dafhi:
I specifically remember 0.0 not being a problem because it's fixed later in the natural progression
Listing .. [main.cc] Calculating reflected ray origins with tolerance.

where in the next sentence 'shadow acne' is mentioned.  I hope I'm not getting my line numbers mixed up

hollasch:
Excellent catch! Change incoming...

PR #403, opened 2020-03-20
InOneWeekend: repair hit tolerance from fix 391

hollasch:


dafhi:
new OS .. Slack is gone.  gimme a sec to examine the PR

dafhi:
looks good.

PR #406, opened 2020-03-21
NextWeek: Remove spurious decl bounding_box()

### LINKED ISSUE #405: Book Two moving_sphere:  aabb not declared
Dev.

Listing: [moving_sphere.h] A moving sphere
`
    virt bool bounding_box(tReal t0, tReal t1, aabb& output_box) const;`

boom.

(the line isn't needed)

hollasch:
We had a premature declaration for the virtual function
`moving_sphere::bounding_box()`. Removed it.

Fixes #405

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -171,7 +171,6 @@ times need not match up with the camera aperture open and close.
             {};
 
             virtual bool hit(const ray& r, double tmin, double tmax, hit_record& rec) const;
-            virtual bool bounding_box(double t0, double t1, aabb& output_box) const;
 
             vec3 center(double time) const;
 


trevordblack:
Uh oh.

I suppose we consider cutting some text...

line 30-33
```
The two hardest parts of this book are the BVH and the Perlin textures. This is why the title
suggests you take a week rather than a weekend for this endeavor. But you can save those for last if
you want a weekend project. Order is not very important for the concepts presented in this book, and
without BVH and Perlin texture you will still get a Cornell Box!
```

__... Order is not very important for the concepts presented in this book ...__

While order isn't important, I don't know of anyone that goes through the chapters out of order. And it also creates a weird thing where bounding box (which arguably should be the first chapter) is the second chapter.

We've run into a couple of bugs where a loose attitude with the ordering of the chapters has created weird bugs in the coed

trevordblack:
For a massive 11th hour change, maybe move BVH to Chapter 1 of NextWeek?

PR #426, opened 2020-04-02
Vec3 types

hollasch:
Feels to me like a solid improvement.


trevordblack:
My build system is borked atm.

Does this all build?

hollasch:
Yes, it's big, but if I missed something it's no big deal from a correctness perspective. Everything's still `vec3` under the hood.

petershirley:
Much better IMO

PR #451, opened 2020-04-06
Move acknowledgments to single common document

hollasch:
The prior version had an identical acknowledgments section copy and
pasted to the end of each book, a solution that was both tedious and
error prone.

This change uses Markdeep's include directive to import the section
from a new single `acknowledgments.md.html` file.

Additionally, this change uses Markdeep's auto reference, so that
`[acknowledgments](#acknowledgements)` is now replaced with the text
"acknowledgments section", which Markdeep auto links.

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -30,8 +30,8 @@ terms you will need to study the others.
 
 As before, https://in1weekend.blogspot.com/ will have further readings and references.
 
-Thanks to everyone who lent a hand on this project. You can find them in the [acknowledgments][] at
-the end of this book.
+Thanks to everyone who lent a hand on this project. You can find them in the acknowledgments section
+at the end of this book.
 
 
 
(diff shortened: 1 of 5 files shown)


books/acknowledgments.md.html
```diff
@@ -0,0 +1,76 @@
...
+  </div>
+</div>
+
+
+
+<!-- Markdeep: https://casual-effects.com/markdeep/ -->
+<link rel='stylesheet' href='../style/book.css'>
+<style class="fallback">body{visibility:hidden;white-space:pre;font-family:monospace}</style>
+<script src="markdeep.min.js"></script>
+<script src="https://casual-effects.com/markdeep/latest/markdeep.min.js"></script>
```
trevordblack, line 75: Also got rid of wrong link issue +1

books/RayTracingTheRestOfYourLife.html
```diff
@@ -2405,73 +2405,8 @@
...
-
-  <div class="indented">
-  Thanks to the team at [Limnu][] for help on the figures.
-
-  Huge shout out to Morgan McGuire for his fantastic [Markdeep][] library.
-  </div>
-</div>
-
-
+                               (insert acknowledgments.md.html here)
```
trevordblack, line 2408: Is this adequate?

hollasch: Yup. Classic Markdeep magical incantation.

books/acknowledgments.md.html
hollasch: Yeah, I closed that specific PR in favor of this one.

books/RayTracingTheRestOfYourLife.html
hollasch: Similar to how “... the acknowledgments section ...” gets an automagic link to the section.

PR #456, opened 2020-04-07
book1: Update highlighting for new headers

hollasch:
We weren't highlighting some of the changes to main.cc after creating
the new rtweekend.h, hittable_list.h, and sphere.h.


hollasch:
In master, some inline code spans look one way, other inline code spans look slightly different. They're all styled the same now.

hollasch:
@dafhi — you need to view it from the `fix-421` branch. Top of listing 23.

PR #458, opened 2020-04-08
Book 1: Rework math to new notation

hollasch:
*PREVIEW*

Points are uppercase bold, vectors are lowercase bold, no barb.

So far I like it, but we now have a problem that the notation sometimes doesn't match the figures.


petershirley:
Looks good to me.  I think I should just fix the figures.

hollasch:
Completed in PR #483.

Issue #407, opened 2020-03-22
Further Reading Documetn


trevordblack:
Okay, I have a bit of a jumbled mess where you'd typically find a brain.

For the `FurtherReading.html` (or `Bibliography.html`, I prefer `FurtherReading.html` in the first draft), how do we want to structure it?

From the wiki: https://github.com/RayTracing/raytracing.github.io/wiki/Aggregation-of-Possible-Next-Steps

We have a couple of dissimilar things jumbled together.

1. LOCATIONS of content
    - Chapters in the books
    - Blogs from the in1weekend blog spots
2. TYPES of content
   - External resources
   - Possible next steps (exercises)
     - Some are found in the next book
     - Some are not found in the next book

This is biting off more than we can chew, but, for my own sanity, at least,
I would like the project to get to a place that,

1. Overhaul of all the "Next Steps" chapters in the book.
    - One chapter at the end of each book
    - This chapter has 
      - brief overview of topics covered in the next book
      - Possible exercises that are _not_ covered in the next book
      - (don't mention topics from 2+ books on)
2. The `FurtherReading.html` doc contains
    - No exercises
    - External resources for every chapter in the books
    - An Assumed Knowledge section covering the basics (math, programming)
    - A Common section for resources that are relevant to multiple books, chapters
    - An Extension section for resources that are the next places to learn
    - (Maybe) An Unsolved Research Problems section for state-of-the-art resources/questions

This is explicitly not a text book, but at some point we might need to have a conversation about Exercises. The books already nod toward having them (in the Further Reading sections), but the explicit outlining of Exercises introduces a lot of friction for the reader.

Indeed, all of the structure outlined above is an increase in friction for the reader. So I don't strongly recommend any of it, but it's a topic I've been ruminating on for a few weeks, and have largely not had any success in congealing. The excessive structure is simply a coping mechanism.

_Originally posted by @trevordblack in https://github.com/RayTracing/raytracing.github.io/issues/299#issuecomment-568301471_

hollasch:
Old branch rebased to dev-major: `further-reading-2`.

In my opinion, I'd rather see the bibliography grow more organically — the current outline is quite aggressive in scope and easy to put off as a hard, large task.

trevordblack:
Propose punting to backlog

hollasch:
Agreed.

PR #467, opened 2020-04-12
Correct "Lights" paragraph in "What's next?" list, adjust numbering.

shaunplee:
In the "What's next?" list in Chapter 13, it seems like the paragraphs:

> 1. Lights. You can do this explicitly, by sending shadow rays to lights. Or it can be done implicitly by making some objects emit light,
> 2. Biasing scattered rays toward them, and then downweighting those rays to cancel out the bias. Both work. I am in the minority in favoring the latter approach.

should actually be part of the same paragraph. Combining these paragraphs also means renumbering the remaining paragraphs.


trevordblack:
Please create a separate issue.

Also, as far as "What's Next" we're in the middle of figuring out how we want to organize all of them together. See #407 

If you have any ideas, feel free to throw them in there

shaunplee:
Ok. New issue opened #472

books/RayTracingInOneWeekend.html
```diff
@@ -2811,25 +2811,24 @@
 You now have a cool ray tracer! What next?
 
   1. Lights. You can do this explicitly, by sending shadow rays to lights. Or it can be done
```
hollasch, line 2813: Comma before conjunction: “... sending shadow ray to lights, or it can be done ...”.

shaunplee: Suggesting that we use a semicolon instead of a comma to join these sentences, in order to more clearly separate the "explicit" and "implicit" approaches to lights.

books/RayTracingInOneWeekend.html
```diff
@@ -2810,7 +2810,7 @@
 
 You now have a cool ray tracer! What next?
 
-  1. Lights. You can do this explicitly, by sending shadow rays to lights. Or it can be done
+  1. Lights. You can do this explicitly, by sending shadow rays to lights; or it can be done
```
hollasch, line 2813: `,`, not `;`.

books/RayTracingInOneWeekend.html
```diff
@@ -2810,26 +2810,25 @@
 
 You now have a cool ray tracer! What next?
 
-  1. Lights. You can do this explicitly, by sending shadow rays to lights. Or it can be done
-     implicitly by making some objects emit light,
+  1. Lights. You can do this explicitly, by sending shadow rays to lights; or it can be done
```
hollasch, line 2813: Comma (`,`), not a semi-colon (`;`).

PR #474, opened 2020-04-13
Fix an image-url error in book2

### LINKED ISSUE #473: Image url wrong In Book2
https://github.com/RayTracing/raytracing.github.io/blob/51112291826e606047c6bbc3e0fadf3a2386ff73/books/RayTracingTheNextWeek.html#L2110

That should be img.cornell-first.jpg

https://github.com/RayTracing/raytracing.github.io/blob/51112291826e606047c6bbc3e0fadf3a2386ff73/books/RayTracingTheNextWeek.html#L767

We don't use qsort anymore

I've created a PR. It's first time for me. Did I make it in right way?

zgxmy:
Resolves #473

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2245,7 +2245,7 @@ We get:
 
   <div class="render">
 
-  ![Empty Cornell box](../images/img.cornell-empty.jpg)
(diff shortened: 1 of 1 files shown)


oxine:
May we should remove the flip_face part instead of changing the picture

PR #487, opened 2020-04-18
Update image generation code

hollasch:
Generally, define aspect ratio up front, then derive image height from
that and the image width.

This change also fixes an issue where the sampling was slightly off due
to an off-by-one error when computing the image basis vectors.

Finally, some naming and commenting tweaks to `aspect_ratio` and `vfov`.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -466,9 +466,9 @@ for now because we’ll add antialiasing later):
     }
 
     int main() {
-        const int image_width = 768;
-        const auto aspect_ratio = 9.0 / 16.0;
-        const int image_height = static_cast<int>(aspect_ratio * image_width);
+        const auto aspect_ratio = 16.0 / 9.0;
+        const int image_width = 384;
+        const int image_height = static_cast<int>(image_width / aspect_ratio);
 
         std::cout << "P3\n" << image_width << " " << image_height << "\n255\n";
 
@@ -1180,8 +1180,9 @@ And the new main:
     }
 
     int main() {
-        const int image_width = 200;
-        const int image_height = 100;
+        const auto aspect_ratio = 16.0 / 9.0;
+        const int image_width = 384;
+        const int image_height = static_cast<int>(image_width / aspect_ratio);
 
         std::cout << "P3\n" << image_width << ' ' << image_height << "\n255\n";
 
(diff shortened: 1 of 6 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -466,9 +466,9 @@
     }
 
     int main() {
-        const int image_width = 768;
-        const auto aspect_ratio = 9.0 / 16.0;
-        const int image_height = static_cast<int>(aspect_ratio * image_width);
+        const auto aspect_ratio = 16.0 / 9.0;
```
dafhi, line 469: i like this idea

books/RayTracingInOneWeekend.html
```diff
@@ -466,9 +466,9 @@
...
+        const int image_width = 384;
```
dafhi, line 470: image height * aspect seems more intuitive

books/RayTracingInOneWeekend.html
```diff
@@ -2563,14 +2567,14 @@
         public:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
             camera(
-                double vfov, // top to bottom, in degrees
-                double aspect
+                double vfov, // vertical field-of-view in degrees
```
dafhi, line 2570: good

books/RayTracingInOneWeekend.html
hollasch: W/H is consistent with the aspect ratio we've been using for the camera code. Specifying the single image size parameter as width feels more natural as it's the maximum dimension, and text also typically has a sensible bounded and known width (while height is varying).

books/RayTracingTheNextWeek.html
```diff
@@ -109,9 +109,10 @@
         public:
             camera(
                 point3 lookfrom, point3 lookat, vec3 vup,
-                double vfov, // top to bottom, in degrees
+                double vfov, // vertical field-of-view in degrees
+                double aspect_ratio, double aperture, double focus_dist,
```
dafhi, line 113: double double double  o.O

src/InOneWeekend/main.cc
```diff
@@ -113,8 +113,8 @@ int main() {
         for (int i = 0; i < image_width; ++i) {
             color pixel_color;
             for (int s = 0; s < samples_per_pixel; ++s) {
-                auto u = (i + random_double()) / image_width;
-                auto v = (j + random_double()) / image_height;
+                auto u = (i + random_double()) / (image_width-1);
```
dafhi, line 116: not off-by-1 adept atm

hollasch: ?

dafhi: i may analyze this later.  you're probably correct here.

dafhi: just did a mental visualization.  if image width was 1, width-1 as a divisor would not sample correctly amirite?  think im rite

hollasch: Vectors `u` and `v` are mapped across the window to the [0,1] interval. For a 2×2 image, the old code you give you vectors from 0 to 0.5. For a 3×3 image, this would give you a vector from 0 to 0.67.

With a 1×1 image, you don't need a vector to "span" the pixel, as a single pixel is a point. The width of a 1×1 image is thus zero. The fact that every image larger than 1×1 can't have an accurate normalized (unit) width or height reinforces this.

hollasch: Related reading: _A Pixel Is Not A Little Square, A Pixel Is Not A Little Square, A Pixel Is Not A Little Square!_ → http://alvyray.com/Memos/CG/Microsoft/6_pixel.pdf

PR #507, opened 2020-05-02
Rendered images use Markdeep class=pixel

### LINKED ISSUE #498: Rendered images should use Markdeep `class=pixel`
Currently, we're wrapping rendered images in `div.render`. We don't need to do that, as Markdeep already has a pixel image type built in.

Instead of this:
```
<div class="render">

![A simple red sphere](../images/img.red-sphere.png)

</div>
```

we should just use this:
```
![A simple red sphere](../images/img.red-sphere.png class=pixel)
```

hollasch:
Markdeep provides a `class=pixel` attribute for included images in order
to render with a point-sampled result. I didn't realize this when I
originally created CSS styling and an enclosing `div` element to do this
before. Switching the prior styling to now use Markdeep's feature.

Resolves #498


trevordblack:
Update Changelog

PR #510, opened 2020-05-02
Move vec3::write_color to standalone helper

### LINKED ISSUE #502: Remove vec3::write_color as member function, change to helper function
I think there is merit in making vec3 be a strict subset of functionality provided by nearly every other linear algebra package and the analogous classes built into various graphics languages and toolkits. This would allow users to easily substitute an alternate implementation that they might prefer -- one they are already using in other projects, or one in common use at their company, or one they previously wrote themselves.

There are currently a few minor impediments to this in the class as presented: the `write_color(std::ostream &out)` method, and later the `vec3::random()` methods. Those can perhaps be moved out of the vec3 class and into standalone functions.

hollasch:
This change introduces a new common project header: color.h. Updated
books and source code.

Resolves #502

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Change Log -- Ray Tracing in One Weekend
 ====================================================================================================
 
-----------------------------------------------------------------------------------------------------
+
 # v3.1.0 (in progress)
 
 ### Common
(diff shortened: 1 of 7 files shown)


trevordblack:
Why did this get it's own header file? Why didn't it just go into rtweekend.h?

hollasch:
That was my first move. However, since this function depends on `vec3.h`, and `rtweekend.h` is expected to be the first include, and also includes `vec3.h`, things got into a weird cycle. I wrestled with putting the includes in the middle of `rtweekend.h`, but that broke the simple cascade of dependencies. Eventually I realized that putting this in its own header with a clear dependency hierarchy was the correct choice.

PR #515, opened 2020-05-05
Move Texture Coordinates for Spheres from Ch 6 to Ch4 to define (u,v)

shaunplee:
Replaces #497 (I couldn't figure out how to change an existing pull request to target a different branch), addressing #496.

PR #518, opened 2020-05-06
Move Texture Coordinates for Spheres from Ch 6 to Ch4 to define (u,v).

### LINKED ISSUE #496: TheNextWeek: Ch. 4: Solid Textures - update sphere::hit() based on change to hit_record
Chapter 4, skips over a bit, explaining how surface coordinates `(u,v)` are computed for a sphere.

In particular, Listing 20 updates `hit_record` to add surface coordinates `double u` and `double v`:

    struct hit_record {
      vec3 p;
      vec3 normal;
      shared_ptr<material> mat_ptr;
      double t;
      double u;
      double v;
      bool front_face;
      ...

Two dimensional coordinates (u,v) correspond to the location at which a ray hit the (2D) surface of an object (sphere).

Listing 21 then updates `scatter` to pass in the `rec.u` and `rec.v` values from `hit_record rec`:

            attenuation = albedo->value(rec.u, rec.v, rec.p);

However, Chapter 4 does not explain how coordinates `(u,v)` are calculated and set in the `hit_record`, presumably when `sphere::hit()` is called with an incoming ray.

Skipping ahead to Chapter 6, `get_sphere_uv()` in Listing 42 seems to solve this problem, which is what I eventually added to my `hit()` for my updates for Chapter 4 to compile. It produces a plausible result:

checkeredFloor

hollasch:
Resolves #496

This is a rebased version of Shaun Lee's fix (#515).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Change Log -- Ray Tracing in One Weekend
 ### _In One Weekend_
 - Change: The C++ `<random>` version of `random_double()` no longer depends on `<functional>` header
 
+### _The Next Week_
+- Fix: Introduce "Texture Coordinates for Spheres" in Chapter 4 to support (u,v) coordinates in
+  `hit_record` (#496)
+
 
 ----------------------------------------------------------------------------------------------------
 # v3.1.0 (2020-05-03)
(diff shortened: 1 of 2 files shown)


shaunplee:
Looks good to me! Thanks!

Issue #533, opened 2020-05-14
`get_sphere_uv` does not correspond to its explanation


D-K-E:
The function `get_sphere_uv` in NextWeek listing 42 at section Texture 6.1, is given as 
```c++
void get_sphere_uv(const point3& p, double& u, double& v) {
    auto phi = atan2(p.z(), p.x());
    auto theta = asin(p.y());
    u = 1-(phi + pi) / (2*pi);
    v = (theta + pi/2) / pi;
}
```
whereas in the formula given in the explanation is:

```
To compute θ and ϕ, for a given hitpoint, the formula for 
spherical coordinates of a unit radius sphere on the origin is:

x=cos(ϕ)cos(θ)
y=sin(ϕ)cos(θ)
z=sin(θ)

We need to invert that. Because of the lovely <cmath> 
function atan2() (which takes any number proportional 
to sine and cosine and returns the angle), we can pass in x
and y (the cos(θ) cancel):

ϕ=atan2(y,x)

The atan2
returns values in the range −π to π, so we need to 
take a little care there. The θ is more straightforward:

θ=asin(z)

which returns numbers in the range −π/2
to π/2. 
```
The problem also exist in `dev-major` branch.

hollasch:
Thanks for reporting! If you would like, feel free to submit a pull request of your suggested changes to the `dev-patch` branch.

D-K-E:
Sure, very solid work by the way. I am also porting the Weekend version to opengl with compute shaders in [here](https://github.com/D-K-E/raytracing-gl)

D-K-E:
The PR that was suppose to change the code for implementing the formula above is closed now, because most of the sources I have cited there favor the existing implementation. The wording in this section might be improved to show how we arrive to the implementation of `get_sphere_uv` function, instead of using unit sphere at origin to represent the problem.

trevordblack:
Suggest punting to 3.2

PR #534, opened 2020-05-14
fixes get_sphere_uv issue

D-K-E:
Fixes the `get_sphere_uv` function mentioned in #533

### DIFF
--- a/src/TheNextWeek/hittable.h
+++ b/src/TheNextWeek/hittable.h
@@ -19,8 +19,8 @@
 class material;
 
 void get_sphere_uv(const point3& p, double& u, double& v) {
-    auto phi = atan2(p.z(), p.x());
-    auto theta = asin(p.y());
+    auto phi = atan2(p.y(), p.x());
+    auto theta = asin(p.z());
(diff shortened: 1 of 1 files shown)


hackmad:
Although the code is taking us through the derivation it is assuming which axis is unwrapping along the horizontal and vertical texture direction. I wonder if the coordinate system plays some part in it since we assume z points into the screen/image. 

If I make this same change, the result of mapping the earth texture is this:



However with the code presented in the book I get the correct result:

D-K-E:
I took the explanation in the book as the basis for the `get_sphere_uv` code.
I should have checked other sources about uv mapping to sphere. The sources below all
favor unanimously the existing code, so I am closing this pr. 

http://www.cse.msu.edu/~cse872/tutorial4.html
https://en.wikipedia.org/wiki/UV_mapping
https://stackoverflow.com/questions/19357290/convert-3d-point-on-sphere-to-uv-coordinate
https://gamedev.stackexchange.com/a/114416

However the explanation in the book is a little misleading, since the formula shown and the code uses the same variable names.

PR #536, opened 2020-05-15
Update camera code

### LINKED ISSUE #527: Fix camera code and associated text to account for new aspect ratio calculations
After moving the initial image generation from a resolution of 200×100 to a larger resolution, we forgot to update the associated text.

In addition, in the move to a computed image resolution height based on aspect ratio, the associated viewport does not receive the same treatment, but instead uses a horizontal vector 4 units in length, and a vertical vector 2 units in length (matching the original 200:100 aspect ratio). We need to update the viewport computation throughout the books, and in source code.

hollasch:
In the prior code, changes to camera variables would require associated
changes to a number of other related variables, in ways that were not
apparent. This change uses variables a more consistent manner, where
changing any particular variable should yield consistent results in
other values.

In addition, this provides a more consistent series of iterative steps
as the camera develops through the text.

Resolves #527

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -466,12 +466,21 @@ simple camera for getting the code up and running. I also make a simple `ray_col
 that returns the color of the background (a simple gradient).
 
 I’ve often gotten into trouble using square images for debugging because I transpose $x$ and $y$ too
-often, so I’ll stick with a 200×100 image. I’ll put the “eye” (or camera center if you think of a
-camera) at $(0,0,0)$. I will have the y-axis go up, and the x-axis to the right. In order to respect
-the convention of a right handed coordinate system, into the screen is the negative z-axis. I will
-traverse the screen from the lower left hand corner, and use two offset vectors along the screen
-sides to move the ray endpoint across the screen. Note that I do not make the ray direction a unit
-length vector because I think not doing that makes for simpler and slightly faster code.
+often, so I’ll use a non-square image. For now we'll use a 16:9 aspect ratio, since that's so
+common.
+
+In addition to setting up the pixel dimensions for the rendered image, we also need to set up a
+virtual viewport through which to pass our scene rays. For the standard square pixel spacing, the
+viewport's aspect ratio should be the same as our rendered image. We'll just pick a viewport two
+units in height. Ultimately, changing the scale of the viewport (while holding the focal distance
+constant) is equivalent to changing the viewing angle, or “zoom” of the image.
+
+I’ll put the “eye” (or camera center if you think of a camera) at $(0,0,0)$. I will have the y-axis
+go up, and the x-axis to the right. In order to respect the convention of a right handed coordinate
+system, into the screen is the negative z-axis. I will traverse the screen from the lower left hand
+corner, and use two offset vectors along the screen sides to move the ray endpoint across the
+screen. Note that I do not make the ray direction a unit length vector because I think not doing
+that makes for simpler and slightly faster code.
 
   ![Figure [cam-geom]: Camera geometry](../images/fig.cam-geom.jpg)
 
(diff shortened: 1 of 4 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -466,12 +466,21 @@
...
-traverse the screen from the lower left hand corner, and use two offset vectors along the screen
-sides to move the ray endpoint across the screen. Note that I do not make the ray direction a unit
-length vector because I think not doing that makes for simpler and slightly faster code.
+often, so I’ll use a non-square image. For now we'll use a 16:9 aspect ratio, since that's so
+common.
+
+In addition to setting up the pixel dimensions for the rendered image, we also need to set up a
+virtual viewport through which to pass our scene rays. For the standard square pixel spacing, the
+viewport's aspect ratio should be the same as our rendered image. We'll just pick a viewport two
+units in height. Ultimately, changing the scale of the viewport (while holding the focal distance
```
dafhi, line 475: focal d hasn't been intro'd yet, complicates

trevordblack: Agree with David. Cut this sentence. If the user wants to play with the height, they'll intuit how focal length and zoom interact.

trevordblack: Also, you have it written `focal_distance` here when it's `focal_length` in the code below. I'm pretty sure those are the same thing. If this sentence is kept, those should be synced.

books/RayTracingInOneWeekend.html
```diff
@@ -497,19 +506,25 @@
...
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+
         for (int j = image_height-1; j >= 0; --j) {
             std::cerr << "\rScanlines remaining: " << j << ' ' << std::flush;
             for (int i = 0; i < image_width; ++i) {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
                 auto u = double(i) / (image_width-1);
                 auto v = double(j) / (image_height-1);
-                ray r(origin, lower_left_corner + u*horizontal + v*vertical);
+                ray r(origin, lower_left_corner + u*horizontal + v*vertical - origin);
```
trevordblack, line 527: OOOoooooh. NICE CATCH

books/RayTracingInOneWeekend.html
hollasch: The book has confusing terminology between "focus distance" and "focal distance" (or "focal length", which I think I prefer). The focus distance is the virtual plane at which objects appear in perfect focus. The focal length is the distance between the projection point and the projection plane. These are very different (in fact, they _must_ be).

I'll clarify the text.

trevordblack: Yeah, sorry, I just realized my mistake as well

trevordblack: I think we should keep things as `focal length` and `focus distance`.

hollasch: I need to be more careful when flinging terms.

trevordblack: Naaaaaaah

books/RayTracingInOneWeekend.html
```diff
@@ -2728,16 +2729,19 @@
...
-In a physical camera, the distance to that plane where things are in focus is controlled by the
-distance between the lens and the film/sensor. That is why you see the lens move relative to the
-camera when you change what is in focus (that may happen in your phone camera too, but the sensor
-moves). The “aperture” is a hole to control how big the lens is effectively. For a real camera, if
-you need more light you make the aperture bigger, and will get more defocus blur. For our virtual
-camera, we can have a perfect sensor and never need more light, so we only have an aperture when we
-want defocus blur.
+We call the distance between the projection point and the plane where everything is in perfect focus
+the _focus distance_. Be aware that the focus distance is not the same as the focal length -- the
+_focal length_ is the distance between the projection point and the image plane.
```
trevordblack, line 2737: You use projection plane instead of image plane. And then later in the book, it's written as:

```
We don’t need to simulate any of the inside of the camera. For the purposes of rendering an image outside the camera, that would be unnecessary complexity. Instead, I usually start rays from the surface of the lens, and send them toward a virtual film plane, by finding the projection of the film on the plane that is in focus (at the distance focus_dist). 
```

"virtual film plane"

It doesn't need to be rigorously consistent within the book. And, what we're referring to here (image plane/film plane/projection plane) has all three terms used interchangeably in the literature.
I'm fine with it as you've written, since I think it's still reasonably clear.

hollasch: Hmmm, excellent point. I dislike "film plane", and would prefer either "projection plane" or "image plane", but lean toward the latter. Reviewing...

books/RayTracingInOneWeekend.html
```diff
@@ -2759,8 +2759,8 @@
 <div class="together">
 We don’t need to simulate any of the inside of the camera. For the purposes of rendering an image
 outside the camera, that would be unnecessary complexity. Instead, I usually start rays from the
-surface of the lens, and send them toward a virtual film plane, by finding the projection of the
-film on the plane that is in focus (at the distance `focus_dist`).
+lens aperture, and send them toward the projection plane (`focus_dist` away), where everything on
```
trevordblack, line 2762: the `focus_dist` is the distance from the focus plane to the projection point. NOT the focus plane to projection plane

trevordblack: Actually, I don't understand this sentence...

hollasch: I can swap out “projection plane” for “focus plane”. In this model, they are both one and the same, as are the focal length and focus distance. Also note that there's no “projection point” any more, but a projection _disk_, though I guess you could say the disk of projection points.

hollasch: Though I suppose you could say that the image/projection plane still exists, but where would it be? Behind the aperture? I think it's basically done away with in this model. As another argument that the image/projection plane and the focus plane have been unified, consider that the image sampling happens directly on the (what would you call it?) focus/image rectangle.

A figure would definitely help here.

hollasch: Actually, I was looking at the text and not the figures. What's really happened here is a bit of slight-of-hand. The image-in-front-of-projection-point model has been swapped out for the image-behind-the-projection-point, and then the back half is immediately dropped.

hollasch: Heh, the original is still messed up — it describes sending rays “toward the virtual film plane”, but in the figures, the “film” is behind the lens. Will rework again.

PR #615, opened 2020-05-23
Correction in the rays section

argoopjmc:
in the main.cc along with ray.h , color.h is also needed because of write_color ( this is for the weekend section)


hollasch:
This is going into the wrong branch. Please see CONTRIBUTING.md.

PR #631, opened 2020-05-31
Consolidated source README into a single readme at directory root.

trevordblack:
Should this change go in the CHANGELOG?


README.md
```diff
@@ -29,19 +34,59 @@ refactoring, adding new functionality where it made sense, and generally trying
...
+Directory Structure
+-------------------
+The organization of this repository is meant to be simple and self-evident at a glance:
 
-Branches
----------
-The `master` branch contains the code at latest release. All ongoing development, with all of the
-latest changes, can be found in the `dev-patch`, `dev-minor`, and `dev-major` branches.
+### books/
+This folder contains the text of the books in the markdeep markup language. This folder also
```
hollasch, line 42: Coincidentally, I was thinking today about mixing the actual books (ray*.html) with supporting materials (markdeep.min.js, ackowledgments.md.html). The problem is that it's probably more useful to say that this directory contains the actual books, not just "the text of the books". The latter suggests that these files are used to build the final products, rather than that they are directly consumable as the books themselves.

How about saying something like "this folder contains the three ray tracing books (as HTML), plus supporting materials"?

README.md
```diff
@@ -21,6 +21,11 @@ versions, use the print function in your browser.
 
 News
 -----
+2020-May-5 — v3.1.0 is out! A smaller minor level patch to iron out a few of the larger things we
+wanted to change after a year of organization. The biggest change in the text is the separation of
+chapters into subchapters. There is no one biggest change in the source, but there are a large
+number of small and significant changes. 
+
 2020-March-23 — v3.0.0 is out! Following the major v2.0.0 release, we finally had the chance to dig
```
hollasch, line 29: Let's strike this section.

hollasch, line 29: Let's strike this old update.

trevordblack:
Should changes to the README go into the CHANGELOG?

PR #633, opened 2020-06-05
Fix source name

tyfkda:


hollasch:
Looks good. Please target dev-patch.

tyfkda:
@hollasch Thank you for your review!
I've changed target branch to dev-path.

hollasch:
@tyfkda — can you let me know your real name, so I can add you to the book credits? Thanks.

tyfkda:
@hollasch Thank you for your kindness, but my change was very small so I don't think my name should be added to the credit.

hollasch:
Well, ok, but that means you need to add a bigger change then. `:D`

Issue #634, opened 2020-06-07
Missing line in OneWeekend Listing 18 


srayuws:
Current code [link](https://github.com/RayTracing/raytracing.github.io/blob/957b20a1694d7028e089a3788c32701d789266d6/books/RayTracingInOneWeekend.html#L972) :

```
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
    #ifndef HITTABLE_H
    #define HITTABLE_H

    #include "ray.h"

    struct hit_record {
        point3 p;
        vec3 normal;
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
        bool front_face;

        inline void set_face_normal(const ray& r, const vec3& outward_normal) {
            front_face = dot(r.direction(), outward_normal) < 0;
            normal = front_face ? outward_normal :-outward_normal;
        }
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
    };
```

The `double t` is missing in `struct hit_record`.

I can make a PR for this issue if no one else is doing so.

hollasch:
Fixed in #625.

PR #635, opened 2020-06-07
fix missing line in OneWeekend Listing 18 

srayuws:
fix the missing line #634


hollasch:
Thank you for the contribution, but it was fixed in #625.

PR #638, opened 2020-06-14
Book2 img updates

hollasch:


hollasch:
Ugh. As soon as I was ready to merge, I realized that I hadn't changed the book to reference the new PNGs instead of the old JPGs. Will add a new commit soon.

PR #641, opened 2020-06-18
acks: link to GitHub user pages where known

hollasch:
(Direct into master)

This change updates our acknowledgments, linking contributor real names to their corresponding public GitHub user pages where known.

In addition, I realized that we never added public thanks to Helen Hu for donating the RayTracing GitHub organization, so added that.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -4,8 +4,8 @@
 
 
                                    **Ray Tracing in One Weekend**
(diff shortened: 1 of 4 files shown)


trevordblack:
This is a really nice change

trevordblack:
Errr. Should this be into master?

hollasch:
Generally, updates to acknowledgments, general project READMEs, and administrative info go directly into the publish/release/master branch.

Issue #639, opened 2020-06-16
Suggestion: `override` for derived classes


D-K-E:
This might seem rather unnecessary, since if the reader is careful, overriding methods of the base class shouldn't be a problem. However I noticed that it took me quite a while to track down the bugs related to this problem. An `override` qualifier can add an extra layer of warning for such bugs.
I am basically suggesting to turn this:
```c++
class Material {
public:
  virtual bool scatter(const Ray &r_in, const HitRecord &rec,
                       ScatterRecord &srec) const {
    return false;
  }
};
class lambertian : public material {
public:
    virtual bool scatter(const ray &r_in, const hit_record &rec,
                       scatter_record &srec) const  {}
public:
  shared_ptr<Texture> albedo;
};
```
to this:
```c++
class Material {
public:
  virtual bool scatter(const Ray &r_in, const HitRecord &rec,
                       ScatterRecord &srec) const {
    return false;
  }
};
class lambertian : public material {
public:
    bool scatter(const ray &r_in, const hit_record &rec,
                       scatter_record &srec) const override {...}
public:
  shared_ptr<Texture> albedo;
};
```

hollasch:
I like it.

D-K-E:
I can do a PR if you would like to ?

hollasch:
Sure. Let's go ahead and drop this in the `dev-minor` branch for our v3.2.0 release.

You will need to update all code and all instances in the three books. In addition, please work with @trevordblack, as he is in the middle of running through book 2 for an image update pass, updating numerous code issues as he goes.

hollasch:
Also, I keep getting confused with real names versus GitHub user IDs — have you been added to the acknowledgements yet? (https://raytracing.github.io/books/acknowledgments.md.html). At this point, I think I should be linking a lot of these names to associated GitHub user pages.

D-K-E:
I tried adding the `override` to all code related to "In One Weekend" and it compiled. I'll wait for @trevordblack before proceeding with the other two books. Yes I'd been added to acknowledgements: Kaan Eraslan

trevordblack:
The big thing here is that the `override` keyword would need to be added to the text of the book as these functions are added in.

All code changes need to go into the book. We've not been diligent about this in the past, but going forward, that's our plan.

If you want to tread forward, you can try to add override to both the source and text, or you can just do the source, or you can just do the text.

[...]

PR #642, opened 2020-06-21
Override keyword added to all books and code

D-K-E:
"override" qualifier is added to all derived classes.
The qualifier is also added to the text of all the  books. I might have missed one or two but normally it should be there. 
A quick note: The `scatter` method of `isotropic` class does not override that of base class but it overloads it in the last book, so I have not touched upon it. 
I also don't have markdeep formatting library, so it is very much possible that the text of the book looks poorly formatted.

All the code compiles and no runtime error
during the execution.
I have only tested the three main executables:
inOneWeekend
theNextWeek
theRestOfYourLife

Related to #639

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -888,7 +888,8 @@ And here’s the sphere:
             sphere() {}
             sphere(point3 cen, double r) : center(cen), radius(r) {};
 
-            virtual bool hit(const ray& r, double tmin, double tmax, hit_record& rec) const;
+            bool hit(const ray& r, double tmin, double tmax, hit_record& rec)
+            const override;
 
         public:
             point3 center;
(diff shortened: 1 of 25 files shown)


trevordblack:
what's with all of the changes to the source code at the end of the diff? Was that your changes, or did you forget to `git pull`.

There's a lot in here that's not just `override`

trevordblack:
what's with all of the changes to the source code at the end of the diff? Was that your changes, or did you forget to `git pull`.

There's a lot in here that's not just `override`

D-K-E:
I assume the diff is due to formatting? I am using clang-format at each save, I have not noticed that it created such a huge diff. The same did not apply for the source text since I don't have a markdeep formatter. I did fetch and pull the latest change from dev-minor, and I really just added the `override` and deleted the `virtual` at the beginning of the method as I have promised in the issue.

hollasch:
Lots of indentation differences and the like, so more than just being out of sync.

I'd suggest cloning directly from the project repo (not a fork), creating a new `overrides` branch from `dev-minor`, and pushing and PR'ing those changes.

If you still want to work from a fork, you should probably create a feature branch to work in, instead of working directly in a fork of one of our integration branches (like `dev-minor`).

D-K-E:
All right. I'll rebase to dev-minor and do the commits in smaller chunks to an "override" branch to make it more digestable

PR #647, opened 2020-06-24
Add isotropic constructor taking color argument

### LINKED ISSUE #644: material.isotropic should have a constructor that takes a color value, like other materials
Such as `lambertian` and `diffuse_light`.

hollasch:
Resolves #644


hollasch:
It ain't always this easy.

PR #670, opened 2020-07-16
Overrides

### LINKED ISSUE #639: Suggestion: `override` for derived classes

hollasch:
Resolves #639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,18 @@ Change Log -- Ray Tracing in One Weekend
 # v3.2.0 (in progress)
 
 ### Common
+  - Bug: Found a bug in book 3 source `isotropic::scatter()` method. Commented out, using default
+    (as it was previously). (#669)
   - Removed: now that the code handles ray-surface intersection from either side, we no longer need
     the `flip_face` class, so we've deleted it from the text and from the code (#482, #270)
   - Change: Added header guards to the text of all three books whenever a new header file was
     introduced
+  - Change: Added `override` keywords throughout. This keyword marks a subclass method as one that
+    is intended to override a superclass method. It makes the code a bit easier to understand, and
+    ensures that your function is actually overriding the method you think it is. Which is good,
+    because it already caught an existing bug in _The Rest of Your Life_ source. This change
+    includes commenting out the book 3 `isotropic::scatter()` method, which was accidentally ignored
+    anyway. (#639, #669)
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -4,10 +4,18 @@ Change Log -- Ray Tracing in One Weekend
...
+  - Bug: Found a bug in book 3 source `isotropic::scatter()` method. Commented out, using default
+    (as it was previously). (#669)
   - Removed: now that the code handles ray-surface intersection from either side, we no longer need
     the `flip_face` class, so we've deleted it from the text and from the code (#482, #270)
   - Change: Added header guards to the text of all three books whenever a new header file was
     introduced
+  - Change: Added `override` keywords throughout. This keyword marks a subclass method as one that
+    is intended to override a superclass method. It makes the code a bit easier to understand, and
+    ensures that your function is actually overriding the method you think it is. Which is good,
+    because it already caught an existing bug in _The Rest of Your Life_ source. This change
```
trevordblack, line 16: Is this comment necessary/good?

hollasch: It's sort of like the *BUGS* section in a manpage. We've identified a new bug, and this release does _something_ about it, but doesn't fix it, and points to the owning issue. If you take this fix and wonder what's up, we should document the situation.

trevordblack: Makes sense to me

PR #678, opened 2020-07-21
Tmin tmax

### LINKED ISSUE #418: Inconsistent use of t0 and tmin
In geometric objects with hit, bounding_box,

the function signatures between objects vary between using t0,t1 and t_min, t_max

hollasch:
Resolves #418

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -887,7 +887,7 @@ And here’s the sphere:
     class sphere : public hittable {
         public:
             sphere() {}
-            sphere(point3 cen, double r) : center(cen), radius(r) {};
+            sphere(point3 ctr, double r) : center(ctr), radius(r) {};
 
             virtual bool hit(
                 const ray& r, double tmin, double tmax, hit_record& rec) const override;
@@ -1995,8 +1995,8 @@ within `hit_record`. See the highlighted lines below:
         public:
             sphere() {}
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            sphere(point3 cen, double r, shared_ptr<material> m)
-                : center(cen), radius(r), mat_ptr(m) {};
+            sphere(point3 ctr, double r, shared_ptr<material> m)
+                : center(ctr), radius(r), mat_ptr(m) {};
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
 
             virtual bool hit(
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -866,7 +866,8 @@ the abstract class:
 
     class hittable {
         public:
-            virtual bool hit(const ray& r, double t_min, double t_max, hit_record& rec) const = 0;
+            virtual bool hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec)
+                const = 0;
     };
 
     #endif
@@ -889,15 +890,15 @@ And here’s the sphere:
             sphere() {}
             sphere(point3 ctr, double r) : center(ctr), radius(r) {};
 
-            virtual bool hit(
-                const ray& r, double tmin, double tmax, hit_record& rec) const override;
+            virtual bool hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec)
+                const override;
 
         public:
             point3 center;
             double radius;
     };
 
-    bool sphere::hit(const ray& r, double t_min, double t_max, hit_record& rec) const {
+    bool sphere::hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec) const {
         vec3 oc = r.origin() - center;
         auto a = r.direction().length_squared();
         auto half_b = dot(oc, r.direction());
@@ -908,7 +909,7 @@ And here’s the sphere:
             auto root = sqrt(discriminant);
 
             auto temp = (-half_b - root) / a;
-            if (temp < t_max && temp > t_min) {
+            if (ray_tmin < temp && temp < ray_tmax) {
                 rec.t = temp;
                 rec.p = r.at(rec.t);
                 rec.normal = (rec.p - center) / radius;
@@ -916,7 +917,7 @@ And here’s the sphere:
             }
 
             temp = (-half_b + root) / a;
-            if (temp < t_max && temp > t_min) {
+            if (ray_tmin < temp && temp < ray_tmax) {
                 rec.t = temp;
                 rec.p = r.at(rec.t);
                 rec.normal = (rec.p - center) / radius;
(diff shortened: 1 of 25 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -392,10 +392,14 @@
         auto dist_to_focus = 10.0;
         auto aperture = 0.1;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
+        const auto time_start = 0.0;
+        const auto time_end = 1.0;
```
trevordblack, line 398: This belongs in a separate issue, but we should consider moving `aperture`, `dist_to_focus`, `time_start`, `time_end` to above the scene switch. After being introduced, we never really change these values, but _someone_ probably does

books/RayTracingTheNextWeek.html
```diff
@@ -1254,8 +1254,8 @@
         public:
             checker_texture() {}
 
-            checker_texture(shared_ptr<texture> t0, shared_ptr<texture> t1)
-                : even(t0), odd(t1) {}
+            checker_texture(shared_ptr<texture> texture_even, shared_ptr<texture> texture_odd)
+                : even(texture_even), odd(texture_odd) {}
```
trevordblack, line 1268: Oooooooohh, I _like_ that!

books/RayTracingTheNextWeek.html
```diff
@@ -190,8 +190,8 @@
         public:
             moving_sphere() {}
             moving_sphere(
-                point3 cen0, point3 cen1, double t0, double t1, double r, shared_ptr<material> m)
-                : center0(cen0), center1(cen1), time0(t0), time1(t1), radius(r), mat_ptr(m)
+                point3 ctr0, point3 ctr1, double t0, double t1, double r, shared_ptr<material> m)
+                : center0(ctr0), center1(ctr1), time0(t0), time1(t1), radius(r), mat_ptr(m)
```
trevordblack: Hmmmmm....

I don't want the member variables to be `center0`, `center1`, `time0`, `time1`.
We're already removing the `0` and `1` suffix from other places in the code.
However, `center_start`, `center_end`, `time_start`, `time_end` are potentially a _BAD_ idea, because someone might be confused into thinking that `time_start` and `time_end` are the same values as the camera's `time_start` and `time_end` when they very much might not be. The same goes for `center_start` and `center_end` being confused as being center location at time of `time_start` and `time_end`.

The sphere also just linearly interpolates off to infinity in both directions so the member variables are not `_start` and `_end` per se. I guess I'm fine leaving them as `*0` and `*1`, but I wonder if we'd be better off something else.

Wikipedia seems to think suffix with `*0` and `*1` are acceptable:
https://en.wikipedia.org/wiki/Linear_interpolation

trevordblack:
WAIT. Why doesn't RestOfYourLife have movingsphere????
It's just _gone_

books/RayTracingTheNextWeek.html
```diff
@@ -190,9 +190,12 @@
...
             moving_sphere() {}
             moving_sphere(
-                point3 ctr0, point3 ctr1, double t0, double t1, double r, shared_ptr<material> m)
-                : center0(ctr0), center1(ctr1), time0(t0), time1(t1), radius(r), mat_ptr(m)
-            {};
+                point3 ctr0, point3 ctr1, double r, shared_ptr<material> m,
+                double time_start, double time_end)
+                :
+                center0(ctr0), center1(ctr1), radius(r), mat_ptr(m),
+                time0(time_start), time1(time_end)
```
trevordblack, line 197: See earlier comments re moving_sphere member variables (`time_start` and `time_end` param names may be a bad idea)

books/RayTracingTheNextWeek.html
```diff
@@ -333,7 +336,7 @@
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
                         auto center2 = center + vec3(0, random_double(0,.5), 0);
                         world.add(make_shared<moving_sphere>(
-                            center, center2, 0.0, 1.0, 0.2, sphere_material));
+                            center, center2, 0.2, sphere_material, 0.0, 1.0));
```
trevordblack, line 338: I really like the change of cnstr param order

books/RayTracingInOneWeekend.html
```diff
@@ -866,7 +866,8 @@
 
     class hittable {
         public:
-            virtual bool hit(const ray& r, double t_min, double t_max, hit_record& rec) const = 0;
+            virtual bool hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec)
```
trevordblack, line 869: This is just a simple but outrageously effective change.

books/RayTracingTheNextWeek.html
```diff
@@ -710,7 +710,8 @@
```
trevordblack, line 711: Found in the text:
```
In addition, moving objects will have a bounding box that encloses the object for the entire time interval [time1,time2]. 
```

books/RayTracingTheNextWeek.html
```diff
@@ -710,7 +710,8 @@
 
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            virtual bool bounding_box(double t0, double t1, aabb& output_box) const = 0;
+            virtual bool bounding_box(double time_start, double time_end, aabb& output_box)
```
trevordblack, line 713: ...
`time_start` `time_end` business again.
It might not be too big of a problem, but these do represent independent values to camera `time_start` and `time_end`

trevordblack: Maybe just go ahead and change the camera's values in main to `camera_time_start`, `camera_time_stop`, or maybe `shutter_time_start`, `shutter_time_stop`

trevordblack: Which, oh, _jesus_.
This `time_start` and `time_stop` values can also be independent from the `time0` and `time1` values of `moving_sphere`. But are probably (and always in the text) the same values.
"Fun" fact: if the bounding box's `time_start` is -inf and `time_end if +inf, the bounding encompasses all of space.

trevordblack:
This PR is absolutely filled with great changes, but see inline

books/RayTracingTheNextWeek.html
hollasch: Good idea. I'll add that.

books/RayTracingTheNextWeek.html
hollasch: Seems unavoidable. Could be `time_start_center` and `time_end_center`, or `time0_center` and `time1_center`. Wordy, but definitely clear.

hollasch:
src/TheRestOfYourLife/moving_sphere.h was removed in ac5409b6fb31c45e42ec8c7a315f7aecf09e4e3f, since it is unused.

books/RayTracingTheNextWeek.html
hollasch: Good catch.

books/RayTracingTheNextWeek.html
hollasch: The bbox seems reasonably infinite in that latter case. No surprise there.

hollasch: I'm ok with `camera_time_{start,stop}`.

hollasch: I think our views are different though. You seem to consider `time_start` and `time_end` as a sort of global time for all things, like `world_start` or `world_stop`. I just consider it a context-sensitive name, like `duration` or `length` or `magnitude`.

trevordblack: That's fair. I'm just not used to many contexts where objects don't have sync'd times

books/RayTracingTheNextWeek.html
hollasch: Done.

trevordblack:
Looks good to me.

PR #685, opened 2020-08-08
Remove unused vars in initial perlin::noise()

### LINKED ISSUE #684: TheNextWeek: Unused variables in Perlin noise
In listing 29, function `noise()`, we first calculate `u`, `v` and `w` from `p.x`, `p.y` and `p.z` but then do nothing with them.

```cpp
        double noise(const point3& p) const {
            auto u = p.x() - floor(p.x());
            auto v = p.y() - floor(p.y());
            auto w = p.z() - floor(p.z());

            auto i = static_cast<int>(4*p.x()) & 255;
            auto j = static_cast<int>(4*p.y()) & 255;
            auto k = static_cast<int>(4*p.z()) & 255;

            return ranfloat[perm_x[i] ^ perm_y[j] ^ perm_z[k]];
        }
```

hollasch:
Resolves #684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Change Log -- Ray Tracing in One Weekend
 ### _In One Weekend_
   - Fix: replace old anti-alias result image with before-and-after image (#679)
 
+### _The Next Week_
+  - Delete: remove unused u,v,w variables in initial `perlin::noise()` function (#684)
+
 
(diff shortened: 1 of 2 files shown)


trevordblack:
The highlight in listing 33 needs to be adjusted

hollasch:
Note the last statement:
```
return trilinear_interp(c, u, v, w);
```

PR #689, opened 2020-08-09
Remove unnecesary include of cstdlib

### LINKED ISSUE #687: Book1 Listing 23: cstdlib is not necessary yet
As `rand()` is not used at this stage, I think `#include <cstdlib>` can be removed.

https://github.com/RayTracing/raytracing.github.io/blob/7b92650a045a887391bf8148af86c5e6c46c7a9e/books/RayTracingInOneWeekend.html#L1186

hollasch:
It's not needed until we use the std `rand()` function.

Resolves #687


hollasch:
Tested & verified with stub file.

hollasch:
Merge conflicts; superseded by #703

PR #705, opened 2020-08-20
Fix book 2 listing 51 missing hittable, rtweekend includes

### LINKED ISSUE #693: Book2: Listing 51: include hittable.h
`hittable.h` should be included as it inherits hittable.

rjkilpatrick:
Resolves #693


hollasch:
Thanks!

PR #708, opened 2020-08-22
Fix: Book2: Listing 70 (#694)

### LINKED ISSUE #694: Book2: Listing 70: fuzz value and include bvh.h
- `bvh.h` should be included as it uses`bvh_node`.
- Fuzz value for the metal sphere is set to 10.0, but 1.0 would be better, which is the maximum value.

https://github.com/RayTracing/raytracing.github.io/blob/7b92650a045a887391bf8148af86c5e6c46c7a9e/books/RayTracingTheNextWeek.html#L3279-L3281

https://github.com/RayTracing/raytracing.github.io/blob/7b92650a045a887391bf8148af86c5e6c46c7a9e/src/TheNextWeek/material.h#L60

estshorter:
Resolves #694


trevordblack (approved):

estshorter:
@hollasch Would you merge this and #709?

PR #724, opened 2020-09-09
Fix: Book2: Listing 33, 39, perlin.h

### LINKED ISSUE #722: `trilinear_interp` is inconsistent between listings and src
When we are introduced to `trilinear_interp`, it is a function outside of the `perlin` class (listing 33):
https://github.com/RayTracing/raytracing.github.io/blob/80a955b23d7c95844bccabb6127abd4733d18a28/books/RayTracingTheNextWeek.html#L1597-L1611

However, when later amended (listing 39), it is moved into the class and marked with `const` (only valid for class member functions).
https://github.com/RayTracing/raytracing.github.io/blob/80a955b23d7c95844bccabb6127abd4733d18a28/books/RayTracingTheNextWeek.html#L1842-L1865

In the `src` directory, the `const` is missing but it's given a static 
https://github.com/RayTracing/raytracing.github.io/blob/80a955b23d7c95844bccabb6127abd4733d18a28/src/common/perlin.h#L100-L116

I propose changing [listing 33](https://github.com/RayTracing/raytracing.github.io/blob/80a955b23d7c95844bccabb6127abd4733d18a28/books/RayTracingTheNextWeek.html#L1597-L1611) to a private member function for consistency.

Can someone advise about which function signature (i.e. `static` and/or `const`) should be the authoritative version? I can then put a PR together

rjkilpatrick:
Add consistent function signature for `perlin_interp`.
Also adds highlighting to changes between listing 33 and 39

Resolves #722

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -1568,21 +1568,6 @@ To make it smooth, we can linearly interpolate:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
 
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-    inline double trilinear_interp(double c[2][2][2], double u, double v, double w) {
-        auto accum = 0.0;
-        for (int i=0; i < 2; i++)
-            for (int j=0; j < 2; j++)
-                for (int k=0; k < 2; k++)
-                    accum += (i*u + (1-i)*(1-u))*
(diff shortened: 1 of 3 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -1818,25 +1818,27 @@
...
-
+                for (int j=0; j < 2; j++)
+                for (int k=0; k < 2; k++) {
+                    vec3 weight_v(u-i, v-j, w-k);
+                    accum += (i*uu + (1-i)*(1-uu))
+                    * (j*vv + (1-j)*(1-vv))
+                    * (k*ww + (1-k)*(1-ww))
+                    * dot(c[i][j][k], weight_v);
+                }
+
```
hollasch, line 1837: Indentation corrupted.

hollasch (changes requested):

books/RayTracingTheNextWeek.html
rjkilpatrick: Sorry about that, thanks for picking it up

hollasch (approved):

PR #726, opened 2020-09-11
Update local Markdeep library to v1.11

### LINKED ISSUE #712: Upgrade to Markdeep 1.11
Also note new official location: https://morgan3d.github.io/markdeep/latest/markdeep.min.js

hollasch:
This affects offline book browsing only.

Resolves #712


trevordblack:
Looks Good To Me.

What is our preference for updates in the future?
I can't imagine there being any huge, breaking, Markdeep changes any time soon.
But it's something we might want to think about.

It looks like our book source points to:
```
<script src="https://casual-effects.com/markdeep/latest/markdeep.min.js"></script>
```
The `latest` markdeep branch. Do we want to specify a specific markdeep link and update and markdeep updates?

trevordblack (approved):

hollasch:
My plan for now is to just use latest, and then address any future issues if they arise. I've never had a problem with Markdeep updates before, so this is fairly safe. For offline browsing (where Markdeep pulls from the local library copy), we can just update occasionally or as needed.

PR #728, opened 2020-09-11
Fix problems with material::scatter() methods

### LINKED ISSUE #133: Missing ray.time in scatter function
It seems that you are forgotten to pass the r_in.time into the constructor of the scattered ray in some of  materials.

motion_blur_error

hollasch:
- In book 1, the code for material::dielectric::scatter() divereged from
  the source code.

- In book 2, we neglected to include the code changes to handle ray time
  for material::metal and material::dielectric.

Resolves #133


trevordblack:
The text in the book having
```
            virtual bool scatter(
                const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered
            ) const override {
                attenuation = color(1.0, 1.0, 1.0);
                double etai_over_etat = rec.front_face ? (1.0 / ref_idx) : ref_idx;

                vec3 unit_direction = unit_vector(r_in.direction());
                double cos_theta = fmin(dot(-unit_direction, rec.normal), 1.0);
                double sin_theta = sqrt(1.0 - cos_theta*cos_theta);
                if (etai_over_etat * sin_theta > 1.0 ) {
                    vec3 reflected = reflect(unit_direction, rec.normal);
                    scattered = ray(rec.p, reflected);
                    return true;
                }
                double reflect_prob = schlick(cos_theta, etai_over_etat);
                if (random_double() < reflect_prob)
                {
                    vec3 reflected = reflect(unit_direction, rec.normal);
                    scattered = ray(rec.p, reflected);
                    return true;
                }
                vec3 refracted = refract(unit_direction, rec.normal, etai_over_etat);
                scattered = ray(rec.p, refracted);
                return true;
            }
```

the "redundant" if statement is to make the code as easy to read as possible. I agree that chunking the two if statements together is certainly "cleaner", I think it's harder to understand.

The src has the cleaner:
```
                if (  (etai_over_etat * sin_theta > 1.0)
                   || (random_double() < schlick(cos_theta, etai_over_etat))
                   ) {
                    vec3 reflected = reflect(unit_direction, rec.normal);
                    scattered = ray(rec.p, reflected);
                    return true;
                }
```

The dielectric chapter is already the densest chapter in the book and I wanted the source to be incredibly simple to understand. But, having redundant code makes transcription errors _more_ likely, not less.
I don't like this change, but I tolerate it.


The `r_in.time` change is perfect.

trevordblack (approved):

trevordblack:
I would happy with:

```
double reflect_prob = schlick(cos_theta, etai_over_etat);
if (  (etai_over_etat * sin_theta > 1.0)
       || (random_double() < reflect_prob) )
{
```

hollasch:
@trevordblack -- can you take a look at this latest change? I think my hang-up with the earlier code was that I usually think of indices of refraction, but in our code we're really dealing with a baked-in _ratio_ of indices of refraction. I've updated the naming in the refraction code to reflect that. What do you think?

src/InOneWeekend/material.h
```diff
@@ -71,33 +71,35 @@ class metal : public material {
...
 
-            vec3 refracted = refract(unit_direction, rec.normal, etai_over_etat);
+            vec3 refracted = refract(unit_direction, rec.normal, refract_ratio);
             scattered = ray(rec.p, refracted);
             return true;
         }
 
     public:
-        double ref_idx;
+        double refraction_ratio; // Ratio of the two materials' indices of refraction
```
trevordblack, line 102: _oh no_
This member variable is supposed to be the refractive index.

```
             double etai_over_etat = rec.front_face ? (1.0 / ref_idx) : ref_idx;
```
the variable `etai_over_etat` is the refractive ratio. so your change of name variable is fine.
the `1.0` is because we assume that the ray is coming-from/going-to a region with a refractive index of 1.0 (a vacuum)

trevordblack: We're not dealing with a baked in refractive ratio. At least, that's not what our text is saying. The member variable is supposed to be the refractive index of the media.

It just so happens to also be the reciprocal refractive ratio if the outside is a vacuum

hollasch: Ah! Of course; now I get it.

trevordblack: If you change `refraction_ratio` back to `ref_idx` I'm otherwise perfectly happy with this commit

hollasch: Ok, fixed in next change, I think.

trevordblack:
Looks good to me, make sure changes to source make their way back into the inlined source in the text.

hollasch:
Yup, thanks!

hollasch:
Ok @trevordblack , this last change should do it. Also found updates to the PDF specular component for dielectric materials that were missed in book 3.

hollasch:
All changes now squashed and fixed up in PR #739.

PR #730, opened 2020-09-12
bvh_node: don't modify source object list

### LINKED ISSUE #701: [Book 2]: I'm not sure if bvh_node's constructor is correct
https://raytracing.github.io/books/RayTracingTheNextWeek.html#boundingvolumehierarchies/thebvhnodeclass

I'm talking about this: bvh_node(std::vector<shared_ptr<hittable>>& objects, size_t start, size_t end, double time0, double time1)`

In the definition, there is a `sort()` being performed on `objects`, which would modify the variable.  I think the `&` should be removed.

hollasch:
Instead of reordering the vector of scene objects passed to the
`bvh_node` constructor, it now creates a local copy instead. For some
reason, this is actually slightly faster.

Resolves #701


trevordblack:
Neat. Ship it.

trevordblack (approved):

PR #745, opened 2020-10-02
Another small refactor for dielectric scatter()

hollasch:
I think this simplifies and clarifies the dielectric scatter function a bit more, and eliminates the need for an explanatory comment.

Just a single file for now to illustrate the change. If approved, I'll make the remaining corresponding changes throughout.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -2405,6 +2405,7 @@ And the dielectric material that always refracts is:
 
                 vec3 unit_direction = unit_vector(r_in.direction());
                 vec3 refracted = refract(unit_direction, rec.normal, refraction_ratio);
+
                 scattered = ray(rec.p, refracted);
                 return true;
             }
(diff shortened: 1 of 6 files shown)


mrmcsoftware (approved):
I like the symmetry if you know what I mean - it's easy to see that it's either reflected or refracted.  One small thing - maybe I don't know all the benefits of auto (performance, etc.), but in reflectance, why is r0 auto?  Since the function is double and both parameters are double, what's the benefit of auto?  (Besides not having to change r0's type if function is someday changed to something else, or wanting to be consistent with the coding style of the rest of the program.  Both of which are legitimate reasons, of course).  At any rate, I approve.

trevordblack:
Works for me

trevordblack (approved):

hollasch:
> ... in reflectance, why is r0 auto? Since the function is double and both parameters are double, what's the benefit of auto? (Besides not having to change r0's type if function is someday changed to something else, or wanting to be consistent with the coding style of the rest of the program.

Originally, the source code used `float` everywhere. We eventually moved to `double` for several reasons — it's natural with most math functions, it can be faster on some platforms than `float`, real constants (like `1.37`) are `double` by default, and the extra precision can help in some situations.

Still, it should be easy for a reader to use `float` instead, and using `auto` makes those changes much easier than if we had explicitly restated the type for every value. There are situations where `auto` can make a bad day (or in our case, a couple of months of hidden perf degradation), but it's useful when careful.

mrmcsoftware:
> 
> 
> > ... in reflectance, why is r0 auto? Since the function is double and both parameters are double, what's the benefit of auto? (Besides not having to change r0's type if function is someday changed to something else, or wanting to be consistent with the coding style of the rest of the program.
> 
> Originally, the source code used `float` everywhere. We eventually moved to `double` for several reasons — it's natural with most math functions, it can be faster on some platforms than `float`, real constants (like `1.37`) are `double` by default, and the extra precision can help in some situations.
> 
> Still, it should be easy for a reader to use `float` instead, and using `auto` makes those changes much easier than if we had explicitly restated the type for every value. There are situations where `auto` can make a bad day (or in our case, a couple of months of hidden perf degradation), but it's useful when careful.

Makes sense.  I knew the function had previously been float, but I didn't think of the possibility that the reader may want to change it back to float (I was thinking in terms of the maintainers (you, etc.) being unlikely (I assume) to change back)

PR #751, opened 2020-10-06
Convert book 2 main to use scene param structure

hollasch:
Here's what book 2 main would look like if we gathered the scene parameters in a struct and then co-located those with the scene construction. I think it's better.

Book update to follow if this gets approval. Let me know what you think.

### DIFF
--- a/src/TheNextWeek/main.cc
+++ b/src/TheNextWeek/main.cc
@@ -47,7 +47,30 @@ color ray_color(const ray& r, const color& background, const hittable& world, in
 }
 
 
-hittable_list random_scene() {
+struct scene_description {
+    color         background;
+    int           image_width;
+    double        aspect_ratio;
+    int           sample_count;
+    double        aperture;
+    double        vfov;
+    point3        look_from;
+    point3        look_at;
+    hittable_list world;
+};
+
+
+scene_description random_scene() {
+    scene_description scene;
+    scene.background   = color(0.70, 0.80, 1.00);
+    scene.image_width  = 400;
+    scene.aspect_ratio = 16.0 / 9.0;
+    scene.sample_count = 100;
+    scene.aperture     = 0.1;
+    scene.vfov         = 20.0;
+    scene.look_from    = point3(13,2,3);
+    scene.look_at      = point3(0,0,0);
+
     hittable_list world;
 
     auto checker = make_shared<checker_texture>(color(0.2, 0.3, 0.1), color(0.9, 0.9, 0.9));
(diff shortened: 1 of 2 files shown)


hollasch:
One open question is whether we want to formalize this a bit more across all three books. It's a bit odd that book 2 operates in such a unique way. Of course, we could also create that as a separate work issue for later.

D-K-E:
This looks nice. I think for second book, and first book it might be a good idea to formalize this. For third book, I am not sure if it is entirely necessary, since it uses mostly single scene.

trevordblack:
I think this should definitely be formalized across books 2 and 3 at the least. I would be fine if book 1 didn't have an explicit scene parameter, or have it introduced in the last chapter.

On the whole though, this is pretty great work.

trevordblack (approved):

trevordblack:
And, to be more explicit, as far as book 2 and 3 goes, having explicitly structured variables does a much better job of clarifying the differences between any 2 scenes.

trevordblack:
The color background member should be omitted until the light chapter in book 2.

hollasch:
I'm now starting to think that this is a large enough change (particularly with the confirmation that it should be implemented across books) that it should be included in a minor (rather than patch) release. This should probably be moved to v3.3.0.

Given the encouragement here, I'll throw in another iteration for feedback.

trevordblack:
Agree completely

hollasch:
As an aside, I _continually_ fantasize about creating an image description class alongside the scene description class, and adopting a fluent interface for all three, and then wad that all up in the service of simplicity.

I'm continually rebuilding the wall between this project and Steve's Implementation.

trevordblack:
All three?

Do you mean the camera?

hollasch:
Yeah. Since they all work together, it seems like a scene description should include image parameters, a camera, and a world. But this is all so much structure that really adds very little to the instruction. It's me channeling my inner Java.

That said, I _am_ thinking now that the scene description should just own the camera. It's not feeling right that the scene description takes camera parameters and then the caller pulls the camera parameters out to stick into a camera.

hollasch:
New commit has scene description owning the camera. Cleaner, I think.

hollasch:
Ah, more work to be done (but good, I think). The camera aspect ratio should match the image aspect ratio. Scene description wants to be a class at this point. I'll noodle on this later. I think I can preserve (or gain) simplicity with careful thought.

hollasch:
See #754

PR #754, opened 2020-10-09
Managing scene parameters, approach B

hollasch:
Well, I think this pulls things together nicely, and keeps them simple. Definitely not a patch-level change, however, so I'm wedged. I don't want to make this change in the `dev-minor` branch and then have to continually resolve conflicts from the `dev-patch` branch. I'll have to revisit this when we're all on board the dev-minor train.

Still, let me know what you think here. Good for final form? If so, I think it would be easy to adopt this approach for books 1 and 3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ set ( COMMON_ALL
   src/common/rtweekend.h
   src/common/camera.h
   src/common/ray.h
+  src/common/scene_description.h
   src/common/vec3.h
 )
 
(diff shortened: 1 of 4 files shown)


hollasch:
See also #755

trevordblack:
This is something that will need to be massaged into the text.
I think this works overall.

I'd need to either see or write for myself how this works within the greater narrative of books 1 + 2.

But, conditional approve

We should also get @petershirley in here for his thoughts.

trevordblack (approved):

hollasch:
Sounds good. We'll return to this when tackling v3.3.0 or later.

PR #755, opened 2020-10-09
Managing scene parameters, approach C

hollasch:
Had a thought while driving today about another way to accomplish the original objectives of managing scenes better.

This is not code I'd write for an actual C++ project, but avoids creating a new class and associated overhead. Basically just moves all scene parameters into a static file-level anonymous struct with one helper method, and everything in main just accesses that structure without passing it around. All scene functions still drive their own scene parameters. In addition, I really like pulling out the render function from main, as I think this would help discussion throughout the books.

Yeah, it's old-school, but it's also simple and limits changes to main.cc.

### DIFF
--- a/src/TheNextWeek/main.cc
+++ b/src/TheNextWeek/main.cc
@@ -24,30 +24,29 @@
 
 #include <iostream>
 
+static struct {
+    int           image_width;
+    int           sample_count;
+    color         background;
+    camera        cam;
+    hittable_list world;
 
-color ray_color(const ray& r, const color& background, const hittable& world, int depth) {
-    hit_record rec;
-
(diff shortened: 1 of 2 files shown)


trevordblack:
It's a neat idea.
But this has "Why can't I access the scene parameters in this other class" written all over it.

Variables scoped to contain within object files is one of the few things I can think of that just doesn't have an analog in other languages.
This is one of those c-isms that I think we should avoid.

trevordblack (changes requested):
I prefer approach B

hollasch:
Sounds good to me. I agree that restricting concepts to those common across the most languages makes sense.

PR #757, opened 2020-10-11
Book3 fixes

### LINKED ISSUE #738: Book 3 listing 38 (in section 12.2) doesn't tell the user how to handle the new `lights` parameter in `ray_color()`
https://raytracing.github.io/books/RayTracingTheRestOfYourLife.html#cleaninguppdfmanagement./handlingspecular

When I ran the code, I was getting some segfaults, then I realized that `lights` wasn't populated.  This section should instruct the user how to collect the lights from listing 38 and then pass that into `ray_color()`.

### LINKED ISSUE #740: Book 3, section 6.1, listing 10 (Cornell box, refactored): lights are created, but never used

hollasch:
Resolves #738
Resolves #740

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -757,8 +757,9 @@ model.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     ...
-    color ray_color(...)
+    color ray_color(...) {
         ...
+    }
 
     hittable_list cornell_box() {
         hittable_list objects;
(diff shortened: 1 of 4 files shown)


books/RayTracingTheRestOfYourLife.html
```diff
@@ -2109,10 +2097,31 @@
...
+        lights->add(make_shared<xz_rect>(213, 343, 227, 332, 554, shared_ptr<material>()));
+        lights->add(make_shared<sphere>(point3(190, 90, 190), 90, shared_ptr<material>()));
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+
+        auto world = cornell_box();
+
+        color background(0,0,0);
+
+        for (int j ...) {
+            for (int i ...) {
```
trevordblack, line 2121: Not sure if i like this
`for (int j ...)`

would prefer a full write out:
`    for (int j = image_height-1; j >= 0; --j) {`

trevordblack (changes requested):
Update the CHANGELIST

Otherwise, this is good work.

books/RayTracingTheRestOfYourLife.html
hollasch: Done.

PR #761, opened 2020-10-11
Fix `random_unit_vector()`

### LINKED ISSUE #697: Listing 37: Incorrectly generating random unit vector
Maybe I'm reading this incorrectly but what the code does doesn't seem to match the function name.  The current code appears to do a uniform distribution on cylinder and then stretch to a sphere.  Doesn't that raise the probability of points near the poles?

```cpp
vec3 random_unit_vector() {
    auto a = random_double(0, 2*pi);
    auto z = random_double(-1, 1);  // uniform in z means the same # of hits over less surface area
    auto r = sqrt(1 - z*z);
    return vec3(r*cos(a), r*sin(a), z);
}
```
I was expecting something more like this:
```cpp
vec3 random_unit_vector() {
        auto a = random_double(0, 2 * pi);
        auto b = random_double(-pi/2, pi/2);
        auto z = sin(b);
        auto r = sqrt(1 - z * z);
        return vec3(r * cos(a), r * sin(a), z);
}
```

hollasch:
I added an alternate formula for uniformly-distributed points on a unit
sphere for this function. It's presented in
https://mathworld.wolfram.com/SpherePointPicking.html, equations 6-8.

Unfortunately, I misinterpreted "u in [-1, 1]" to be equivalent to
`u = random_double(-1,1)`. However, you're supposed to pick u = cos(phi)
to be uniformly distributed.

I replaced this body with the original one, just generating a unit point
_in_ the unit sphere, and normalizing the result.

Resolves #697

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ Change Log -- Ray Tracing in One Weekend
 # v3.2.2 (in progress)
 
 ### Common
-  - Fix: Added `fmin` to book text for `cos_theta` of `refract`
+  - Fix: Added `fmin` to book text for `cos_theta` of `refract` (#732)
   - Fix: Standardized naming for ray-t and time parameters (#746)
+  - Fix: `random_unit_vector()` was incorrect (#697)
 
 ### In One Weekend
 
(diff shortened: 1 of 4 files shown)


trevordblack:
Please patch this in ASAP.

trevordblack (approved):

PR #762, opened 2020-10-12
Explain sphere uv

### LINKED ISSUE #533: `get_sphere_uv` does not correspond to its explanation

hollasch:
Resolves #533


CHANGELOG.md
```diff
@@ -10,8 +10,10 @@ Change Log -- Ray Tracing in One Weekend
 ### In One Weekend
 
 ### The Next Week
+  - Add: Improve explanation of `get_sphere_uv()` function (#533)
```
hollasch, line 15: I'll fix the typo when I reorganize for release.

src/TheNextWeek/sphere.h
```diff
@@ -43,10 +43,17 @@ bool sphere::bounding_box(double time0, double time1, aabb& output_box) const {
 }
 
 
-void get_sphere_uv(const point3& p, double& u, double& v) {
+static void get_sphere_uv(const point3& p, double& u, double& v) {
```
trevordblack, line 46: If you want to make the source `static`, you'll need to make the inline src in the text `static`.

books/RayTracingInOneWeekend.html
```diff
@@ -901,31 +901,25 @@
...
+        auto discriminant = half_b*half_b - a*c;
+        if (discriminant < 0) return false;
+        auto sqrtd = sqrt(discriminant);
+
+        // Find the nearest root that lies in the acceptable range.
+        auto root = (-half_b - sqrtd) / a;
+        if (root < t_min || t_max < root) {
+            root = (-half_b + sqrtd) / a;
+            if (root < t_min || t_max < root)
+                return false;
```
trevordblack, line 914: Did you confirm that this works correctly?

books/RayTracingTheNextWeek.html
```diff
@@ -1215,7 +1215,7 @@
                 vec3 outward_normal = (rec.p - center) / radius;
                 rec.set_face_normal(r, outward_normal);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-                get_sphere_uv((rec.p-center)/radius, rec.u, rec.v);
+                get_sphere_uv(outward_normal, rec.u, rec.v);
```
trevordblack: This has some interesting implications.
The functionality is identical before and after change.

A texture shaded on the inside is reversed. Which is what you'd expect.

trevordblack:
Uhhh, Steve...

Source:
```
    ...
    auto phi = atan2(p.z(), p.x());
    auto theta = asin(p.y());
    ...
```

Text:
```
x=cos(ϕ)cos(θ)
y=sin(ϕ)cos(θ)
z=sin(θ)
```

The `z` and `y` in the latex equations in the text need be flipped. The source is correct

trevordblack:
I'm pretty sure this is what the original issue was bringing up

books/RayTracingInOneWeekend.html
hollasch: Yes.

src/TheNextWeek/sphere.h
hollasch: This really should be private, but everything's awkward with our approach of doing everything in a header.

hollasch:
Interesting. Later comments suggest that there's no issue. I wasn't aware there was a disagreement between the source and the text; I'll revisit.

src/TheNextWeek/sphere.h
hollasch: Made private static.

trevordblack:
I added another comment that was what I think @D-K-E was referring to.

hollasch:
Thanks. Having gone over it, I still think there are issues. What I'm wrestling with right now is a better explanation for the atan2 arguments, which are juggled to move past the sawtooth discrepancy in the return values (0→π / -π→0). I understand that it works, but am trying to find a way to develop it naturally.

rjkilpatrick:
The convention in ce97f02 is taking θ as the angle from the x-z plane and then ϕ as the polar angle from the z-axis, is this intended?
i.e.
```
x=cos(ϕ)cos(θ)
z=sin(ϕ)cos(θ)
y=sin(θ)
```
Relates to the geometry:
IMG_20201021_233023
Which is atypical for [spherical coordinates](https://en.wikipedia.org/wiki/Spherical_coordinate_system#Cartesian_coordinates).
I would suggest following the original definition of θ being from the pole [0, π) and ϕ being the angle from the x axis, in the x-y plane [0, 2π).

trevordblack:
I think you might be slurring your variables a bit.
Our `y` axis points up.

Do you propose we make the `z` axis point up?

If you propose making the θ as the angle coming down from the `up` axis (in this case `y`), then, yes, we are considering this (and will probably do it). It's an old bug that no one caught cause there isn't a test for it

hollasch:
I've always seen θ going around the equator (for example, longitude), and φ indicating the angle from one pole to the other (like latitude). Like John, I was all set to assert that the book was "nonstandard". Then I did an image search on {spherical coordinates}, and realized it was just another item on the list of

- left-handed versus right-handed coordinates
- clockwise vs counter-clockwise winding order
- row major vs column major matrix memory layout
- column vs row vectors
- +Y up vs +Y down
- and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on

What's really needed here is a proper illustration (in the same manner that John posted above) and a bit more explanation.

trevordblack:
I have an immense urge to go through your list and explain how there's *EXACTLY* one right option for each...

hollasch:
This change has been split up into and is superseded by #774 and #775

PR #763, opened 2020-10-12
Cleanup

hollasch:
**I
    AM
             LOSING
                            MY
                                    MIND**


hollasch:
Find and fix cases where source files and text drifted away from each other.

The change in `hittable_list.h` was a cleanup done in April (5119b05) in book 2 source, but wasn't duplicated in book 3 source.

PR #764, opened 2020-10-13
lambertian::scatter degenerate direction vec test

### LINKED ISSUE #619: Bug: Image corruption if lambertian scattered ray direction is (0,0,0).
There is a small chance on lambertian materials that the scattered ray has direction (0,0,0):

```c++
        virtual bool scatter(
            const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered
        ) const {
            vec3 scatter_direction = rec.normal + random_unit_vector();
            scattered = ray(rec.p, scatter_direction);
            attenuation = albedo;
            return true;
        }
```

In line:
```c++
vec3 scatter_direction = rec.normal + random_unit_vector();
```

If the random_unit_vector() is the opposite of the normal.

Later, on the ray_color() function, the scattered ray will be the ray r of the next recursive call:

```c++
color ray_color(const ray& r, const hittable& world, int depth) {
    hit_record rec;

    // If we've exceeded the ray bounce limit, no more light is gathered.
    if (depth <= 0)
        return color(0,0,0);

    if (world.hit(r, 0.001, infinity, rec)) {
        ray scattered;
        color attenuation;
        if (rec.mat_ptr->scatter(r, rec, attenuation, scattered))
            return attenuation * ray_color(scattered, world, depth-1);
        return color(0,0,0);
    }

    vec3 unit_direction = unit_vector(r.direction());
    auto t = 0.5*(unit_direction.y() + 1.0);
    return (1.0-t)*color(1.0, 1.0, 1.0) + t*color(0.5, 0.7, 1.0);
}

```

On the next recursive call, if there is no hit, on this line, unit_direction() will become (-nan, -nan, -nan) since unit_vector() divides by length zero:
```c++
vec3 unit_direction = unit_vector(r.direction());
```

And this will cause ray_color() to return a -nan color.

hollasch:
In cases where the random unit vector is exactly or very closely equal
to the reversed normal vector, the scatter direction vector will be zero
or close to it. This can lead to infinite or NaN values after the call.

Resolves #619

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -2065,7 +2065,7 @@ it could be a mixture of those strategies. For Lambertian materials we get this 
             virtual bool scatter(
                 const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered
             ) const override {
-                vec3 scatter_direction = rec.normal + random_unit_vector();
+                auto scatter_direction = rec.normal + random_unit_vector();
                 scattered = ray(rec.p, scatter_direction);
                 attenuation = albedo;
                 return true;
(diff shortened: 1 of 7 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -2081,6 +2081,54 @@
...
+            lambertian(const color& a) : albedo(a) {}
+
+            virtual bool scatter(
+                const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered
+            ) const override {
+                auto scatter_direction = rec.normal + random_unit_vector();
+
+                // Catch degenerate scatter direction
+                if (scatter_direction.near_zero())
+                    scatter_direction = rec.normal;
```
trevordblack, line 2118: This section needs to be highlighted

trevordblack (approved):
I'm getting quesy with the amount of caveats and gotcha that we're inlining into the text.
I'm concerned that it's removing some of the fun of the project.

books/RayTracingInOneWeekend.html
hollasch: Done.

PR #770, opened 2020-10-20
Removed bounding_box calls with magic numbers

### LINKED ISSUE #749: rotate_y has bounding box with incorrect time
```
rotate_y::rotate_y(shared_ptr<hittable> p, double angle) : ptr(p) {
  auto radians = degrees_to_radians(angle);
  sin_theta = sin(radians);
  cos_theta = cos(radians); 
  hasbox = ptr->bounding_box(0, 1, bbox);
  ...
```

where
```
hasbox = ptr->bounding_box(0, 1, bbox);
```
the 0 and 1 need to be determined at runtime

trevordblack:
Resolves #749

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -989,6 +989,13 @@ This yields:
     ) {
         auto objects = src_objects; // Create a modifiable array of the source scene objects
 
+        aabb box_check;
+        for (size_t i = start; i < end; i++) {
+            if (!objects[i]->bounding_box(time0,time1,box_check)) {
+                std::cerr << "No bounding box in " << i << " bvh_node constructor.\n";
+            }
+        }
+
         int axis = random_int(0,2);
         auto comparator = (axis == 0) ? box_x_compare
                         : (axis == 1) ? box_y_compare
@@ -1056,10 +1063,6 @@ function.
     inline bool box_compare(const shared_ptr<hittable> a, const shared_ptr<hittable> b, int axis) {
         aabb box_a;
         aabb box_b;
-
-        if (!a->bounding_box(0,0, box_a) || !b->bounding_box(0,0, box_b))
-            std::cerr << "No bounding box in bvh_node constructor.\n";
-
         return box_a.min().e[axis] < box_b.min().e[axis];
     }
 
@@ -2929,14 +2932,13 @@ For a y-rotation class we have:
 
             virtual bool bounding_box(double time0, double time1, aabb& output_box) const override {
                 output_box = bbox;
-                return hasbox;
+                return ptr->bounding_box(time0, time1, output_box);
             }
 
         public:
             shared_ptr<hittable> ptr;
             double sin_theta;
             double cos_theta;
-            bool hasbox;
             aabb bbox;
     };
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -2951,7 +2953,6 @@ With constructor:
         auto radians = degrees_to_radians(angle);
         sin_theta = sin(radians);
         cos_theta = cos(radians);
-        hasbox = ptr->bounding_box(0, 1, bbox);
 
         point3 min( infinity,  infinity,  infinity);
         point3 max(-infinity, -infinity, -infinity);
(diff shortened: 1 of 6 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -2929,14 +2932,13 @@
 
             virtual bool bounding_box(double time0, double time1, aabb& output_box) const override {
                 output_box = bbox;
-                return hasbox;
```
hollasch, line 2932: This looks wrong to me. In the constructor, we compute the bbox of the rotated object and store it in `this->bbox`. Here we compute the original (unrotated) geometry bbox and return that instead, clobbering the computed bbox.

hollasch (changes requested):

trevordblack:
We should consider changing the `ptr` value in `rotate_*` to something more specific like, something else.

Uhh.

`primitive`, `base`, `unmoved_hittable_ptr`

books/RayTracingTheNextWeek.html
trevordblack: how about:
```
 virtual bool bounding_box(double time0, double time1, aabb& output_box) const override {
aabb unrotated_box;
if (ptr->bounding_box(time0, time1, unrotated_box) {
    output_box = bbox;
    return true;
}
else {
    return false;
}
```

hollasch: Ah, I see what's going on. The rotated object is constructed with a given rotation angle, so is constant ever after. The rotated object, however, may be time-varying.

Given this, we can't compute the bounding box of the child object in the constructor, since we have no time parameters at construction. Even if we did, there's no point, because the `bounding_box` method takes some (possibly different) time parameters. So we need to recompute the bounding box on every call.

Thus:
1. Remove the `bbox` member from the class in addition to the `hasbox` flag. They're both unusable.
2. We have code in the constructor that computes the bounding box of the rotated child object's box. That code no longer has any use in the constructor, so move that into the `rotate_y::bounding_box()` method.
3. You'll thus need access to the rotation angle, so you'll have to store that in a member variable.

hollasch: The end result will be that the body of `rotate_y::bounding_box()` will end up looking a lot like what the constructor currently looks like.

hollasch: Note that this is pretty much the way that the `translate` class operates.

hollasch: Ugh.
Wait.

There's a discrepancy here. Like, a big one.

A ray has a _specific_ time value. So a hit result samples the object over time. However, `bounding_box()` methods take a _range_ of times. Thus, they must yield the bounds of the object over that range, as opposed to a tight bounds for a specific instance in time.

That range of time is the range of time values for the entire rendering, and is thus invariant. Therefore, we _should_ know the time range at object construction, and can safely cache the result. So `rotate_y` is actually the way that things should be, except that it hard-coded the time range.

`translate` incorrectly recomputes the bounding box over and over for every query — the time range is always the same, and therefore the result should just have been cached at construction and returned.

But _that_ means that logically, _every hittable should take the time range as parameters to their constructor_. Which would be a major change.

hollasch: I think this change is screwed, and should be lumped in with the other bounding-box work slated for v4.0.0.

hollasch: There's another possible (major) design approach, which is that instead of every hittable taking a time range in the constructor, they have a `set_time_range()` method. For most simple primatives, a no-op. For time-varying primitives, this sets the time range, which will be used by `bounding_box()`. For hittables that contain children, they proxy the `set_time_range()` call to each child, and possibly compute and cache their bounding box. (The bounding box could be memoized in the `bounding_box()` method, but no point doing that on each call; `bounding_box()` might be called many times, while `set_time_range()` should be called only once.

Just a thought.

hollasch:
Ok, we're going to rethink this for v4.0.0 instead.

hollasch (commented):
This turns out to be a larger issue than a small patch change. Deferring to v4.0.0.

hollasch:
See #799

books/RayTracingTheNextWeek.html
hollasch: Clarifying the comment above, just prior to scene render, you'd call `set_time_interval()` (better name than `set_time_range()`) on the root of the scene, to set up the scene graph once prior to rendering, after the scene time interval is known. Prior to rendering the second frame, you'd call `set_time_interval()` with arguments for the second frame.

PR #774, opened 2020-10-25
Refactor sphere hit method to use common code

hollasch:
**→ This is a cleanup and factoring out of the sphere::hit improvements originally reviewed and approved in #762.**

The sphere::hit() function calls get_sphere_uv() with the intersection
point transformed to its position on a unit sphere centered at the
origin. But this calculation was already performed for the normal
vector, so we should just reuse that result instead of calculating it
twice.

Also, the prior sphere::hit() methods duplicated the main update block
with a single change in variable value. This change solves and selects
the intersection root first, then updates the hit record for a good hit,
simplifying the code and the text.

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -222,9 +222,7 @@ The `moving_sphere::hit()` function is almost identical to the `sphere::hit()` f
 just needs to become a function `center(time)`:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-    bool moving_sphere::hit(
-        const ray& r, double t_min, double t_max, hit_record& rec) const {
-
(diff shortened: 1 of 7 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -233,35 +231,26 @@
...
-            auto temp = (-half_b - root) / a;
-            if (temp < t_max && temp > t_min) {
-                rec.t = temp;
-                rec.p = r.at(rec.t);
+        rec.t = root;
+        rec.p = r.at(rec.t);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-                auto outward_normal = (rec.p - center(r.time())) / radius;
+        auto outward_normal = (rec.p - center(r.time())) / radius;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
```
trevordblack, line 249: this should not be highlighted. it is not new

trevordblack (approved):

PR #775, opened 2020-10-25
Corrected sphere UV coordinates

### LINKED ISSUE #533: `get_sphere_uv` does not correspond to its explanation

hollasch:
**→ This is a continuation and final cleanup of the changes originally presented in #762.**

Fixed in text and code. This addresses the following problems:

- Confusion between UV space and texture image space. There were places
  that assumed Y/V grows, down, others where Y/V grow upwards. Correct
  UV coordinates have Y/V growing upwards, as in regular 2D Cartesian
  coordinates. The V coordinate is flipped for texture-map lookup, left
  alone for all other coordinate uses.

- The Y and Z coordinates were confused in several equations.

- The text had Y = sin(theta), where it should have been
  Y = -cos(theta).

- There was nothing in the text to explain the rearrangement of atan2()
  arguments in order to get a continuous 0->2pi return value. I
  introduced the equivalence formula to get continuous return values.

- get_sphere_uv() was in sphere.h for book 2, but hittable.h for book 3.
  This change also moves the function to sphere.h for book 3.

- get_sphere_uv() moved to be a private static method on class sphere.

- I am used to (and therefore prefer) phi corresponding to latitude, and
  theta corresponding to longitude, where the text and code flip these
  two angles. In a quick search of definitions, it looks like this is
  yet another case of competing conventions with both sides roughly
  equal in size. Thus, no clear winner, so I'm leaving this as is.

Resolves #533

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -1148,52 +1148,81 @@ We will also need to compute $(u,v)$ texture coordinates for hittables.
 Texture Coordinates for Spheres
 --------------------------------
 <div class='together'>
-For spheres, this is usually based on some form of longitude and latitude, _i.e._, spherical
-coordinates. So if we have a $(\theta,\phi)$ in spherical coordinates, we just need to scale
-$\theta$ and $\phi$ to fractions. If $\theta$ is the angle down from the pole, and $\phi$ is the
-angle around the axis through the poles, the normalization to $[0,1]$ would be:
+For spheres, texture coordinates are usually based on some form of longitude and latitude, _i.e._,
+spherical coordinates. So we compute $(\theta,\phi)$ in spherical coordinates, where $\theta$ is the
+angle up from the bottom pole (that is, up from -Y), and $\phi$ is the angle around the Y-axis (from
+-X to +Z to +X to -Z back to -X).
+
+We want to map $\theta$ and $\phi$ to texture coordinates $u$ and $v$ each in $[0,1]$, where
+$(u=0,v=0)$ maps to the bottom-left corner of the texture. Thus the normalization from
+$(\theta,\phi)$ to $(u,v)$ would be:
 
   $$ u = \frac{\phi}{2\pi} $$
   $$ v = \frac{\theta}{\pi} $$
 </div>
 
 <div class='together'>
-To compute $\theta$ and $\phi$, for a given hitpoint, the formula for spherical coordinates of a
-unit radius sphere on the origin is:
+To compute $\theta$ and $\phi$ for a given point on the unit sphere centered at the origin, we start
+with the equations for the corresponding Cartesian coordinates:
 
-  $$ x = \cos(\phi) \cos(\theta) $$
-  $$ y = \sin(\phi) \cos(\theta) $$
-  $$ z = \sin(\theta) $$
+  $$ \begin{align*}
+      x &= -\cos(\phi) \sin(\theta) \\
+      y &= -\cos(\theta)            \\
+      z &= \sin(\phi) \sin(\theta)
+     \end{align*}
+  $$
 </div>
 
 <div class='together'>
-We need to invert that. Because of the lovely `<cmath>` function `atan2()` which takes any number
-proportional to sine and cosine and returns the angle, we can pass in $x$ and $y$ (the
-$\cos(\theta)$ cancel):
+We need to invert these equations to solve for $\theta$ and $\phi$. Because of the lovely `<cmath>`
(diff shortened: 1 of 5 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -1148,52 +1148,81 @@
...
-$\theta$ and $\phi$ to fractions. If $\theta$ is the angle down from the pole, and $\phi$ is the
-angle around the axis through the poles, the normalization to $[0,1]$ would be:
+For spheres, texture coordinates are usually based on some form of longitude and latitude, _i.e._,
+spherical coordinates. So we compute $(\theta,\phi)$ in spherical coordinates, where $\theta$ is the
+angle up from the bottom pole (that is, up from -Y), and $\phi$ is the angle around the Y-axis (from
+-X to +Z to +X to -Z back to -X).
+
+We want to map $\theta$ and $\phi$ to texture coordinates $u$ and $v$ each in $[0,1]$, where
+$(u=0,v=0)$ maps to the bottom-left corner of the texture. Thus the normalization from
+$(\theta,\phi)$ to $(u,v)$ would be:
```
trevordblack, line 1158: axes need $$, e.g. $X$

books/RayTracingTheNextWeek.html
```diff
@@ -1148,52 +1148,81 @@
...
+To compute $\theta$ and $\phi$ for a given point on the unit sphere centered at the origin, we start
+with the equations for the corresponding Cartesian coordinates:
 
-  $$ x = \cos(\phi) \cos(\theta) $$
-  $$ y = \sin(\phi) \cos(\theta) $$
-  $$ z = \sin(\theta) $$
+  $$ \begin{align*}
+      x &= -\cos(\phi) \sin(\theta) \\
+      y &= -\cos(\theta)            \\
+      z &= \sin(\phi) \sin(\theta)
```
trevordblack, line 1171: propose:
```
      x &= -\cos(\phi) \sin(\theta) \\
      z &=   \sin(\phi) \sin(\theta)
      y &= -\cos(\theta)            \\
```

books/RayTracingTheNextWeek.html
```diff
@@ -1148,52 +1148,81 @@
...
 --------------------------------
 <div class='together'>
-For spheres, this is usually based on some form of longitude and latitude, _i.e._, spherical
-coordinates. So if we have a $(\theta,\phi)$ in spherical coordinates, we just need to scale
-$\theta$ and $\phi$ to fractions. If $\theta$ is the angle down from the pole, and $\phi$ is the
-angle around the axis through the poles, the normalization to $[0,1]$ would be:
+For spheres, texture coordinates are usually based on some form of longitude and latitude, _i.e._,
+spherical coordinates. So we compute $(\theta,\phi)$ in spherical coordinates, where $\theta$ is the
+angle up from the bottom pole (that is, up from -Y), and $\phi$ is the angle around the Y-axis (from
+-X to +Z to +X to -Z back to -X).
```
trevordblack, line 1154: Why from -X and not +X?

hollasch: It's an arbitrary call, and I saw no need to change the way we've texture-map our spheres (if we did, it would be a behavior change, and thus a major-level fix).

books/RayTracingTheNextWeek.html
hollasch: Will do.

books/RayTracingTheNextWeek.html
hollasch: Just tried that, and it looks very ugly.

PR #790, opened 2020-11-03
Book 1: add commentary about double vs float

### LINKED ISSUE #752: Book 1: Double vs Float more explanation
It would be nice to have a small overview of pros and cons of using `double` or `float` for the end of first book section 3.1. The difference became relevant as I was implementing the second book in cuda. 
- `double`: more numerical precision
- `double`: twice the memory use compared to `float`
- `float`: less numerical precision.
- `float`: cuts down the memory use by half compared to `double`

When I tested the final scene of the second book, I got out of ressources error from cuda, but everything worked fine with float. 
Though obvious it is, it might be important to mention that.

hollasch:
Resolves #752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Change Log -- Ray Tracing in One Weekend
   - Change: `hittable:hit()` methods use new interval class for ray t parameter
 
 ### In One Weekend
+  - Added: More commentary about the choice between `double` and `float` (#752)
 
 ### The Next Week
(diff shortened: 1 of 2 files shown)


rjkilpatrick (approved):
I think that's the most you can say without the reader getting bogged down in the perils of floating-point.

hollasch:
~the perils of floating-point~ the _joys_ of floating-point. :D 

Thank you for reviewing!

PR #793, opened 2020-11-04
Convert clamp() to interval::clamp()

hollasch:
Third iteration, incorporating prior pull requests.


hollasch:
Merging, but comments always welcome.

Issue #779, opened 2020-10-27
Validate Book 2 progression for milestone alpha.2


hollasch:
After the messy merge of `dev-patch` (v3.2.2) into `dev-minor` and all of the work going on in v4, we should run through the book progression before v4 release.

Validate/update rendered images. Make sure that they are PNG format, and sized according to the rendered size specified in the code.

- [x] 2.1: Bouncing spheres
- [x] 2.2: Spheres on checkered ground
- [x] 2.3: Checkered spheres
- [x] 2.5: Earth-mapped sphere
- [x] 2.6: White noise
- [x] 2.7: White noise, blurred
- [x] 2.8: Tiled random patterns
- [x] 2.9: Hashed random texture
- [x] 2.10: Perlin texture with trilinear interpolation
- [x] 2.11: Perlin texture, trilinearly interpolated, smoothed
- [x] 2.12: Perlin texture, higher frequency
- [x] 2.13: Perlin texture, shifted off integer values
- [x] 2.14: Perlin texture with turbulence
- [x] 2.15: Perlin noise, marbled texture
- [x] 2.16: Scene with rectangle light source
- [x] 2.17: Scene with rectangle and sphere light sources
- [x] 2.18: Empty Cornell box
- [x] 2.19: Cornell box with two blocks
- [x] 2.20: Standard Cornell box scene
- [x] 2.21: Cornell box with blocks of smoke
- [x] 2.22: Final scene

hollasch:
Done.

Issue #778, opened 2020-10-27
Validate Book 1 Progression (Near End of v4 Development)


hollasch:
After the messy merge of `dev-patch` (v3.2.2) into `dev-minor` and all of the work going on in v4, we should run through the book progression before v4 release.

Validate/update rendered images. Make sure that they are PNG format, and sized according to the rendered size specified in the code.
- [ ] 1.1: First PPM image
- [ ] 1.2: A blue-to-white gradient depending on ray Y coordinate
- [ ] 1.3: A simple red sphere
- [ ] 1.4: A sphere colored according to its normal vectors
- [ ] 1.5: Resulting render of normals-colored sphere with ground
- [ ] 1.6: Before and after anti aliasing
- [ ] 1.7: First render of a diffuse sphere
- [ ] 1.8: Diffuse sphere, with gamma correction
- [ ] 1.9: Correct rendering of Lambertian spheres
- [ ] 1.10: Rendering of diffuse spheres with hemispherical scattering
- [ ] 1.11: Shiny metal
- [ ] 1.12: Fuzzed metal
- [ ] 1.13: Glass first
- [ ] 1.14: Glass sphere that always refracts
- [ ] 1.15: Glass sphere that sometimes refracts
- [ ] 1.16: A hollow glass sphere
- [ ] 1.17: A wide-angle view
- [ ] 1.18: A distant view
- [ ] 1.19: Zooming in
- [ ] 1.20: Spheres with depth-of-field
- [ ] 1.21: Final scene

Also double-check with the following issues:
- [ ] #778
- [ ] #914

hollasch:
Note to double check with #914

PR #801, opened 2020-11-13
Move method defs inside classes, remove inline

hollasch:
Draft changes. Still need to update books.

Thoughts?

### DIFF
--- a/src/InOneWeekend/hittable.h
+++ b/src/InOneWeekend/hittable.h
@@ -24,7 +24,7 @@ class hit_record {
     double t;
     bool front_face;
 
-    inline void set_face_normal(const ray& r, const vec3& outward_normal) {
+    void set_face_normal(const ray& r, const vec3& outward_normal) {
         front_face = dot(r.direction(), outward_normal) < 0;
         normal = front_face ? outward_normal :-outward_normal;
     }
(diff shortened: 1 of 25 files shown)


trevordblack:
I feel like we should add a 5 sentence blurb in the Overview explaining how we treat the code in book. Where we try to be some what language agnostic but use c++ as a lingua franca.

I'm also super tempted to apologize profusely for the coding style.

Also, we should probably rewrite the Overview of the first book anyway. Could do with some polish.

trevordblack:
Do we want to explain `inline`.

We've pretty much always had a higher priorities than performance, but I feel like we had a teaching oppurtunity with `inlines` in.
Not that we used it (or that `inline` was actually doing anything)

trevordblack:
Still need to update the books (as you mention above)

This looks thorough (and possibly complete), I'm accepting this merge.

My request is that the issue not be closed out for a while. I would like to make a once-over the code at some point (at the very very end), and this is something that I want to check.

trevordblack (approved):

hollasch:
“Once-over the code”? Is this different than #778, #779?

trevordblack:
Nope. that's exactly what I had in mind.

hollasch:
This was just a sketch of what was possible, the actual implementation is proceeding more incrementally. In addition, re-re-realized some aspects of `inline` that are necessary in some scenarios (comments in the issue and in the inline-clean pull request).

Closing out this draft.

PR #810, opened 2020-11-23
Add new rtw_image class to manage image files

### LINKED ISSUE #807: Improve texture file locating
We need a better solution for our texture map file loading. I have scripts that put the image file in just the right place for the way I build and run the project, but that will be unique to each developer.

Thinking of writing a utility function that tries file, images/file, ../images/file, ../../images/file, ../../../images/file, and so on until ... some point. I can't do too much with the file system in C++11 in an OS-agnostic way, so it'll have to be somewhat simple-minded. Thinking of putting this in the `rtw_stb_image.h` file as a helper function.

hollasch:
- Honors RTW_IMAGES environment variable to locate RTW image directory
- If RTW_IMAGES is unset, searches for image files in several likely
  image directory locations.
- Separates image data management from diffuse texture map class.
- Better error reporting on image load failure.
- More robust getters in the face of absent image data.

Resolves #807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Change Log -- Ray Tracing in One Weekend
   - Change: `hittable:hit()` methods use new interval class for ray-t parameter
   - Change: Class public/private access labels get two-space indents (#782)
   - Change: `interval::clamp()` replaces standalone `clamp` utility function
+  - New: `rtw_image` class for easier image data loading, searches more locations (#807)
 
 ### In One Weekend
   - Added: More commentary about the choice between `double` and `float` (#752)
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2084,31 +2084,27 @@ files even easier, we provide a helper class to manage all this -- `rtw_image`:
 
         rtw_image(const char* image_filename) {
             // Loads image data from the given file name. If the RTW_IMAGES environment variable is
-            // defined, looks only in that directory for the image file. If RTW_IMAGES is not
-            // defined, this constructor will search for the specified image file first from the
-            // current directory, then in the images/ subdirectory, then the _parent's_ images/
-            // subdirectory, and then _that_ parent, on so on, for six levels up. If the image was
-            // not loaded successfully, width() and height() will return 0.
+            // defined, looks only in that directory for the image file. If the image was not found,
+            // searches for the specified image file first from the current directory, then in the
+            // images/ subdirectory, then the _parent's_ images/ subdirectory, and then _that_
+            // parent, on so on, for six levels up. If the image was not loaded successfully,
+            // width() and height() will return 0.
 
+            auto filename = std::string(image_filename);
             const auto imagedir = getenv("RTW_IMAGES");
-            if (imagedir) {
-                // If RTW_IMAGES is defined, then only use that value to locate the image file.
-                if (load(std::string(imagedir) + "/" + image_filename)) return;
-            } else {
-                // Hunt for the image file in some likely locations.
-                auto filename = std::string(image_filename);
-                if (load(filename)) return;
-                if (load("images/" + filename)) return;
-                if (load("../images/" + filename)) return;
-                if (load("../../images/" + filename)) return;
-                if (load("../../../images/" + filename)) return;
-                if (load("../../../../images/" + filename)) return;
-                if (load("../../../../../images/" + filename)) return;
-                if (load("../../../../../../images/" + filename)) return;
-            }
+
+            // Hunt for the image file in some likely locations.
+            if (imagedir && load(std::string(imagedir) + "/" + image_filename)) return;
+            if (load(filename)) return;
+            if (load("images/" + filename)) return;
+            if (load("../images/" + filename)) return;
+            if (load("../../images/" + filename)) return;
+            if (load("../../../images/" + filename)) return;
+            if (load("../../../../images/" + filename)) return;
+            if (load("../../../../../images/" + filename)) return;
+            if (load("../../../../../../images/" + filename)) return;
 
             std::cerr << "ERROR: Could not load image file '" << image_filename << "'.\n";
-            std::cerr << "STBI: " << stbi_failure_reason() << '\n';
         }
 
         ~rtw_image() { STBI_FREE(data); }
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2057,105 +2057,147 @@ This is just a fractional position.
 
 Storing Texture Image Data
 ---------------------------
-Now we also need to create a texture class that holds an image. I am going to use my favorite image
-utility [stb_image][]. It reads in an image into a big array of unsigned char. These are just packed
-RGBs with each component in the range [0,255] (black to full white). The `image_texture` class uses
-the resulting image data.
-
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-    #include "rtweekend.h"
-    #include "rtw_stb_image.h"
-    #include "perlin.h"
-
-    #include <iostream>
-
-    ...
-
-    class image_texture : public texture {
-      public:
-        const static int bytes_per_pixel = 3;
-
-        image_texture() : data(nullptr), width(0), height(0), bytes_per_scanline(0) {}
-
-        image_texture(const char* filename) {
-            auto components_per_pixel = bytes_per_pixel;
-
-            data = stbi_load(
-                filename, &width, &height, &components_per_pixel, components_per_pixel);
-
-            if (!data) {
-                std::cerr << "ERROR: Could not load texture image file '" << filename << "'.\n";
-                width = height = 0;
-            }
-
-            bytes_per_scanline = bytes_per_pixel * width;
-        }
-
-        ~image_texture() {
-            delete data;
-        }
-
-        virtual color value(double u, double v, const vec3& p) const override {
-            // If we have no texture data, then return solid cyan as a debugging aid.
-            if (data == nullptr)
-                return color(0,1,1);
-
-            // Clamp input texture coordinates to [0,1] x [1,0]
-            u = interval(0,1).clamp(u);
-            v = 1.0 - interval(0,1).clamp(v);  // Flip V to image coordinates
-
-            auto i = static_cast<int>(u * width);
-            auto j = static_cast<int>(v * height);
-
-            // Clamp integer mapping, since actual coordinates should be less than 1.0
-            if (i >= width)  i = width-1;
-            if (j >= height) j = height-1;
-
-            const auto color_scale = 1.0 / 255.0;
-            auto pixel = data + j*bytes_per_scanline + i*bytes_per_pixel;
-
-            return color(color_scale*pixel[0], color_scale*pixel[1], color_scale*pixel[2]);
-        }
-
-      private:
-        unsigned char *data;
-        int width, height;
-        int bytes_per_scanline;
-    };
-
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [img-texture]: <kbd>[texture.h]</kbd> Image texture class]
-
-<div class='together'>
-The representation of a packed array in that order is pretty standard. Thankfully, the [stb_image][]
-package makes that super simple -- just write a header called `rtw_stb_image.h` that also deals with
-some compiler warnings:
+Now it's time to create a texture class that holds an image. I am going to use my favorite image
+utility, [stb_image][]. It reads image data into a big array of unsigned char. These are just packed
+RGBs with each component in the range [0,255] (black to full white). To help make loading our image
+files even easier, we provide a helper class to manage all this -- `rtw_image`:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Change Log -- Ray Tracing in One Weekend
   - Change: `hittable:hit()` methods use new interval class for ray-t parameter
   - Change: Class public/private access labels get two-space indents (#782)
   - Change: `interval::clamp()` replaces standalone `clamp` utility function
+  - New: `rtw_image` class for easier image data loading, searches more locations (#807)
```
trevordblack, line 12: , ... better error reporting

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
 
 Storing Texture Image Data
 ---------------------------
-Now we also need to create a texture class that holds an image. I am going to use my favorite image
-utility [stb_image][]. It reads in an image into a big array of unsigned char. These are just packed
-RGBs with each component in the range [0,255] (black to full white). The `image_texture` class uses
-the resulting image data.
+Now it's time to create a texture class that holds an image. I am going to use my favorite image
+utility, [stb_image][]. It reads image data into a big array of unsigned char. These are just packed
```
trevordblack, line 2061: `unsigned char`s ?

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
-    ...
-
-    class image_texture : public texture {
-      public:
-        const static int bytes_per_pixel = 3;
+    #ifndef RTWEEKEND_STB_IMAGE_H
+    #define RTWEEKEND_STB_IMAGE_H
 
-        image_texture() : data(nullptr), width(0), height(0), bytes_per_scanline(0) {}
+    // Disable pedantic warnings for this header from the Microsoft Visual C++ compiler.
```
trevordblack, line 2069: I understand the pull of snark, but consider removing "pedantic"

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
+        #pragma warning (pop)
+    #endif
+
+    #endif
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [img-texture]: <kbd>[texture.h]</kbd> Image texture class]
+    [Listing [rtw_image]: <kbd>[rtw_stb_image.h] The rtw_image helper class]
+
+The preceding listing assumes that you have copied the `stb_image.h` header into a folder called
+`external`. Adjust according to your directory structure.
```
trevordblack, line 2153: Consider moving this descriptor to above the source block.

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
-            // If we have no texture data, then return solid cyan as a debugging aid.
-            if (data == nullptr)
-                return color(0,1,1);
+        ~rtw_image() { STBI_FREE(data); }
 
-            // Clamp input texture coordinates to [0,1] x [1,0]
-            u = interval(0,1).clamp(u);
-            v = 1.0 - interval(0,1).clamp(v);  // Flip V to image coordinates
+        bool load(const std::string filename) {
+            // Loads image data from the given file name. Returns true if the load succeeded.
```
trevordblack, line 2113: "given file location"

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
+            auto components_per_pixel = bytes_per_pixel; // Components per pixel
```
trevordblack, line 2114: remove the comment, or make it clarifying somehow

trevordblack: Consider creating a class member variable
`const int components_per_pixel = 3` below

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
+            // Clamp pixel coordintes to [0,width-1] x [0,height-1]
+            x = (x < 0) ? 0 : (x < image_width)  ? x : (image_width  - 1);
+            y = (y < 0) ? 0 : (y < image_height) ? y : (image_height - 1);
 
-            return color(color_scale*pixel[0], color_scale*pixel[1], color_scale*pixel[2]);
+            return data + y*bytes_per_scanline + x*bytes_per_pixel;
         }
 
       private:
+        const int bytes_per_pixel = 3;
```
trevordblack, line 2137: move to
```
        unsigned char *data;
        int image_width, image_height;
        const int bytes_per_pixel = 3;
        int bytes_per_scanline;
```

Also consider components_per_pixel member variable

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
+        ~rtw_image() { STBI_FREE(data); }
 
-            // Clamp input texture coordinates to [0,1] x [1,0]
-            u = interval(0,1).clamp(u);
-            v = 1.0 - interval(0,1).clamp(v);  // Flip V to image coordinates
+        bool load(const std::string filename) {
+            // Loads image data from the given file name. Returns true if the load succeeded.
+            auto components_per_pixel = bytes_per_pixel; // Components per pixel
+            data = stbi_load(filename.c_str(),
+                &image_width, &image_height, &components_per_pixel, bytes_per_pixel);
```
trevordblack, line 2116: consider modifying a local variable instead of member variables,
then
```
if(data -= nullptr)
    return false;

image_width = local_width;
image_height = local_width;
bytes_per_scanline = image_width * bytes_per_pixel;
```

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
+            bytes_per_scanline = image_width * bytes_per_pixel;
+            return data != nullptr;
+        }
 
-            auto i = static_cast<int>(u * width);
-            auto j = static_cast<int>(v * height);
+        int width()  const { return (data == nullptr) ? 0 : image_width; }
+        int height() const { return (data == nullptr) ? 0 : image_height; }
```
trevordblack, line 2122: maybe return -1?

CHANGELOG.md
hollasch: Sadly, the better error reporting didn't really pan out. See subsequent commit.

books/RayTracingTheNextWeek.html
trevordblack: what does stbi return if it fails to load?

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
 
-            // Clamp integer mapping, since actual coordinates should be less than 1.0
-            if (i >= width)  i = width-1;
-            if (j >= height) j = height-1;
+        const unsigned char* pixel_data(int x, int y) const {
+            // Return the address of the three bytes of the pixel at x,y (or magenta if no data).
+            static unsigned char magenta[] = { 255, 0, 255 };
+            if (!data) return magenta;
```
trevordblack, line 2127: data == nullptr

`if (!ptr)` is an anti-pattern that must be destroyed....

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
 
-            const auto color_scale = 1.0 / 255.0;
-            auto pixel = data + j*bytes_per_scanline + i*bytes_per_pixel;
+            // Clamp pixel coordintes to [0,width-1] x [0,height-1]
+            x = (x < 0) ? 0 : (x < image_width)  ? x : (image_width  - 1);
+            y = (y < 0) ? 0 : (y < image_height) ? y : (image_height - 1);
```
trevordblack, line 2131: we might want to briefly discuss different clamping functions.

books/RayTracingTheNextWeek.html
hollasch: Actually wasn't being snarky here, but it looks like "pedantic" is more exclusively pejorative than I'd realized. Will fix.

books/RayTracingTheNextWeek.html
hollasch: Sounds good.

books/RayTracingTheNextWeek.html
hollasch: Good call.

books/RayTracingTheNextWeek.html
trevordblack: Consider chunking these calls to make the ternary operator a little more clear
or, just rewriting as if-else statements.

It's pretty clever as is, though. Just not the easiest to parse.

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
 
-            return color(color_scale*pixel[0], color_scale*pixel[1], color_scale*pixel[2]);
+            return data + y*bytes_per_scanline + x*bytes_per_pixel;
```
trevordblack, line 2133: I really like this construction!

books/RayTracingTheNextWeek.html
trevordblack: I get that you're using "pedantic" as a strictly technical term here, but it sounds like snark to a lay audience.

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
-
-    #include <iostream>
-
-    ...
-
-    class image_texture : public texture {
-      public:
-        const static int bytes_per_pixel = 3;
+    #ifndef RTWEEKEND_STB_IMAGE_H
+    #define RTWEEKEND_STB_IMAGE_H
```
trevordblack, line 2067: RTW_STB_IMAGE_H ?

books/RayTracingTheNextWeek.html
hollasch: No, this is correct; the STBI interface is clumsy here. If the last argument is 0, then the library will determine the number of bytes per pixel and store it in the out (penultimate) parameter. If the last argument is 1, 2, 3 or 4, the load function will transform the pixel to that format (I, IA, RGB, or RGBA) on load. However, it still takes the penultimate parameter and sets it to the actual value from the file.

We basically always want RGB, so we don't care about this value. I'll just rename it `dummy` to try to alleviate some of the confusion.

books/RayTracingTheNextWeek.html
hollasch: STBI returns `nullptr` on load failure.

Also, I want to return zero, so the single magenta indicator pixel still works. Basically, if the load fails, but you don't really care, everything will still work, with a magenta color. I don't want to leave this in a zombie state.

books/RayTracingTheNextWeek.html
hollasch: Matter of opinion. Implicit boolean coercion is a thing in many languages (and libraries). Still, I'll change it.

books/RayTracingTheNextWeek.html
```diff
@@ -2057,105 +2057,147 @@
...
+      public:
+        image_texture() {}
+        image_texture(const char* filename) : image(filename) {}
 
-    // Restore warning levels.
-    #ifdef _MSC_VER
-        // Microsoft Visual C++ Compiler
-        #pragma warning (pop)
-    #endif
+        virtual color value(double u, double v, const vec3& p) const override {
```
trevordblack, line 2176: <For prosperity>

remove virtual for override alone

books/RayTracingTheNextWeek.html
trevordblack: does `interval` work here?

hollasch: Keep in mind that this file is really intended as boilerplate code we provide (like the STBI header itself) that the user is not supposed to worry about, particularly since it's so tailored to one particular external. Think of it more as our private extension to STBI to make our lives easier, and to move it off the reader's plate of things to understand.

Also, because of this, I don't have access to the other RTW functions in our codebase (like clamp or interval::clamp). I had stronger wording to the user about this (“don't worry your pretty little head”), but then pulled it because readers who choose to implement this in other languages or with some other image data library will have to wrestle with this somehow anyway.

Still, I'll improve the code above.

books/RayTracingTheNextWeek.html
```diff
@@ -2084,31 +2084,27 @@
...
-                if (load("../images/" + filename)) return;
-                if (load("../../images/" + filename)) return;
-                if (load("../../../images/" + filename)) return;
-                if (load("../../../../images/" + filename)) return;
-                if (load("../../../../../images/" + filename)) return;
-                if (load("../../../../../../images/" + filename)) return;
-            }
+
+            // Hunt for the image file in some likely locations.
+            if (imagedir && load(std::string(imagedir) + "/" + image_filename)) return;
```
trevordblack, line 2097: what happens if RTW_IMAGES is defined with a trailing `/` or `\`, e.g.
`C:\Users\trevor\raytracing.github.io\images\`
`/home/trevor/raytracing.github.io/images/`
`/Users/trevor/raytracing.github.io/images/`

trevordblack (approved):
See inlined comments, mostly pedantic (*ahem*), but, in general, approve.

books/RayTracingTheNextWeek.html
hollasch: The library guarantees these out parameters will be left unchanged on load failure.

books/RayTracingTheNextWeek.html
trevordblack: Ah. Then, yes, I was overthinking it.

books/RayTracingTheNextWeek.html
hollasch: Check.

books/RayTracingTheNextWeek.html
hollasch: Preview of coming attractions.

books/RayTracingTheNextWeek.html
hollasch: On Windows, at least, paths my contain sequences of slashes, which are interpreted as a single slash. I _think_ this is also true for Posix/OSX/Linux. If not, `¯\_(ツ)_/¯`

hollasch:
Thanks for the feedback @trevordblack — good stuff.

PR #811, opened 2020-11-24
Scene

hollasch:
This is a major change to the codebase, refactoring multiple components that were previously uncoordinated.

Generally, we introduce a new `scene` class, which is responsible for managing image parameters, the camera, the world geometry, and lights. This is particularly useful in the book 2 code, where we maintain a number of different scenes, with the ability to choose & render any one of them. Since all of the above properties can be unique to each scene, it became cumbersome to ensure that all parts were properly set up. The new `scene` class makes this significantly easier, and also simplifies the render code as well.

The camera class is changed so that camera parameters are exposed as public member variables, and may be updated in random order as needed. A new `camera::initialize()` method now explicitly sets the camera up for rendering after parameters are set, and before `camera::get_ray` is used to generate scene rays.

Aspect ratio is maintained now by the `scene` class, so that it coordinates the rendered image dimensions and the camera viewport.

The `ray_color()` function is now located in the `scene` class as a private method, and the render loop that used to be in the `main()` functions is now moved into the `scene` class as a public method. This also allows some coordination between the two methods at render time.

An additional change tucked into this commit is the fact that the book 2 code changes the default scene from the Cornell box to the book 2 final scene with simplified parameters for a faster render. This scene exercises significantly more of the book 2 code, so also now serves as a good sanity and performance test.


trevordblack (commented):
For all scene descriptor functions, e.g. `random_spheres`, `cornell_box` I think that it's more clear if the functions don't have pass-by-address parameters and are non-void
```
void main() {
  scene scene_desc = random_spheres();
  scene_desc.render();
}
```
I think it's a little easier to understand

src/TheNextWeek/main.cc
```diff
@@ -235,163 +304,44 @@ hittable_list final_scene() {
...
-    int image_width = 400;
-    int samples_per_pixel = 100;
-    int max_depth = 50;
-
-    // World
+void main() {
+    scene scene_desc;
 
-    hittable_list world;
+    scene_desc.background = color(0.70, 0.80, 1.00);
```
trevordblack, line 327: This probably belongs in the scene function?

Most scenes redefine background to black, so move this line of code into those that don't.

trevordblack: oh, wait. I see more `scene_desc` changes below. I guess it's fine for them to be outside of scene function calls.

I do think that these `scene_desc` changes should probably be placed _below_ the `switch` statement

books/RayTracingInOneWeekend.html
```diff
@@ -484,11 +486,11 @@
 common.
 
 In addition to setting up the pixel dimensions for the rendered image, we also need to set up a
-virtual viewport through which to pass our scene rays. For the standard square pixel spacing, the
-viewport's aspect ratio should be the same as our rendered image. We'll just pick a viewport two
-units in height. We'll also set the distance between the projection plane and the projection point
-to be one unit. This is referred to as the “focal length”, not to be confused with “focus distance”,
-which we'll present later.
+virtual viewport through which to pass our scene rays. For the standard pixel spacing (equidistant
```
trevordblack, line 489: equidistant -> equally spaced

books/RayTracingTheNextWeek.html
```diff
@@ -155,6 +144,15 @@
...
 
+      public:
+        double vfov       = 40;
+        double aperture   = 0;
+        double focus_dist = 10;
+
+        point3 lookfrom = point3(0,0,-1);
+        point3 lookat   = point3(0,0,0);
+        vec3   vup      = vec3(0,1,0);
+
```
trevordblack, line 155: needs `time_start` and `time_end`. delete `time0` and `time1` below.

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
 
 
 
+Creating a Scene
+====================================================================================================
+Sometimes you just have to make a scene. At this point, our `main.cc` has a standalone function to
```
trevordblack, line 2787: cut "Sometimes you just have to make a scene."

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
...
+image. Things have gotten complex enough to warrant a refactoring.
+
+
+The scene Class
+----------------
+<div class='together'>
+Let's gather together all of the components of a scene into a single class that manages everything.
+Naturally, we'll name this the `scene` class.
+
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
```
trevordblack, line 2799: split up the inline source of the scene class.

Something like
1. Scene class and variables
2. ray_color
3. render

While everything in this class is collated from elsewhere, if you have a really long wall of inline source, that encourages the reader to just copy/paste the code over. Which will reduce the educational value.

Have writable chunks

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
...
+        world.add(make_shared<sphere>(point3( 1.0,    0.0, -1.0),   0.5, material_right));
+
+        scene_desc.render();
+    }
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [main-with-scene]: <kdb>[main.cc]</kbd> main.cc using the new scene class]
+
+</div>
+
+Run the new program, you should get the same result as image 16 again.
```
trevordblack, line 2911: same result as before

I  cringe at the idea of specifying an image number. This could break as images as moved around or removed....

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
...
+
+
+Refactoring Camera Setup
+-------------------------
+Our camera has values that are either hard-coded, or can only be specified on construction. That
```
trevordblack, line 2916: "Our camera has values that are either hard-coded, or can only be specified on construction."
->
"Our camera class is currently a pain to use because we have a bunch of values that are either hard-coded into the code, or that can only be defined on camera construction"

it takes a second to figure out what the word "construction" means in the initial sentence.

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
...
+makes things difficult with the scene class, particularly if we want to start making scene-creating
+functions that also set up the camera. To make this easier, we'll allow camera settings to be
```
trevordblack, line 2918: "particularly if we want to start making scene-creating functions that also set up the camera."
->
"particularly if we want to use some of our scene values or functions when setting up the camera"

I think what you're trying to communicate is that camera class needs to be rewritten to allow for fore more runtime flexibility and interop with the new scene class. This paragraph needs to be less specific, more general, and a little more kid-friendly

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
...
+altered at will, and then perform the camera initialization explicitly with a call to
+`camera::initialize()`.
```
trevordblack, line 2920: We should stop and explain why initialization is very important.

Since we're decoupling the camera init from it's constructor, there is some private member variables that are determined once the programmer explicitly says that the public variables have been set.

If a reader doesn't fully understand what the `camera::initizalize()` function does, they could mistake it for a constructor of sorts, and do something like:

```
cam.intialize();
cam.fov = 40;
```

books/RayTracingInOneWeekend.html
```diff
@@ -2779,6 +2782,239 @@
...
+
+But before we do _that_, let's revisit the image aspect ratio. When we first introduced aspect ratio
+in section [Sending Rays Into the Scene], we commented that we'd be using the standard pixel spacing
+where the distance between pixels is the same horizontally as it is vertically. There _are_
+situations where pixel spacing differs in the horizontal and vertical directions, but they're rare,
+and we needn't consider them for our purposes. It's easy enough to modify your code if you need to
+do this.
```
trevordblack, line 2927: I like this paragraph!

src/TheNextWeek/main.cc
hollasch: These statements effectively set up default scene values for this program. Subsequent scene modifying functions may either use these defaults or override them.

books/RayTracingInOneWeekend.html
```diff
@@ -2839,43 +3077,40 @@
...
-
-        for (int j = image_height-1; j >= 0; --j) {
-        ...
+        scene_desc.render();
+    }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     [Listing [scene-wide-angle]: <kbd>[main.cc]</kbd> Scene with wide-angle camera]
 
-gives:
+This gives us this rendering:
```
trevordblack, line 3113: This gives us the rendering

books/RayTracingInOneWeekend.html
hollasch: ✓

books/RayTracingInOneWeekend.html
```diff
@@ -2996,8 +3243,8 @@
 Defocus Blur
 ====================================================================================================
 
-Now our final feature: defocus blur. Note, all photographers will call it “depth of field” so be
-aware of only using “defocus blur” among friends.
+Now our final feature: defocus blur. (Photographers will call it “depth of field”, so just use the
+term “defocus blur” among your raytracing friends.)
```
trevordblack, line 3247: I actually kind of like the original sentence. With modification:

Now our final feature: _defocus blur_. Note, photographers will call it _depth of field_ so be
sure to only use _defocus blur_ among friends.

books/RayTracingTheNextWeek.html
hollasch: Ooh — good catch!

hollasch: Might need to have a `camera.h` for each project in the future. Pain either way.

books/RayTracingInOneWeekend.html
hollasch: Awwwww.....

hollasch: Fun while it lasted.

trevordblack: If you want to keep it, just move it down

books/RayTracingInOneWeekend.html
hollasch: Cutting and pasting is exactly the point of this step. All of the code already exists in `main.cc`, and the work routines of the `scene` class are moved from elsewhere. The point of the listing is to show where these old chunks land in the new class.

books/RayTracingInOneWeekend.html
hollasch: Changed.

Yeah, this is all a retread of all the different schemes we discussed at length, ending up and the ultimately least-work brute force approach. Issues inevitably come in with the image number, and then we always had to relate the image number from their version back to the current symbolically-named image. Sigh.

books/RayTracingInOneWeekend.html
hollasch: > Our camera class is currently a pain to use because we have a bunch of values that are either hard-coded into the code, or that can only be defined at camera construction. That makes things difficult with the scene class, particularly if we want to write scene-creating functions that also set up the camera. To make this easier, we'll allow camera settings to be altered at will and at various points in the code, and then perform the camera initialization explicitly with a call to `camera::initialize()`.

books/RayTracingInOneWeekend.html
hollasch: See rewrite above.

books/RayTracingInOneWeekend.html
hollasch: Good point.

hollasch: I added an additional sentence, but the above is very unlikely, since we have an explicit `render()` function, which initializes the camera prior to sampling the scene.

books/RayTracingInOneWeekend.html
hollasch: ✔

books/RayTracingInOneWeekend.html
hollasch: Took this sentence, but kept “... among your raytracing friends.” I mean, some friends can be photographers. :)

hollasch:
> For all scene descriptor functions, e.g. random_spheres, cornell_box I think that it's more clear if the functions don't have pass-by-address parameters and are non-void
> 
>     void main() {
>       scene scene_desc = random_spheres();
>       scene_desc.render();
>     }
>
> I think it's a little easier to understand

Heh, you're talking about something like iteration 6 of 13. This was indeed one of my early approaches. It's yet another example where our code aims short of modern C++, and puts aside engineering for exposition.

To go with your approach, a modern codebase would create move constructors for `scene`, `hittable_list`, and perhaps `camera`. Move constructors would be an unnecessary complication for many of our readers, so we'd be left with copy construction for the return value. That's pretty painful, because the `scene` basically contains every single durable object we ever allocate and create, so every return would be quite expensive -- a copy of the entire world. Note that the scene class _doesn't_ use `std::shared_ptr<hittable_list>` in an effort to keep things simple. Perhaps if you fought me on this, you'd have a chance of winning, but then we should add an explicit `scene::add_geometry()` and `scene::add_light()` method that takes a `std::shared_ptr<hittable>`, because scene building functions should _not_ be messing with another class's shared pointer.

Also note that scene functions are _partial_. For example, look at the new default scene for book 2: `default_scene()`. This passes the scene object to `final_scene()` to set up the expensive fancy final render, then tweaks image width, samples per pixel and max depth to return a modified result (yielding a scene that takes a few minutes instead of a few hours). One could also imagine a set of different world scene building functions that all call a common scene camera setup function.

Given all this, scene builders need to take in an existing scene (new, defaulted, or partial), modify it, and return the result. So given all _that_, I originally had all the scene builders take in a scene reference, modify it, and then return the reference, mostly because-as you commented-it looked nice to say something like

    scene scene_desc;
    scene_desc = setup_world(scene_desc);
    scene_desc = setup_camera(scene_desc);

Indeed, this would also provide a way to do this in a quasi fluid style:

    scene scene_desc;
    default_scene(final_scene(scene_desc));

In the end, I figured that scene build flow always starts with a clear scene owner, possibly setting some default values for the program, passing it off to one or more builder methods, and then ultimately invoking render(). So an early iteration had:

    switch (0) {
        case 1:  random_spheres(scene_desc).render();     break;
        case 2:  two_spheres(scene_desc).render();        break;
        case 3:  two_perlin_spheres(scene_desc).render(); break;
        case 4:  earth(scene_desc).render();              break;
        case 5:  simple_light(scene_desc).render();       break;
        case 6:  cornell_box(scene_desc).render();        break;
        case 7:  cornell_smoke(scene_desc).render();      break;
        case 8:  final_scene(scene_desc).render();        break;
        default: default_scene(scene_desc).render();      break;
    }

and another had this:

    scene scene_desc;
    scene& result_scene;

    switch (0) {
        case 1:  result_scene = random_spheres(scene_desc);     break;
        case 2:  result_scene = two_spheres(scene_desc);        break;
        case 3:  result_scene = two_perlin_spheres(scene_desc); break;
        case 4:  result_scene = earth(scene_desc);              break;
        case 5:  result_scene = simple_light(scene_desc);       break;
        case 6:  result_scene = cornell_box(scene_desc);        break;
        case 7:  result_scene = cornell_smoke(scene_desc);      break;
        case 8:  result_scene = final_scene(scene_desc);        break;
        default: result_scene = default_scene(scene_desc);      break;
    }

    result_scene.render();

But it just seemed simpler to do this:

    switch (0) {
        case 1:  random_spheres(scene_desc);     break;
        case 2:  two_spheres(scene_desc);        break;
        case 3:  two_perlin_spheres(scene_desc); break;
        case 4:  earth(scene_desc);              break;
        case 5:  simple_light(scene_desc);       break;
        case 6:  cornell_box(scene_desc);        break;
        case 7:  cornell_smoke(scene_desc);      break;
        case 8:  final_scene(scene_desc);        break;
        default: default_scene(scene_desc);      break;
    }

    scene_desc.render();

You _could_ do this:

    scene scene_desc;

    switch (0) {
        case 1:  scene_desc = random_spheres();     break;
        case 2:  scene_desc = two_spheres();        break;
        case 3:  scene_desc = two_perlin_spheres(); break;
        case 4:  scene_desc = earth();              break;
        case 5:  scene_desc = simple_light();       break;
        case 6:  scene_desc = cornell_box();        break;
        case 7:  scene_desc = cornell_smoke();      break;
        case 8:  scene_desc = final_scene();        break;
        default: scene_desc = default_scene();      break;
    }

    scene_desc.render();

But now in addition to the overhead of the everything-in-the-world copy (or use `std::unique_ptr` or `std::shared_ptr`), you lose the ability to do default scene initialization and you lose partial construction. It's also a break from our current philosophy of "just expose everything as a bag of public properties and some associated methods".

Phew. As you can see, fun ride on the playground, but you're still a bit dizzy when you get off.

trevordblack (approved):
LGTM

hollasch:
Thanks for the review Trevor; I really appreciate it.

hollasch:
I'll merge in the latest and submit later on. Heading out for Thanksgiving shopping.

PR #812, opened 2020-11-24
aabb ctor treats 2 params as extrema in any order

### LINKED ISSUE #733: `bvh_node` is broken with negative-radius spheres
I wanted to create some "Matryoshka hollow glass spheres", so I setup this scene:

```c++
    hittable_list world;

    const color ground_yellow(0.8, 0.8, 0);
    const color gold(0.8, 0.6, 0.2);
    const color blue(0.1, 0.2, 2);
    const color green(0.1, 0.9, 0.3);
    const double glass_refractive_index = 1.5;
    const auto ground_mat = make_shared<lambertian>(ground_yellow);
    const auto blue_mat   = make_shared<lambertian>(blue);
    const auto left_mat   = make_shared<dielectric>(glass_refractive_index);
    const auto center_mat = make_shared<lambertian>(green);
    const auto right_mat  = make_shared<metal>(gold, 1);
    const double tiny_y = 0.005;

    world.add(make_shared<sphere>(point3(-1,  tiny_y, -1),  0.5,  left_mat));
    world.add(make_shared<sphere>(point3(-1,  tiny_y, -1), -0.35, left_mat));
    world.add(make_shared<sphere>(point3(-1,  tiny_y, -1),  0.3,  left_mat));
    world.add(make_shared<sphere>(point3(-1,  tiny_y, -1), -0.2,  left_mat));
    world.add(make_shared<sphere>(point3(-1,  tiny_y, -1),  0.15, left_mat));
    world.add(make_shared<sphere>(point3(-1,  tiny_y, -1),  0.1,  blue_mat));
    world.add(make_shared<sphere>(point3( 0,  0.15 , -1),   .5,  center_mat));
    world.add(make_shared<sphere>(point3( 1,  tiny_y, -1),  0.5,  right_mat));
    world.add(make_shared<sphere>(point3( 0, -100.5, -1),   100,  ground_mat));   // The ground (Big boi)

    const bvh_node world_bvh(world, 0, 1); 

    point3 lookfrom(0, 0.25, 3.25);
    point3 lookat(0, 0, -1);
    auto vfov = 25.0;
    auto aperture = 0.0;
    color background = color(0.70, 0.80, 1.00);

    // Camera

    const vec3 vup(0,1,0);
    const auto dist_to_focus = (lookfrom - lookat).length();
    const int image_height = static_cast<int>(image_width / aspect_ratio);

    camera cam(lookfrom, lookat, vup, vfov, aspect_ratio, aperture, dist_to_focus, 0.0, 1.0);
```

When rendering on `world`, I get this, which is correct:
three_spheres_no_bvh

But when I render with `world_bvh`, this is the result:
three_spheres_bvh

I don't think that's right.

hollasch:
The original constructor expected that the first parameter was the
minimum point, and the second was the maximum. The new constructor now
selects the minimums of all coordinates in the two parameters, and the
maximums of all the coordinates. In this way, the two parameters are
just two extreme points of the bounds, so is significantly more robust.

This change was done because the original code had bugs when handling
spheres of negative radius (used to model glass spheres with voids).

Resolves #733

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Change Log -- Ray Tracing in One Weekend
   - Change: Class public/private access labels get two-space indents (#782)
   - Change: `interval::clamp()` replaces standalone `clamp` utility function
   - New: `rtw_image` class for easier image data loading, searches more locations (#807)
+  - Change: `aabb` class constructor treats two params as extreme points in any orientation (#733)
 
 ### In One Weekend
   - Added: More commentary about the choice between `double` and `float` (#752)
--- a/src/common/aabb.h
+++ b/src/common/aabb.h
@@ -17,7 +17,12 @@
 class aabb {
   public:
     aabb() {}
-    aabb(const point3& a, const point3& b) { minimum = a; maximum = b; }
+    aabb(const point3& a, const point3& b) {
+        // Treat the two points a and b as extrema for the bounding box, so we don't require a
+        // particular minimum/maximum coordinate order.
+        minimum = point3(fmin(a[0],b[0]), fmin(a[1],b[1]), fmin(a[2],b[2]));
+        maximum = point3(fmax(a[0],b[0]), fmax(a[1],b[1]), fmax(a[2],b[2]));
+    }
 
     point3 min() const {return minimum; }
     point3 max() const {return maximum; }


trevordblack:
Not sure how I feel about this change.

trevordblack:
1. The inline source in the book needs to be rewritten
2. I strongly recommend changing `surrounding_box` to a second `aabb` constructor that takes two `aabb` as input.

trevordblack (changes requested):
See above comments

trevordblack:
This is defo overkill, but this is a good place where a templated constructor could exist to create a bounding box that can accept 1 to N vec3 and 1 to N aabb

hollasch:
I'm trying to keep this change surgical to specifically address the weakness exposed with negative radius spheres. Personally, I don't like the negative radius hack to implement spheres of air — I feel like you should be able to express all of this with the index of refraction, “a sphere of glass in air“, or “a sphere of glass in water”, or “a sphere of air in glass”.

Still the attempts that one reader tried to model this failed for various mysterious reasons, where the negative radius trick work (and was suggested by Peter). The one catch was that the AABB failed because it passed inverted min/max points for the constructor.

I suggested this fix because there could be other situations where this effect happens, and mostly because there's no good reason to be this picky about the two extrema. We really shouldn't care the all coordinates of the first parameter should be less than all corresponding coordinates of the second parameter, when that's really a requirement for the current AABB implementation.

As an example, if I know the two extrema for a given object, and then apply a chain of transforms, I should be able to just pass the two transformed extrema, without requiring the modeling/transform code to fix things up. That way you don't require a fixup block for translate, and a fixup block for rotate, and a fixup block for scale and so forth (_and_ applied at each step of the chain).

This felt like a case of “be conservative in what you send, be liberal in what you accept”.

hollasch:
Also note that AABB is due for rework per other issues in v4.

trevordblack (approved):

PR #814, opened 2020-11-24
Remove virtual keyword for override methods

### LINKED ISSUE #805: Delete `virtual` keyword for overriding methods
Just had this come up at work: the `virtual` keyword is redundant with `override` or `final`. Either of the last two imply that the function is virtual.

Hence a common recommendation is that the first declaration of a virtual function get the `virtual` keyword, and that all derived classes use either `override` or `final` and do _not_ use the `virtual` keyword. Makes sense, and reduces unnecessary keyword clutter. I think we should adopt this convention.

hollasch:
If a method is declared with the `override` keyword, then it is
implicitly virtual, so the `virtual` keyword is redundant and can be
removed.

Resolves #805


trevordblack (commented):
Missed a `virtual` books/RayTracingTheNextWeek.html line 307

trevordblack (changes requested):
remove the last virtual, then LGTM

Issue #773, opened 2020-10-24
Redesign the aabb (axis-aligned bounding box) class


hollasch:
Specifically, we should support AABB values empty (or null) and universe:

- universe = <-∞, -∞, -∞> → <+∞, +∞, +∞>
- empty = <+∞, +∞, +∞> → <-∞, -∞, -∞>

I've used this formulation earlier in my career, and the math just flows through calculations beautifully. The only other change we'd need to make is to update aabb::hit(), aabb::area(), and aabb::longest_axis().

If we do this, we can rip out all of the boolean checks for whether a hittable has a bounding box. Instead, we enforce the requirement that all hittables have a bounding box.

We have a single place in our codebase where a hittable did not have a bounding box, and that was a hittable_list with zero components. In my proposed scheme, we should just return the empty AABB.

A lighter weight implementation of this could be just a special case where the minimum of any dimension is greater than the maximum of any dimension. We could add a method aabb::is_empty() to check for this. The lightweight implementation should be sufficient for our use case; the version with infinite values is mostly useful when computing the intersection and union of bounding boxes.

trevordblack:
This is conceptually very cool. I'd have to write out the code, or see it written out, to determine if it's easier/simpler/more learnable.

Having
> where the minimum of any dimension is greater than the maximum of any dimension
to represent an empty AABB is interesting. Zero is a special case in FP, so there shouldn't (!!) be round off error associated with an empty if max and min are both all zeroed out. But I think having an explicit definition for empty will be valuable.
whether that is:
<+∞, +∞, +∞> → <-∞, -∞, -∞>
OR
<0, 0, 0> → <0, 0, 0>

Hell, maybe we can get some of those "signed zeros" I keep hearing about in the literature (twitter)

hollasch:
Actually, <0,0,0>→<0,0,0> is the bounding box for the single point <0,0,0>, which is very different from the empty bounding box.

For example, if b1 = <10,10,10>→<11,11,11>, b2 = <0,0,0>→<0,0,0>, and b3 = _empty_ = <+∞,+∞,+∞>→<-∞,-∞,-∞>:

- intersection(b1, b2) = <10,10,10>→<0,0,0> = _empty_
- intersection(b1, _empty_) = <+∞,+∞,+∞>→<-∞,-∞,-∞> = _empty_
- union(b1, b2) = <0,0,0>→<11,11,11>
- union(b1, _empty_) = b1

[...]

PR #823, opened 2020-12-08
Aabb intervals

hollasch:
Axis-Aligned Bounding Box class now uses intervals for each axis.

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -261,6 +261,20 @@ just needs to become a function `center(time)`:
     [Listing [moving-sphere-hit]: <kbd>[moving_sphere.h]</kbd> Moving sphere hit function]
 </div>
 
+We need to implement the `interval::contains()` method mentioned above:
+
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+    class interval {
+      public:
+        ...
+        bool contains(double x) const {
+            return min <= x && x <= max;
+        }
+        ...
+    };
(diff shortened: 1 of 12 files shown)


hollasch:
This change also addresses #773

books/RayTracingTheNextWeek.html
```diff
@@ -1024,7 +1037,7 @@
         if (!a->bounding_box(0,0, box_a) || !b->bounding_box(0,0, box_b))
             std::cerr << "No bounding box in bvh_node constructor.\n";
 
-        return box_a.min().e[axis] < box_b.min().e[axis];
+        return box_a.axis(axis).min < box_b.axis(axis).min;
```
trevordblack, line 1040: maybe make it more explicit that `axis()` is the function that returns an interval, and `axis` is an int tells that function what to return.
Maybe change the `axis` variable to `a` or `axis_int`

src/TheNextWeek/bvh.h
```diff
@@ -49,7 +49,7 @@ inline bool box_compare(const shared_ptr<hittable> a, const shared_ptr<hittable>
     if (!a->bounding_box(0,0, box_a) || !b->bounding_box(0,0, box_b))
         std::cerr << "No bounding box in bvh_node constructor.\n";
 
-    return box_a.min().e[axis] < box_b.min().e[axis];
+    return box_a.axis(axis).min < box_b.axis(axis).min;
```
trevordblack, line 52: See comment above

src/TheRestOfYourLife/bvh.h
```diff
@@ -49,7 +49,7 @@ inline bool box_compare(const shared_ptr<hittable> a, const shared_ptr<hittable>
     if (!a->bounding_box(0,0, box_a) || !b->bounding_box(0,0, box_b))
         std::cerr << "No bounding box in bvh_node constructor.\n";
 
-    return box_a.min().e[axis] < box_b.min().e[axis];
+    return box_a.axis(axis).min < box_b.axis(axis).min;
```
trevordblack, line 52: see comments above

trevordblack (approved):

books/RayTracingTheNextWeek.html
hollasch: Yeah, I can do better than what I've got.

PR #829, opened 2020-12-09
Update book.css

### LINKED ISSUE #826: Code appears to come out of the box
Code in Ray Tracing in One Weekend seems to be appearing out of the box(see screenshot below).
A way you could fix this is to replace
``
overflow: visible;
``
with
``
overflow-x: scroll;
``
in
```
.md pre.listing.tilde {
    width: 96%;
    border-width: 3px;
    border-style: solid;
    border-color: rgb(212, 212, 212);
    border-image: initial;
    padding: 1.5ex;
    overflow: scroll;
    background: rgb(228, 228, 224);
}
```
in book.css.
<br>
Screenshot

JC-ProgJava:
Updates book.css to make code blocks centered and span out when text overflows.
Fixes #826.

### DIFF
--- a/style/book.css
+++ b/style/book.css
@@ -81,11 +81,8 @@ div.indented {
 .md pre.listing.tilde {
     border: solid 3px #d4d4d4;
     padding: 1.5ex;
-    width: fit-content !important;
-    display: inline-block;
-    position: relative;
-    left: 50%;
-    transform: translateX(-50%);
+    min-width: 96%;
+    width: fit-content;
     background: #e4e4e0;
 }
 
--- a/style/book.css
+++ b/style/book.css
@@ -81,8 +81,11 @@ div.indented {
 .md pre.listing.tilde {
     border: solid 3px #d4d4d4;
     padding: 1.5ex;
-    width: 96%;
-    overflow: visible;
+    width: fit-content !important;
+    display: inline-block;
+    position: relative;
+    left: 50%;
+    transform: translateX(-50%);
     background: #e4e4e0;
 }
 


hollasch (changes requested):
These changes don't work well, at least in my browser.

For one thing, the right edge of all listing boxes "goes ragged" — it expands to the right for longer lines, but also shrinks down for shorter lines.

The centering properties make this look worse, as now listings align randomly on the left margin, according to the longest line in the box.

Finally, I generally don't like `!important` tags in CSS, as it tends to highlight a layout that is brittle in some fashion when working with other styles, instead of cascading.

All that said, your changes did lead me quickly to this solution:

    .md pre.listing.tilde {
        border: solid 3px #d4d4d4;
        padding: 1.5ex;
        min-width: 96%;
        width: fit-content;
        background: #e4e4e0;
    }

Can you run that by your gauntlet and see if that works in your set? If it does, just add another commit and push (it would be good to preserve your original solution).

JC-ProgJava:
> For one thing, the right edge of all listing boxes "goes ragged" — it expands to the right for longer lines, but also shrinks down for shorter lines.

I see what you mean. I wanted to fit all the code inside of the box, so the boxes resize their width according to the code in them. I centered each box because they didn't align with the listing caption and text below them.

> The centering properties make this look worse, as now listings align randomly on the left margin, according to the longest line in the box.
> 
> Finally, I generally don't like `!important` tags in CSS, as it tends to highlight a layout that is brittle in some fashion when working with other styles, instead of cascading.
> 
> All that said, your changes did lead me quickly to this solution:
> 
> ```
> .md pre.listing.tilde {
>     border: solid 3px #d4d4d4;
>     padding: 1.5ex;
>     min-width: 96%;
>     width: fit-content;
>     background: #e4e4e0;
> }
> ```

I don't think the `!important` tag is necessary, so you can remove that. The edited version seems to work quite well, aligning with the left margin. But the code boxes that need the `width: fit-content;` attribute to fit all code inside can't seem to align with all the other code boxes. It looks 'displaced' when with the listing caption. See a screenshot below:
Screen Shot 2020-12-11 at 1 26 23 PM
(I zoomed out so you can see both boxes)
Above, the second code box aligns with the previous code box on the left margin, but its right margin stretches to the side, which makes the text below it seem displaced. Is this fine? If it is, I'll go ahead and update the PR. If not, I have another solution that combines both and seems to look better:
```
.md pre.listing.tilde {
    border: solid 3px #d4d4d4;
    padding: 1.5ex;
    width: fit-content;
    display: inline-block;
    min-width: 96%;
    position: relative;
    left: 50%;
    transform: translateX(-50%);
    background: #e4e4e0;
}
```
Output: (zoomed out)
Screen Shot 2020-12-11 at 1 34 19 PM
Here, the normal code boxes align, and the ones that stretch outwards are centered. Does it look better?

hollasch:
Doesn't work in my tests. Try using a standard desktop browser, and then setting the zoom level to something like 300%. You get a full flush-left horizontal scrollbar, but the code listing runs off the page to the left, and you can't scroll there.

Generally, I don't like the centering aspect, as that means now I have to scroll one direction to find the beginning of the code lines, and then the _other_ direction to find the end of the code lines. This is made worse by the fact that as coded, I can't even scroll to the beginning of the lines.

hollasch:
Wait, I just tested against your latest version, which you may not have changed. Perhaps you were expecting me to try out the CSS in your comments. I'll do that ...

hollasch:
Ok, just tested the CSS in your comment (not in your branch), and it does indeed have the flaws I mentioned above:

1. Code now overflows _two_ margins instead of just one.
2. I have no way to scroll left to read the code that overflows the left margin (though I can scroll to read the code the overflows the right margin).

Again, I'm testing with long lines and browser zooms of 200% to 300%. There may well be other scenarios that exercise overflow in different ways.

Typographically, a consistent left margin is vital, and I won't give up on that. There's a reason why centered text is considered poor typographic layout — it loses the anchoring point when scanning a new line. Ragged right is fine. For general text, lines should be limited to 1-2 alphabets in line length (again, to facilitate eye-movement when scanning), though code is a (slightly) different beast.

Given all the above, I like your changes to overflow to the right. Preserve a consistent left margin as outlined above and we're good.

JC-ProgJava:
Ok then, I have updated it.

hollasch (approved):

hollasch:
Thank you for catching and fixing!

hollasch:
Oh, one more thing — if you wouldn't mind, let me know your full name and I'll add you to the acknowledgements section.

JC-ProgJava:
> Oh, one more thing — if you wouldn't mind, let me know your full name and I'll add you to the acknowledgements section.

I don't really care if I'm attributed for the fix, but if you really need to do it, you can just use 'JC-ProgJava' and just link my profile or something like that. :)

PR #836, opened 2020-12-16
Tweak CMake settings: no version, better C++ ver

hollasch:
- The prior CMakeLists.txt file specified a version for the project,
  which we hadn't been updating. This isn't of much use in our project,
  since users are expected to change this program all the time. Thus,
  pulling any explicit version stamping.

- When specifying the C++ version (we use C++11 currently), you should
  set all three CMake variables CMAKE_CXX_STANDARD,
  CMAKE_CXX_STANDARD_REQUIRED, and CMAKE_CXX_EXTENSIONS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,13 +6,12 @@
 
 cmake_minimum_required ( VERSION 3.1.0 )
 
-project ( RTWeekend
-  VERSION 3.0.0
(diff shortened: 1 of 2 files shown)


trevordblack (approved):

hollasch:
Gah, this was accidentally targeted to master. Will redirect to dev-major.

PR #849, opened 2021-01-21
Fix hittable when hittable_list is meant.

NAThompson:


NAThompson:
This is incorrect since hittable_list inherits from hittable; my bad!

PR #854, opened 2021-02-03
Tweaks

hollasch:
Random tweaks

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -2062,7 +2062,7 @@ that the pointer is to a class, which the “class material” in the hittable c
         point3 p;
         vec3 normal;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-        shared_ptr<material> mat_ptr;
+        shared_ptr<material> mat;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
         double t;
         bool front_face;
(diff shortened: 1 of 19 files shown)


trevordblack:
Near as I can tell, The changelog needs to have the changes made from surface texture and possibly rotate_y

hollasch:
Regarding the changelog, see the single item under `The Next Week`. For `rotate_y`, are you talking about the bug I just fixed? That was introduced during `dev-major` development, so wouldn't get an entry in the changelog.

trevordblack:
Sounds good.

books/RayTracingInOneWeekend.html
```diff
@@ -1842,7 +1842,8 @@
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     if (world.hit(r, interval(0.001, infinity), rec)) {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [reflect-tolerance]: <kbd>[main.cc]</kbd> Calculating reflected ray origins with tolerance]
+    [Listing [reflect-tolerance]: <kbd>[main.cc]</kbd>
+    Calculating reflected ray origins with tolerance]
```
trevordblack, line 1846: Can/should this be indented?

trevordblack (approved):
Good change.

These might seem like a bunch of small changes, but they really help remove a lot of the clutter in the source.

books/RayTracingInOneWeekend.html
hollasch: Yeah, I tried it, but honestly it all looks awkward. It's sensitive to which characters the line breaks on, and the multiple different brackets just make it look even more awkward. Flush left was the best I found.

PR #855, opened 2021-02-03
Time

hollasch:
Initial draft. This updates the code, but lacks the corresponding text changes.

I believe that we can also eliminate the `bool` return from the `hittable::bounding_box()` methods, as I couldn't find a single derived that didn't (1) return true, or (2) return the logical-and of its child returns. Been wanting to fix this for a long time, as it would be much cleaner to just return the `aabb`. I'd like to introduce the concept of _empty_ or _universe_ to AABB if this is necessary, but I don't think it is.

Also note that in the text changes, I plan to comment on how one might extend time management more generally, as the original (incomplete) design attempted. Not planning on going deep; just enumerating some of the aspects that should be considered.

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -39,28 +39,27 @@ at the end of this book.
 
 Motion Blur
 ====================================================================================================
-When you decided to ray trace, you decided that visual quality was worth more than run-time. In your
-fuzzy reflection and defocus blur you needed multiple samples per pixel. Once you have taken a step
-down that road, the good news is that almost all effects can be brute-forced. Motion blur is
-certainly one of those. In a real camera, the shutter opens and stays open for a time interval, and
-the camera and objects may move during that time. Its really an average of what the camera sees over
-that interval that we want.
+When you decided to ray trace, you decided that visual quality was worth more than run-time. When
+rendering fuzzy reflection and defocus blur, we used multiple samples per pixel. Once you have taken
+a step down that road, the good news is that almost _all_ effects can be similarly brute-forced.
+Motion blur is certainly one of those.
+
+In a real camera, the shutter remains open for a short time interval, during which the camera and
+objects in the world may move. To accurately reproduce such a camera shot, we seek an average of
+all the instant images that the camera perceives while its shutter is open to the world.
 
 
 Introduction of SpaceTime Ray Tracing
 --------------------------------------
-We can get a random estimate by sending each ray at some random time when the shutter is open. As
-long as the objects are where they should be at that time, we can get the right average answer with
-a ray that is at exactly a single time. This is fundamentally why random ray tracing tends to be
-simple.
-
-The basic idea is to generate rays at random times while the shutter is open and intersect the model
-at that one time. The way it is usually done is to have the camera move and the objects move, but
-have each ray exist at exactly one time. This way the “engine” of the ray tracer can just make sure
-the objects are where they need to be for the ray, and the intersection guts don’t change much.
+We can get a random estimate of a single photon by sending a single ray at some random instant in
+time when the shutter is open. As long as the objects are where they should be at that instant, we
+can get an accurate measure of the light for that ray at that same instant. This is yet another
+example of how random (Monte Carlo) ray tracing ends up being quite simple. Brute force wins again!
 
 <div class='together'>
-For this we will first need to have a ray store the time it exists at:
+Since the “engine” of the ray tracer can just make sure the objects are where they need to be for
+each ray, the intersection guts don’t change much. To accomplish this, we need to store the exact
+time for each ray:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class ray {
@@ -82,7 +81,7 @@ For this we will first need to have a ray store the time it exists at:
             return orig + t*dir;
         }
 
-      public:
+      private:
         point3 orig;
         vec3 dir;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
@@ -94,10 +93,35 @@ For this we will first need to have a ray store the time it exists at:
 </div>
 
 
+Managing Time
+--------------
+Before continuing, let's think about time, and how we might manage it across one or more renders.
+There are two aspects of shutter timing to think about: the time from one shutter opening to the
+next shutter opening, and how long the shutter stays open for each frame. Standard movie film used
+to be shot at 24 frames per second. Modern digital films can be 30 or even 60 frames per second.
+Each frame can have its own shutter speed, which need not be -- and typically isn't -- the maximum
+duration of the entire shutter period. You could have the shutter open for 1/1000th of a second per
+frame, or 1/60th of a second. It may surprise you to learn that for most film motion pictures, the
+screen is actually dark more than it is lit up with the movie frames!
+
+If you wanted to render a sequence of images, you would need to set up the camera with the
+appropriate shutter timings: frame-to-frame period, shutter/render duration, and the total number of
+frames (or total shot time). If the camera is moving and the world is static, you're good to go.
+However, if anything in the world is moving, you would need to add a method to `hittable` to
+broadcast the current shot timing to every object in the world. This method would then provide a way
+for all animate objects to set up their motion during that frame.
+
+This is fairly straight-forward, and definitely a fun avenue for you to experiment with if you wish.
+However, for our purposes right now, we're going to proceed with a drastically simplified model. We
+will be render only a single frame, assuming a start at time = 0 and ending at time = 1. Our first
+task is to modify the camera to launch rays with random times in $[0,1]$, and our second task will
+be the creation of an animate sphere class.
+
+
 Updating the Camera to Simulate Motion Blur
 --------------------------------------------
-Now we need to modify the camera to generate rays at a random time between the start time and the
-end time. Should the camera keep track of the time interval, or should that be up to the user of
+We need to modify the camera to generate rays at a random instant between the start time and the end
+time. Should the camera keep track of the time interval, or should that be up to the user of the
 camera when a ray is created? When in doubt, I like to make constructors complicated if it makes
 calls simple, so I will make the camera keep track, but that’s a personal preference. Not many
 changes are needed to camera because for now it is not allowed to move; it just sends out rays over
@@ -130,9 +154,9 @@ a time period.
 
 Adding Moving Spheres
 ----------------------
-We also need a moving object. I’ll create a sphere class that has its center move linearly from
-`center0` at `time_start` to `center1` at `time_end`. Outside that time interval it continues on, so
-those times need not match up with the camera aperture open and close.
+Now to create a moving object. I’ll create a sphere class that has its center move linearly from
+`center0` at time 0 to `center1` at time 1. (It continues on outside that time interval, so it
+really can be sampled at any time.)
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     #ifndef MOVING_SPHERE_H
@@ -147,17 +171,10 @@ those times need not match up with the camera aperture open and close.
         moving_sphere() {}
         moving_sphere(point3 c0, point3 c1, double r, shared_ptr<material> m)
           : center0(c0), center1(c1), center_vec(c1 - c0), radius(r), mat_ptr(m)
-        {
-            // To Be Implemented
-        };
+        { };
 
         bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
-            // To Be Implemented
-        }
-
-        bool bounding_box(aabb& output_box) const override {
-            output_box = bbox;
-            return true;
+            // Implementation below.
         }
 
         point3 center(double time) const {
@@ -171,7 +188,6 @@ those times need not match up with the camera aperture open and close.
         vec3 center_vec;
         double radius;
         shared_ptr<material> mat_ptr;
-        aabb bbox;
     };
 
     #endif
@@ -226,7 +242,7 @@ just needs to become a function `center(time)`:
     [Listing [moving-sphere-hit]: <kbd>[moving_sphere.h]</kbd> Moving sphere hit function]
 </div>
 
-We need to implement the `interval::contains()` method mentioned above:
+We need to implement the new `interval::contains()` method mentioned above:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class interval {
@@ -305,9 +321,8 @@ Putting Everything Together
 ----------------------------
 <div class='together'>
 The code below takes the example diffuse spheres from the scene at the end of the last book, and
-makes them move during the image render. (Think of a camera with shutter opening at time 0 and
-closing at time 1.) Each sphere moves from its center $\mathbf{C}$ at time $t=0$ to $\mathbf{C} +
-(0, r/2, 0)$ at time $t=1$, where $r$ is a random number in $[0,1)$:
+makes them move during the image render. Each sphere moves from its center $\mathbf{C}$ at time
+$t=0$ to $\mathbf{C} + (0, r/2, 0)$ at time $t=1$:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     ...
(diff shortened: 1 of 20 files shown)


trevordblack (commented):
Ahhh. I see.

So a bounding box is
1. Created at construction time, and is constant for the duration of the program
2. An immutable object that is constant for any ray_time in the simulation

I feel like this might have been a thing that we mutually brainstormed.

On the one hand, the reduction in "flexibility" is kind of sad.
But, this is just such a monumental improvement in simplicity and implementation that it's not even a contest.

That, and the initial impl was broken in subtle ways anyway.

books/RayTracingTheNextWeek.html
```diff
@@ -106,64 +106,22 @@
...
-            time0 = time_start;
-            time1 = time_end;
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-        }
-
+        ...
         ray get_ray(double s, double t) const {
             vec3 rd = lens_radius * random_in_unit_disk();
             vec3 offset = u * rd.x() + v * rd.y();
+            const auto ray_time = random_double(0.0, 1.0);
```
trevordblack, line 113: When writing text, this needs to be justified (but I figured you already know this)

books/RayTracingTheNextWeek.html
```diff
@@ -187,27 +145,33 @@
...
-            point3 ctr0, point3 ctr1, double r, shared_ptr<material> m,
-            double time_start, double time_end)
-            :
-            center0(ctr0), center1(ctr1), radius(r), mat_ptr(m),
-            time0(time_start), time1(time_end)
-        {};
+        moving_sphere(point3 c0, point3 c1, double r, shared_ptr<material> m)
+          : center0(c0), center1(c1), center_vec(c1 - c0), radius(r), mat_ptr(m)
+        {
+            // To Be Implemented
```
trevordblack, line 151: Remove. Change to:

```
        moving_sphere(point3 c0, point3 c1, double r, shared_ptr<material> m)
          : center0(c0), center1(c1), center_vec(c1 - c0), radius(r), mat_ptr(m)
          {}
```

The constructor is complete as far as this chapter is concerned.

books/RayTracingTheNextWeek.html
```diff
@@ -187,27 +145,33 @@
...
+        };
 
         bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
             // To Be Implemented
```
trevordblack, line 155: I actually don't like the "to be implemented" and would rather just have all functions that need to be implemented be written plainly, as a `{ }` and go back and denote changes with highlighting, like we've done so far.

Why the change?

trevordblack (approved):

books/RayTracingTheNextWeek.html
hollasch: I was thinking about this. I generally agree that increments are sufficient, and needn't leave hints at "more to come".

For the "hit" method, the listing ends with some additional explanatory text, and then immediately after the full hit function is revealed. It needs to return a `bool`, as well as properly setting the out parameters. I don't like the idea of making the reader parse dummy code, followed by a couple of paragraphs, followed by ripping it out and replacing it with the actual code.

For the constructor, that's just an awkward stopping point during this draft. At line 745, we show how the constructor still has to initialize the cached bounding box of the total possible bounds of the moving sphere. That it happens so much later is kind of crazy, and definitely needs to be fixed.

Note that I used to make these kinds of large changes in two phases: (1) the source code, debugged and working, and (2) all of the remaining text. In this case, I'm trying something a bit new: (1) running and debugged source code, (2) updated listings only, and (3) explanatory text. I'm hoping that this results in fewer changes missed between source and book, and this way I can use the listings as natural prompts for the text. But of course, exposition and development may change in the process (as here).

Looping back to your comment, I often lament the style of these sorts of articles and books. It's so easy to fall into the trap of just dictating the edits, step by step, and leaving no room for trial or experimentation. I kind of left "To Be Implemented" in as a possible prompt for the reader, like "hey, wanna give this a shot?". I don't know how useful or effective this is, and agree that it's clunky as it currently lies. We'll see how it ends up -- on to the text update now.

books/RayTracingTheNextWeek.html
```diff
@@ -39,28 +39,27 @@
...
-the camera and objects may move during that time. Its really an average of what the camera sees over
-that interval that we want.
+When you decided to ray trace, you decided that visual quality was worth more than run-time. When
+rendering fuzzy reflection and defocus blur, we used multiple samples per pixel. Once you have taken
+a step down that road, the good news is that almost _all_ effects can be similarly brute-forced.
+Motion blur is certainly one of those.
+
+In a real camera, the shutter remains open for a short time interval, during which the camera and
+objects in the world may move. To accurately reproduce such a camera shot, we seek an average of
+all the instant images that the camera perceives while its shutter is open to the world.
```
trevordblack, line 49: "an average of all the instant images"
vs
"an average of what the camera sees over the interval"

Your version is more explicit. And more explicit is usually more good.
But I prefer Shirley's because it allows the reader to come up with their own mental model for how motion blur works.

Your text is explicit that a camera shot is the integral of infinitesimal images captured over the shutter period.
My (different) mental model is that a camera shot is the integral of each individual photon over the shutter period.

Both are equally valid, but your text is choosing the mental model for me :(

trevordblack: So I guess what I'm trying to say is that your revision is worse by being _better_

books/RayTracingTheNextWeek.html
```diff
@@ -39,28 +39,27 @@
...
-We can get a random estimate by sending each ray at some random time when the shutter is open. As
-long as the objects are where they should be at that time, we can get the right average answer with
-a ray that is at exactly a single time. This is fundamentally why random ray tracing tends to be
-simple.
-
-The basic idea is to generate rays at random times while the shutter is open and intersect the model
-at that one time. The way it is usually done is to have the camera move and the objects move, but
-have each ray exist at exactly one time. This way the “engine” of the ray tracer can just make sure
-the objects are where they need to be for the ray, and the intersection guts don’t change much.
+We can get a random estimate of a single photon by sending a single ray at some random instant in
```
trevordblack, line 54: Since we're using sRGB (super duper simplification *grumble* *grumble*), any one of our rays is actually a linear mixture of photons. We cannot, as the text outlines, get a "random estimate" of a single photon.

Wrong, but definitely a *good* change.

trevordblack: Thinking of a single ray as a single photon is --in my opinion-- an immensely fruitful mental model, so I recommend that forever. (See Above)

books/RayTracingTheNextWeek.html
```diff
@@ -39,28 +39,27 @@
...
+time when the shutter is open. As long as the objects are where they should be at that instant, we
```
trevordblack, line 55: time while the shutter is open

trevordblack, line 55: "As long as the objects are where they should be at that instant"
->
"As long as we can determine where the objects should be at that instant"

books/RayTracingTheNextWeek.html
```diff
@@ -39,28 +39,27 @@
...
+can get an accurate measure of the light for that ray at that same instant. This is yet another
+example of how random (Monte Carlo) ray tracing ends up being quite simple. Brute force wins again!
```
trevordblack, line 57: This is actually the first instance of the term "Monte Carlo" in the books.
Consider removing

books/RayTracingTheNextWeek.html
```diff
@@ -39,28 +39,27 @@
...
 
 <div class='together'>
-For this we will first need to have a ray store the time it exists at:
+Since the “engine” of the ray tracer can just make sure the objects are where they need to be for
```
trevordblack, line 60: "can just make sure the objects are where they need to be"
->
"can determine where the objects are supposed to be"

books/RayTracingTheNextWeek.html
```diff
@@ -82,7 +81,7 @@
             return orig + t*dir;
         }
 
-      public:
+      private:
```
trevordblack, line 84: #define private public

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
 </div>
 
 
+Managing Time
+--------------
+Before continuing, let's think about time, and how we might manage it across one or more renders.
```
trevordblack, line 98: manage it across one or more  successive renders.

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+There are two aspects of shutter timing to think about: the time from one shutter opening to the
+next shutter opening, and how long the shutter stays open for each frame. Standard movie film used
```
trevordblack, line 100: ", and the time the shutter stays open for each frame"

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+to be shot at 24 frames per second. Modern digital films can be 30 or even 60 frames per second.
```
trevordblack, line 101: "Standard movie film used to be shot at 24 frames per second"

Do you mean
1. Film - the photochemical sandwich on plastic substrate
2. Film - A movie

Because you use
"Modern digital films" to mean "modern digital movies"

You might be using the two different meanings of "film" across these two sentences -> confusing.

trevordblack, line 101: "Modern films are only limited by the benevolence of their directors, and can be found in 30, 60, 120, or even 48 frames per second. Yes! Even 48 frames per second, who would have thought of it!"

Something like that.

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+Each frame can have its own shutter speed, which need not be -- and typically isn't -- the maximum
```
trevordblack, line 102: it's rare that you'll have one frame be a different shutter speed from an adjacent frame if they're from the same take.

So, "Each frame can have its own shutter speed" is a bit overkill, but it's useful-sorta-right, rather than misleading-sorta-right

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+duration of the entire shutter period. You could have the shutter open for 1/1000th of a second per
+frame, or 1/60th of a second. It may surprise you to learn that for most film motion pictures, the
+screen is actually dark more than it is lit up with the movie frames!
```
trevordblack, line 105: With tech like Black frame insertion and similar, the amount of time that a frame is lit up has more to do with brightness levels than shutter speeds.

Shutter speeds are orthogonal to frame on time.
If you think about it, unless the codec has metadata, the TV has no way of knowing what the shutter speed of any given frame was.

And if the shutter is closed more than it is open, it's due to a quirk of the way films are shot. (I will spare you the specifics)

Consider rewriting to
"It may surprise you to learn that for most motion pictures, the shutter is actually closed as much -- if not more -- than it is open"

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+
+If you wanted to render a sequence of images, you would need to set up the camera with the
+appropriate shutter timings: frame-to-frame period, shutter/render duration, and the total number of
```
trevordblack, line 108: shutter period?

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+frames (or total shot time). If the camera is moving and the world is static, you're good to go.
```
trevordblack, line 109: (total shot time)

trevordblack, line 109: If the camera is moving and the world is static, you're good to go
->
If the camera is moving and the world is static, we wouldn't need to make any code changes.

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+However, if anything in the world is moving, you would need to add a method to `hittable` to
```
trevordblack, line 110: "you would need to add a method to `hittable` to ..."
->
"you would need to rewrite `hittable` so that every object was aware of the time of a ray. This could be done with ..."

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+broadcast the current shot timing to every object in the world. This method would then provide a way
+for all animate objects to set up their motion during that frame.
```
trevordblack, line 112: animated

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+
+This is fairly straight-forward, and definitely a fun avenue for you to experiment with if you wish.
+However, for our purposes right now, we're going to proceed with a drastically simplified model. We
```
trevordblack, line 115: "drastically simplified"
->
"drastically simple"
or
"simplified"

trevordblack: ehhh.

trevordblack: It's fine as-is

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+will be render only a single frame, assuming a start at time = 0 and ending at time = 1. Our first
```
trevordblack, line 116: "we will render only ..." ?

books/RayTracingTheNextWeek.html
```diff
@@ -94,10 +93,35 @@
...
+task is to modify the camera to launch rays with random times in $[0,1]$, and our second task will
+be the creation of an animate sphere class.
```
trevordblack, line 118: animated?

books/RayTracingTheNextWeek.html
hollasch: I think it's important to distinguish between the reality and the model. I agree with your take; will fix.

books/RayTracingTheNextWeek.html
```diff
@@ -130,9 +154,9 @@
 
 Adding Moving Spheres
 ----------------------
-We also need a moving object. I’ll create a sphere class that has its center move linearly from
-`center0` at `time_start` to `center1` at `time_end`. Outside that time interval it continues on, so
-those times need not match up with the camera aperture open and close.
+Now to create a moving object. I’ll create a sphere class that has its center move linearly from
+`center0` at time 0 to `center1` at time 1. (It continues on outside that time interval, so it
```
trevordblack, line 158: Consider:

time 0 -> start time
time 1 -> end time

trevordblack, line 158: It continues on indefinitely outside that time interval

books/RayTracingTheNextWeek.html
hollasch: “... a random estimate of a single (simplified) photon by sending ...”?

hollasch: I don't like presenting simplified models as true to reality. Always bugged me when I learned that electrons don't _really_ "orbit the nucleus like planets", and wished that the more complex reality was at least hinted at.

hollasch: A phoxel!

books/RayTracingTheNextWeek.html
hollasch: Philistine.

books/RayTracingTheNextWeek.html
hollasch: Changed to keep "film" a synonym for celluloid. Used movie for the composite, digital movie for digital only.

hollasch: Did a bunch of rewording.

books/RayTracingTheNextWeek.html
hollasch: You'll note that I made a mistake above, which confuses things even more. Corrected:

“... its own shutter speed, which need not be -- and typically isn't -- the maximum duration of the entire ~shutter period~ frame.”

My intent here was to really drive home the distinction between frame rate and shutter speed, because both would need to be managed if readers when that route and wanted to simulate physical cameras, and just to share/break some assumptions. But this is kind of rat-holing, and probably unnecessary.

> And if the shutter is closed more than it is open, it's due to a quirk of the way films are shot. (I will spare you the specifics)
I presume you're thinking of the circular shutter plate. Yeah, and it's really cool seeing the almost subliminal perception as you mess with these things. But I'm thinking now I should cut this short.

books/RayTracingTheNextWeek.html
hollasch: Correction: every object would need to be aware of the frame period (to compute the bounding box for that frame). The _hit_ method handles ray time.

hollasch: Unfortunately, this statement has to be a bit hand-wavy, since we haven't yet introduced bounding boxes.

books/RayTracingTheNextWeek.html
trevordblack: | Circular shutter plate

Not exactly. The circular shutter plate is the reason you typically want shutter time to be half of the frame time.
For good-looking motion blur.

No. The reason that some films are closed more than they are open is because hey shoot at 24 frames per second, but have a shutter speed of _1/50_ of a second.

Cinematography was a ***mistake***

books/RayTracingTheNextWeek.html
hollasch: Heh, just caught that myself.

books/RayTracingTheNextWeek.html
hollasch: I guess I like to say `animate` as an adjective. Sounds cool. To my ear... Oh all right. :)

books/RayTracingTheNextWeek.html
hollasch: We need to really make clear the implicit time values of 0 and 1, since we're coding to that. For clarify, I've replaced "time 0" and "time 1" with "time=0" and "time=1".

hollasch: Like it.

hollasch:
Follow-ups to https://github.com/RayTracing/raytracing.github.io/pull/861

PR #860, opened 2021-02-09
Bounding box

hollasch:


src/common/interval.h
```diff
@@ -18,6 +18,10 @@ class interval {
     interval(const interval& a, const interval& b)
       : min(fmin(a.min, b.min)), max(fmax(a.max, b.max)) {}
 
+    bool is_empty() const {
+        return max < min;
```
trevordblack, line 22: `is_empty()` and `interval::empty` are slightly different things. I don't think it'll create a problem, but my engineer brain pings when I see stuff like this.

hollasch: You mean the value versus the property? The is_empty property is true if the interval contains no points. Thus, there are infinitely many ways to construct an interval such that it contains no points, yet it's quite useful (and efficient) to have a canonically empty interval.

trevordblack: yeah, exactly.

hollasch: Ha. I think I needed `aabb::is_empty()` and then refactored so I didn't need it after all. And thus I don't need `interval::is_empty()`. Problem solved for now.

src/TheRestOfYourLife/hittable_list.h
```diff
@@ -43,20 +43,7 @@ class hittable_list : public hittable {
...
-
-        for (const auto& object : objects) {
-            if (!object->bounding_box(temp_box)) return false;
-            output_box = first_box ? temp_box : aabb(output_box, temp_box);
-            first_box = false;
-        }
-
-        return true;
-    }
+    aabb bounding_box() const override { return bbox; }
```
trevordblack, line 46: needs `add` function?

hollasch: _Yoiks_. Good catch. Also missing `aabb.h` include. I'll add a new finalization task to compare source from book to book (many files in books 2 & 3 are the same or obviously similar).

books/RayTracingTheNextWeek.html
```diff
@@ -456,13 +457,13 @@
...
-need to be called if precise) axis-aligned bounding boxes, or AABBs. Any method you want to use to
+need to be called if precise) _axis-aligned bounding boxes_, or AABBs. Any method you want to use to
 intersect a ray with an AABB is fine. And all we need to know is whether or not we hit it; we don’t
-need hit points or normals or any of that stuff that we need for an object we want to display.
+need hit points or normals or any of the stuff we need to display the object.
 
 <div class='together'>
 Most people use the “slab” method. This is based on the observation that an n-dimensional AABB is
-just the intersection of n axis-aligned intervals, often called “slabs” An interval is just the
+just the intersection of n axis-aligned intervals, often called “slabs”. An interval is just the
```
trevordblack, line 466: $n$

books/RayTracingTheNextWeek.html
```diff
@@ -534,7 +535,7 @@
     compute (tx0, tx1)
     compute (ty0, ty1)
     compute (tz0, tz1)
-    return overlap?( (tx0, tx1), (ty0, ty1), (tz0, tz1))
+    return overlap ? ((tx0, tx1), (ty0, ty1), (tz0, tz1))
```
trevordblack, line 538: good change.

books/RayTracingTheNextWeek.html
```diff
@@ -739,11 +750,9 @@
     [Listing [sphere-bbox]: <kbd>[sphere.h]</kbd> Sphere with bounding box]
 </div>
 
-<div class='together'>
-For `moving sphere`, we want the bounds of all places the moving sphere could occupy while it moves.
-To do this, we can take the box of the sphere at time = 0, and the box of the sphere at time = 1,
-and compute the box of those two boxes. We'll add a new `aabb` constructor that takes two boxes as
-input below.
+For `moving sphere`, we want the bounds of its entire range of motion. To do this, we can take the
+box of the sphere at time=0, and the box of the sphere at time=1, and compute the box of those two
```
trevordblack, line 754: is `time=0` the right formatting? I don't know.

books/RayTracingTheNextWeek.html
```diff
@@ -739,11 +750,9 @@
...
+boxes.
```
trevordblack, line 755: "compute the box of these two boxes"
->
"compute the box that contains both of these two boxes"
?

trevordblack (approved):
Lots of good changes in this one Steve

PR #864, opened 2021-02-13
DRAFT: For consideration, new `quad` primitive to replace `xy_rect`, `xz_rect`, `yz_rect`

hollasch:
DRAFT ONLY

Sharing my new `quad` and other planar primitives for comments. Design commentary flush left.

### DIFF
--- a/src/TheNextWeek/main.cc
+++ b/src/TheNextWeek/main.cc
@@ -339,10 +339,23 @@ void planar_test(scene& scene_desc) {
 ///// triangles, ellipses, annuli and other 2D shapes as possibilities. Hmmm, I haven't yet
 ///// done stencil for giggles; think I'll do that too for my own fun.
 
-    //auto thing = make_shared<quad>(point3(-2,-2,4), vec3(4,0,0), vec3(0,4,0), mat);
-    auto thing = make_shared<ellipse>(point3(0,0,0), vec3(2,0,0), vec3(0,1,0), mat);
-    //auto thing = make_shared<annulus>(point3(0,0,0), vec3(1,0,0), vec3(0,1,0), 1.0, 2.0, mat);
-    //auto thing = make_shared<tri>(point3(-2,-2,0), vec3(4,0,0), vec3(0,4,0), mat);
+    shared_ptr<hittable> thing;
+    switch (0) {
+        default:
+        case 0:
+            thing = make_shared<quad>(point3(-2,-2,4), vec3(4,0,0), vec3(0,4,0), mat);
+            break;
+        case 1:
+            thing = make_shared<ellipse>(point3(0,0,0), vec3(2,0,0), vec3(0,1,0), mat);
+            break;
+        case 2:
+            thing =
+                make_shared<annulus>(point3(0,0,0), vec3(1,0,0), vec3(0,1,0), 1.0, 2.0, mat);
+            break;
+        case 3:
+            thing = make_shared<tri>(point3(-2,-2,0), vec3(4,0,0), vec3(0,4,0), mat);
+            break;
+    }
 
     scene_desc.world.add(thing);
 }
--- a/src/TheNextWeek/main.cc
+++ b/src/TheNextWeek/main.cc
@@ -19,6 +19,7 @@
 #include "hittable_list.h"
 #include "material.h"
 #include "moving_sphere.h"
+#include "planar.h"
 #include "scene.h"
 #include "sphere.h"
 #include "texture.h"
@@ -318,6 +319,35 @@ void default_scene(scene& scene_desc) {
 }
 
 
+void planar_test(scene& scene_desc) {
+    scene_desc.image_width       = 400;
+    scene_desc.aspect_ratio      = 1.0;
+    scene_desc.samples_per_pixel = 4;
+    scene_desc.max_depth         = 4;
+
+    scene_desc.cam.aperture = 0.0;
+    scene_desc.cam.vfov     = 20.0;
+    scene_desc.cam.lookfrom = point3(0,0,12);
+    scene_desc.cam.lookat   = point3(0,0,0);
+
+    auto earth_texture = make_shared<image_texture>("earthmap.jpg");
+    auto mat = make_shared<lambertian>(earth_texture);
+
+    //scene_desc.world.add(make_shared<sphere>(point3(0,0,0), 2, earth_surface));
+
+///// Four new primitive types below. I'm leaning toward keeping only quads, but hinting at
+///// triangles, ellipses, annuli and other 2D shapes as possibilities. Hmmm, I haven't yet
+///// done stencil for giggles; think I'll do that too for my own fun.
+
+    //auto thing = make_shared<quad>(point3(-2,-2,4), vec3(4,0,0), vec3(0,4,0), mat);
+    auto thing = make_shared<ellipse>(point3(0,0,0), vec3(2,0,0), vec3(0,1,0), mat);
+    //auto thing = make_shared<annulus>(point3(0,0,0), vec3(1,0,0), vec3(0,1,0), 1.0, 2.0, mat);
+    //auto thing = make_shared<tri>(point3(-2,-2,0), vec3(4,0,0), vec3(0,4,0), mat);
+
+    scene_desc.world.add(thing);
+}
+
+
 int main() {
     scene scene_desc;
 
@@ -335,6 +365,8 @@ int main() {
         case 7:  cornell_smoke(scene_desc);      break;
         case 8:  final_scene(scene_desc);        break;
         default: default_scene(scene_desc);      break;
+
+        case 0: planar_test(scene_desc); break;
     }
 
     scene_desc.render();
--- a/src/TheNextWeek/planar.h
+++ b/src/TheNextWeek/planar.h
@@ -22,10 +22,10 @@
 class planar : public hittable {
   public:
     planar(const point3& o, const vec3& side_A, const vec3& side_B, shared_ptr<material> m)
-      : origin(o), a(side_A), b(side_B), mat(m)
+      : plane_origin(o), a(side_A), b(side_B), mat(m)
     {
         normal = unit_vector(cross(a, b));
-        D = -dot(normal, origin);
+        D = -dot(normal, plane_origin);
         a /= dot(a,a);
         b /= dot(b,b);
 
@@ -34,14 +34,14 @@ class planar : public hittable {
         interval u_bounds, v_bounds;
         get_uv_bounds(u_bounds, v_bounds);
         bbox = aabb(
-            origin + (u_bounds.min * side_A) + (v_bounds.min * side_B),
-            origin + (u_bounds.max * side_A) + (v_bounds.max * side_B)
+            plane_origin + (u_bounds.min * side_A) + (v_bounds.min * side_B),
+            plane_origin + (u_bounds.max * side_A) + (v_bounds.max * side_B)
         );
     }
 
     virtual void get_uv_bounds(interval& u_bounds, interval& v_bounds) {
-        u_bounds = interval::unit;
-        v_bounds = interval::unit;
+        u_bounds = interval(0,1);
+        v_bounds = interval(0,1);
     }
 
     virtual bool hit_uv(double u, double v) const = 0;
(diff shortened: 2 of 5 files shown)


trevordblack:
Error I'm getting when  building:
```
/mnt/d/raytracing.github.io/src/common/aabb.h:52:19: error: non-const lvalue reference to type 'interval' cannot bind to a temporary of type 'interval'
        interval& new_x = (x.size() >= delta) ? x : x.expand(delta);
                  ^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/d/raytracing.github.io/src/common/aabb.h:53:19: error: non-const lvalue reference to type 'interval' cannot bind to a temporary of type 'interval'
        interval& new_y = (y.size() >= delta) ? y : y.expand(delta);
                  ^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/d/raytracing.github.io/src/common/aabb.h:54:19: error: non-const lvalue reference to type 'interval' cannot bind to a temporary of type 'interval'
        interval& new_z = (z.size() >= delta) ? z : z.expand(delta);
                  ^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /mnt/d/raytracing.github.io/src/TheNextWeek/main.cc:22:
/mnt/d/raytracing.github.io/src/TheNextWeek/planar.h:78:10: warning: 'bounding_box' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    aabb bounding_box() const { return bbox; }
         ^
```
Do you mean
```
        interval new_x = (x.size() >= delta) ? x : x.expand(delta);
        interval new_y = (y.size() >= delta) ? y : y.expand(delta);
        interval new_z = (z.size() >= delta) ? z : z.expand(delta);
```

src/common/interval.h
```diff
@@ -28,14 +36,18 @@ class interval {
         return x;
     }
 
-    static const interval empty, universe;
+/// Toying with the idea of a global constant [0,1] interval, since it's used in multiple
+/// places. Alternate name: `interval::zero_one`.
```
trevordblack, line 40: normalized?

src/TheNextWeek/main.cc
```diff
@@ -318,6 +319,35 @@ void default_scene(scene& scene_desc) {
...
+
+    //scene_desc.world.add(make_shared<sphere>(point3(0,0,0), 2, earth_surface));
+
+///// Four new primitive types below. I'm leaning toward keeping only quads, but hinting at
+///// triangles, ellipses, annuli and other 2D shapes as possibilities. Hmmm, I haven't yet
+///// done stencil for giggles; think I'll do that too for my own fun.
+
+    //auto thing = make_shared<quad>(point3(-2,-2,4), vec3(4,0,0), vec3(0,4,0), mat);
+    auto thing = make_shared<ellipse>(point3(0,0,0), vec3(2,0,0), vec3(0,1,0), mat);
+    //auto thing = make_shared<annulus>(point3(0,0,0), vec3(1,0,0), vec3(0,1,0), 1.0, 2.0, mat);
```
trevordblack, line 344: Didn't know there was a word for that.
Neat.

src/TheNextWeek/planar.h
```diff
@@ -0,0 +1,158 @@
...
+        rec.p = intersection;
+        rec.u = u;
+        rec.v = v;
+        rec.mat = mat;
+        rec.set_face_normal(r, normal);
+
+        return true;
+    }
+
+    aabb bounding_box() const { return bbox; }
```
trevordblack, line 78: override

src/common/interval.h
```diff
@@ -18,6 +18,14 @@ class interval {
     interval(const interval& a, const interval& b)
       : min(fmin(a.min, b.min)), max(fmax(a.max, b.max)) {}
 
+    double size() const {
+        return max - min;
+    }
+
+    interval expand(double delta) const {
+        return interval(min - delta, max + delta);
```
trevordblack, line 26: should this be:
```
        return interval(min - delta/2, max + delta/2);
```

trevordblack: since it checks to see if it's less than delta, and if it is, makes it twice delta. Anything from 1 to 2 delta will remain as is

src/TheNextWeek/planar.h
```diff
@@ -0,0 +1,158 @@
...
+
+        interval u_bounds, v_bounds;
+        get_uv_bounds(u_bounds, v_bounds);
+        bbox = aabb(
+            origin + (u_bounds.min * side_A) + (v_bounds.min * side_B),
+            origin + (u_bounds.max * side_A) + (v_bounds.max * side_B)
+        );
+    }
+
+    virtual void get_uv_bounds(interval& u_bounds, interval& v_bounds) {
```
trevordblack, line 42: what is this function doing?
It doesn't seem like anything overrides it?

src/TheNextWeek/planar.h
```diff
@@ -0,0 +1,158 @@
...
+        a /= dot(a,a);
+        b /= dot(b,b);
+
+/////// By default, bounded to the world coordinates of UV<0,0> to UV<1,1>.
+
+        interval u_bounds, v_bounds;
+        get_uv_bounds(u_bounds, v_bounds);
+        bbox = aabb(
+            origin + (u_bounds.min * side_A) + (v_bounds.min * side_B),
+            origin + (u_bounds.max * side_A) + (v_bounds.max * side_B)
```
trevordblack, line 38: You mentioned problems with having an infinite plane.
As is, this plane appears to be finite.

src/TheNextWeek/planar.h
```diff
@@ -0,0 +1,158 @@
...
+        v_bounds = interval::unit;
+    }
+
+    virtual bool hit_uv(double u, double v) const = 0;
+
+/// The distinction of this method is that it calls `hit_uv()` to filter on UV coordinates.
+
+    bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
+        auto denom = dot(normal, r.direction());
+        if (fabs(denom) < 1e-8) return false;
```
trevordblack, line 53: return false on next line?

trevordblack: add comment explaining what this if statement does.

Otherwise, we'll need to build this block of source over a couple passes in the text

src/TheNextWeek/planar.h
```diff
@@ -0,0 +1,158 @@
...
+        rec.mat = mat;
+        rec.set_face_normal(r, normal);
+
+        return true;
+    }
+
+    aabb bounding_box() const { return bbox; }
+
+  protected:
+    point3 origin;
```
trevordblack, line 81: my brain defaulted to the origin as `(0,0,0)` which is not the case in general. Made parsing the code more difficult.
Consider renaming to something else.

Center? Modulus? Nexus? Terminus? I don't really have anything I particularly like

src/TheNextWeek/planar.h
```diff
@@ -0,0 +1,158 @@
...
+
+
+class quad : public planar {
+  public:
+    quad(const point3& o, const vec3& side_A, const vec3& side_B, shared_ptr<material> m)
+      : planar(o, side_A, side_B, m)
+    {}
+
+    bool hit_uv(double u, double v) const override {
+        return interval::unit.contains(u) && interval::unit.contains(v);
```
trevordblack, line 97: While I appreciate the simplicity, I think a more explicit check would be preferred here:
```
if (u > 1)
  return false;
if (v > 1)
  return false;
if (u < 0)
  return false;
if (v < 0)
  return false;
return true;
```

trevordblack:
Separate from my quibbles. I really like what you've done here. I think it was good to include ellipses, quads, and tris in this brainstorm. But I would have them removed from the final draft.

trevordblack (approved):

src/TheNextWeek/planar.h
hollasch: Oops. I'm messing with CMake now. Just figured out how to pass compile flags based on compiler. Will try to get this warning enabled for MSVC.

src/common/interval.h
hollasch: Nice catch; will fix.

src/TheNextWeek/planar.h
hollasch: I was thinking of derived (possibly infinite) primitives, like an infinite strip, or something that tiled the infinite plane. Kind of related to my original idea that `planar` could itself be a primitive of the infinite plane. Of course, that brings up the issue of how you handle U,V coordinates (clamp{X,Y}, tile{X,Y}, mirror{X,Y}, ...).

Perhaps instead we should implement quad directly as a primitive, hardcoded to have [0,1] UV bounds, and then derived objects could have alternate UV hit methods but all be constrained to the UV in [0,1]×[0,1].

I was toying with the idea of replacing the big large sphere "floor" with an actual infinite plane, but just using a large quad would be sufficient. If I ever did my own version, I'd also toss in UV transforms.

src/TheNextWeek/planar.h
hollasch: Yes. This one is an abstract base class, so just tends to the common case.

src/TheNextWeek/planar.h
hollasch: Yes, this would get a comment in the final version.

src/TheNextWeek/planar.h
hollasch: Well, it _is_ `(0,0)`. Certainly warrants a comment though. The basic idea is that we're defining a new 2D coordinate system (or frame). Given that, "origin" is the most accurate term, but I can improve the naming. Something like "plane_origin" or "origin_2D".

src/TheNextWeek/planar.h
hollasch: !

hollasch: If I didn't have `interval`, I'd do `return (0 <= u) && (u <= 1) && (0 <= v) && (v <= 1)`. I agree that `interval::unit` is an unfortunate stumbling block. Perhaps

```
auto legal = interval(0,1);
return legal.contains(u) && legal.contains(v);
```

Hmmmm. Perhaps a better notation exists. `return in_range(u, 0, 1) && in_range(v, 0, 1)`?

src/TheNextWeek/planar.h
trevordblack: Since `planar` is a purely abstract class (since `hit_uv = 0`), I would rather not solve for the bbox at all.
I don't see the utility in it.

If the reader (or you) want to implement infinite strips or what have you, then it should be as a child class

trevordblack: i.e.
```
 planar(const point3& o, const vec3& side_A, const vec3& side_B, shared_ptr<material> m)
      : origin(o), a(side_A), b(side_B), mat(m)
    {
        normal = unit_vector(cross(a, b));
        D = -dot(normal, origin);
        a /= dot(a,a);
        b /= dot(b,b);
}
```

src/TheNextWeek/planar.h
trevordblack: While I agree with you, if we're defining a new plane (which I agree with, but since there isn't associated text, that's not where my brain went with), then `vec3 a`, and `vec3 b` should actually probably be `vec3 x` and `vec3 y/z`.
But that would be maximally confusing.

maybe `point3 plane_origin`, `vec3 plane_x`, `vec3 plane_y`.

That depends if we want the normal of the plane to `y` or `z`. Which, I could genuinely be argued into either.

src/TheNextWeek/planar.h
trevordblack: Not sure the best way to do it, but `interval::unit.contains(u)` is not easy to parse

src/common/aabb.h
```diff
@@ -35,6 +35,27 @@ class aabb {
...
+/// is required for all 2D primitives, so it seems better to make such padding an option on the
+/// AABB class instead. This method is called whenever one or more dimensions might have zero
+/// thickness. In this case, I continue with a buried, hard-coded pad -- not production quality,
+/// but certainly appropriate for this project.
+
+    aabb pad() {
+        const double delta = 0.0001;
+        interval& new_x = (x.size() >= delta) ? x : x.expand(delta);
+        interval& new_y = (y.size() >= delta) ? y : y.expand(delta);
+        interval& new_z = (z.size() >= delta) ? z : z.expand(delta);
```
trevordblack, line 54: `interval&` is throwing an error in clang.

hollasch: Yeah, bizarre that it worked in MSVC.

src/TheNextWeek/planar.h
hollasch: Agreed. It's clunky compared to the alternatives.

hollasch:
Feeback incorporated. Now reworking for official quad primitive.

Issue #872, opened 2021-03-13
Signature mismatch: bvh_node constructor


MattTree:
The definition for the `bvh_node` constructor is missing the `const` keyword for the `src_objects` parameter.

I will submit a pull request.

PR #873, opened 2021-03-13
Signature mismatch: bvh_node constructor

MattTree:
Github issue: https://github.com/RayTracing/raytracing.github.io/issues/872


hollasch:
Looks good! How would you like to be listed in the acknowledgements? I can use your GitHub alias, or your full name (which I don't currently have).

MattTree:
Thank you. I would prefer not to be listed, but if it's required I would prefer my GitHub alias.

hollasch:
Totally optional. Thanks for contributing!

PR #897, opened 2021-06-08
Fix scaling in the final Perlin noise texture.

### LINKED ISSUE #896: Book 2 > Perlin Noise > Adjusting the phase: wrong scaling
In the current book (also in dev-major branch), the value for the noise texture looks like [this](https://github.com/RayTracing/raytracing.github.io/blob/dev-major/books/RayTracingTheNextWeek.html#L2235)
```c++
    return color(1,1,1) * 0.5 * (1 + sin(scale*p.z() + 10*noise.turb(p)));
```
Here the z coordinate of the `p` is scaled, but the `p` itself when it is passed to `noise.turb` is not. Therefore the turbulence is not scaled with the texture as it should be (one can also see it in the final image, where instead of a nice marble texture one sees some white noise). I think the correct line would be

 ```c++
    return color(1,1,1) * 0.5 * (1 + sin(scale*p.z() + 10*noise.turb(scale * p)));
```

With this change, one gets a nice marble texture in the final render, like here: https://github.com/mu-lambda/mu-lambda-raytracer/blob/main/final_scene.jpg

There is a related issue #425 : the picture would look even better if the noise was aligned along different axis: when one aligns along z, the unnatural con-axed circles on the marble surface are apparent.

### PR DESCRIPTION
**Pull request by mu-lambda**
Fixes #896.

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2232,7 +2232,7 @@ effect is:
 
         color value(double u, double v, const point3& p) const override {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            return color(1,1,1) * 0.5 * (1 + sin(scale*p.z() + 10*noise.turb(p)));
+            return color(1,1,1) * 0.5 * (1 + sin(scale*p.z() + 10*noise.turb(scale*p)));
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
         }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: books/RayTracingTheNextWeek.html [hunk 1]
```diff
@@ -2232,7 +2232,7 @@
 
         color value(double u, double v, const point3& p) const override {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            return color(1,1,1) * 0.5 * (1 + sin(scale*p.z() + 10*noise.turb(p)));
+            return color(1,1,1) * 0.5 * (1 + sin(scale*p.z() + 10*noise.turb(scale*p)));
```
hollasch, line 2235: Rather than scaling the Z component twice, let's create a vector to hold the scaled result, and then use that:
```
auto s = scale * p;
return color(1,1,1) * 0.5 * (1 + sin(s.z() + 10*noise.turb(s)));
```

This analogous change (defining a temporary scaled point) should be made in the earlier listing `noise-tex-2`.

hollasch (changes requested):
This looks like a good fix in the text. However, it's missing the corresponding change in the source. It also needs a description in the CHANGELOG.

Finally, can you create a new entry in `books/acknowledgements.md.html`?

Oh, and bonus points for directing this change to the correct branch! :D

books/RayTracingTheNextWeek.html
mu-lambda: Done

mu-lambda:
All done, ptal

hollasch (approved):
Thank you Dmitry!

PR #906, opened 2021-08-10
Fix book 2 scene ordering in main.cc

hollasch:


trevordblack (approved):
Wonderful.

Consider making an addition to the CHANGELIST

hollasch:
I believe this error was introduced earlier in our v4 work, so is really part of other items already in the changelog.

Issue #544, opened 2020-05-16
Next week: Instance rotation on y axis does not match the formula in hit function


D-K-E:
The formula for rotating about y axis is given as:
```
Similarly, for rotating about y (as we want to do for the blocks in the box) 
the formulas are:
x′=cos(θ)⋅x+sin(θ)⋅z
z′=−sin(θ)⋅x+cos(θ)⋅z
```

Its application by the `hit` function:
```c++
origin[0] = cos_theta*r.origin()[0] - sin_theta*r.origin()[2];
origin[2] = sin_theta*r.origin()[0] + cos_theta*r.origin()[2];
direction[0] = cos_theta*r.direction()[0] - sin_theta*r.direction()[2];
direction[2] = sin_theta*r.direction()[0] + cos_theta*r.direction()[2];
...
p[0] =  cos_theta*rec.p[0] + sin_theta*rec.p[2];
p[2] = -sin_theta*rec.p[0] + cos_theta*rec.p[2];
normal[0] =  cos_theta*rec.normal[0] + sin_theta*rec.normal[2];
normal[2] = -sin_theta*rec.normal[0] + cos_theta*rec.normal[2];
```

The `origin` and `direction` follows the formula for rotating about z axis:
```
The result for rotating counter-clockwise about z is:
x′=cos(θ)⋅x−sin(θ)⋅y
y′=sin(θ)⋅x+cos(θ)⋅y
```

trevordblack:
Simply changing `rotate_y` to:

```
origin[0] =  cos_theta*r.origin()[0] + sin_theta*r.origin()[2];
origin[2] = -sin_theta*r.origin()[0] + cos_theta*r.origin()[2];
direction[0] =  cos_theta*r.direction()[0] + sin_theta*r.direction()[2];
direction[2] = -sin_theta*r.direction()[0] + cos_theta*r.direction()[2];
...
p[0] =  cos_theta*rec.p[0] + sin_theta*rec.p[2];
p[2] = -sin_theta*rec.p[0] + cos_theta*rec.p[2];
normal[0] =  cos_theta*rec.normal[0] + sin_theta*rec.normal[2];
normal[2] = -sin_theta*rec.normal[0] + cos_theta*rec.normal[2];
```

Produces incorrect results. With the red diffuse showing up on the wrong side of the cube (likewise for green).
I don't remember exactly, but i'm pretty sure the source is correct as written.
[...]

PR #907, opened 2021-08-13
New 2D `quad` primitive

hollasch:
My ability to be self-critical is pretty worn down at this point, as I've lived in this world for almost nine months now. Please comment freely.


src/TheRestOfYourLife/pdf.h
```diff
@@ -13,6 +13,7 @@
 
 #include "rtweekend.h"
 
+#include "hittable_list.h"
```
trevordblack, line 16: This needs to be added to the text of the book.

Either spin up a new issue, or fix it in a new commit

src/TheRestOfYourLife/main.cc
```diff
@@ -64,48 +64,6 @@ void cornell_box(scene& scene_desc) {
     lights.add(make_shared<sphere>(point3(190, 90, 190), 90, shared_ptr<material>()));
 }
 
-void cornell_smoke(scene& scene_desc) {
```
trevordblack, line 67: GOOD CHANGE

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+====================================================================================================
+We've managed to get more than half way through this three-book series using spheres as our only
+geometric primitive. Time to add our second primitive: the quadrilateral.
+
+
+Ray-Plane Intersection
+-----------------------
+Spheres are generally the first ray tracing primitive because their implicit formula makes it so
+easy to solve for ray intersection. Like spheres, planes also have an implicit formula, and
+ray-plane intersection is even _easier_ to solve.
```
trevordblack, line 2269: Like spheres, planes also have an implicit formula, and we can use their implicit formula to produce an algorithm that solves for ray-plane intersection. Indeed, ray-plane intersection is even _easier_ to solve.

trevordblack: Something like that. It might be good to reinforce the knowledge that we use the implicit formula to create an algorithm for the intersection, since working from implicit formula to intersection algorithm is used all over the place in 3D math.

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+  $$ Ax + By + Cz + D = 0 $$
+
+That is, a plane is the set of all points $\mathbf{P} = (x,y,z)$ that satisfy the formula above. It
+makes things slightly easier to use the alternate formulation:
+
+  $$ Ax + By + Cz = D $$
+
+(We didn't flip the sign of D because it's just some constant that we'll figure out later.)
+
+Here's another way to think of this formula: given the plane perpendicular to the normal vector
```
trevordblack, line 2282: Here's a good way to think of this formula:

I don't know if the definition above counts as one way to think of the formula, rather it defines it, so the following is really the _first_ way of thinking about it.

This feels slightly semantic, but that struck out to me.

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+$\mathbf{n} = (A,B,C)$, and the position vector $\mathbf{v} = (x,y,z)$ (that is, the vector from the
+origin to any point on the plane), then
```
trevordblack, line 2284: Here's another way to think of this formula: given the plane perpendicular to the normal vector
$\mathbf{n} = (A,B,C)$, and the position vector $\mathbf{v} = (x,y,z)$ , then we can take the dot product to solve for $D$:

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+  $$ \mathbf{n} \cdot \mathbf{v} = D $$
+
+Now to find the intersection with some ray $\mathbf{R}(t) = \mathbf{P} + t\mathbf{d}$. Plugging in
```
trevordblack, line 2288: This is an equivalent formulation to the $ Ax + By + Cz = D $ formula given above, only now represents in terms that we have code for.

Now to find the intersection with some ray $\mathbf{R}(t) = \mathbf{P} + t\mathbf{d}$. Plugging in

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+  $$ \mathbf{n} \cdot ( \mathbf{P} + t \mathbf{d} ) = D $$
+
+Solving for $t$:
+
+  $$ \mathbf{n} \cdot \mathbf{P} + \mathbf{n} \cdot t \mathbf{d}  = D $$
+
+  $$ \mathbf{n} \cdot \mathbf{P} + t(\mathbf{n} \cdot \mathbf{d}) = D $$
+
+  $$ t = \frac{D - \mathbf{n} \cdot \mathbf{P}}{\mathbf{n} \cdot \mathbf{d}} $$
```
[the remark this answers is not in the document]
trevordblack: We also implicitly check for a positive t below with:

```
            // Return false if the hit point parameter t is outside the ray interval.
            auto t = (-D - dot(normal, r.origin())) / denom;
            if (!ray_t.contains(t))
                return false;
```

This should be called out in the text, and also probably in the comments.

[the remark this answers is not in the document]
trevordblack: It might be worth mentioning something like,

3 possibilities exist:
1. No intersection, zero in denominator
2. A solution where the t value is negative, so no intersection
3. A solution where the t value is positive, so an intersection exists

It's only on case three where we want to continue searching for the actual point of intersection

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+3. How do we orient the ray-plane intersection point within the quadrilateral?
+
+
+Defining the Quadrilateral
+---------------------------
+For our purposes, we'll use three geometric entities to define a quadrilateral. Remember that when
+we say "quadrilateral" here, we're really talking about parallelograms, a specific kind of
+quadrilateral where opposite sides are parallel. These three entities are
+
+1. $\mathbf{O}$, the lower-left corner of the quadrilateral.
```
trevordblack, line 2321: (not to be confused with the origin) ?

Oooh yuck. How do you want to deal with the global origin versus quad origin question?
This buttresses up against https://github.com/RayTracing/raytracing.github.io/issues/544

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+2. $\mathbf{u}$, a vector representing the first side of the quadrilateral.
+   $\mathbf{O} + \mathbf{u}$ gives the lower right corner.
+3. $\mathbf{v}$, a vector representing the second side of the quadrilateral.
+   $\mathbf{O} + \mathbf{v}$ gives the upper left corner.
+
+These values are three-dimensional, even though a quad itself is a two-dimensional object.
```
trevordblack, line 2327: Give examples for all of O, u, v (possibly two different sets)

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+The following figure illustrates the quadrilateral components.
+
+  ![Figure [quad-def]: Quadrilateral Components](../images/fig-2.05-quad-def.jpg)
```
trevordblack, line 2331: I really like this picture

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+
+Finding the Plane That Contains a Given Quadrilateral
+------------------------------------------------------
+We have $O$, $u$, and $v$, and want the corresponding equation of the plane containing all three of
```
trevordblack, line 2336: \mathbf{O}, ...

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+these values (and thus the quad itself).
+
+Fortunately, this is very simple. Recall that in $Ax + By + Cz = D$, $(A,B,C)$ represents the
+normal vector. To get this, we just use the cross product of the two side vectors $u$ and $v$:
+
+  $$ \mathbf{n} = \text{unit_vector}(\mathbf{u} \times \mathbf{v}) $$
+
+The plane is defined as all points $(x,y,z)$ that satisfy the equation $Ax + By + Cz = D$. Well, we
```
trevordblack, line 2344: In general, the implicit formulation of a quadrilateral does not need to be normalized such that the normal vector has a length of one, but we'll want our normal vector to be unit length so we get the unit vector of the cross product here.

That's kind of bad english, but maybe explain (briefly) the addition of the `unit_vector`

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+-----------------------
+Spheres are generally the first ray tracing primitive because their implicit formula makes it so
+easy to solve for ray intersection. Like spheres, planes also have an implicit formula, and
+ray-plane intersection is even _easier_ to solve.
+
+You may already know this implicit formula for a plane:
+
+  $$ Ax + By + Cz + D = 0 $$
+
+That is, a plane is the set of all points $\mathbf{P} = (x,y,z)$ that satisfy the formula above. It
```
trevordblack, line 2275: Where A,B,C,D are just numbers, and x,y,z are the values of vector, That is, a plane is the set of all points $\mathbf{P} = (x,y,z)$ that satisfy the formula above.

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+Fortunately, this is very simple. Recall that in $Ax + By + Cz = D$, $(A,B,C)$ represents the
+normal vector. To get this, we just use the cross product of the two side vectors $u$ and $v$:
+
+  $$ \mathbf{n} = \text{unit_vector}(\mathbf{u} \times \mathbf{v}) $$
+
+The plane is defined as all points $(x,y,z)$ that satisfy the equation $Ax + By + Cz = D$. Well, we
+know that $\mathbf{O}$ lies on the plane, so we can use that to find $D$:
+
+  $$ D = \mathbf{n}_x\mathbf{O}_x + \mathbf{n}_y\mathbf{O}_y + \mathbf{n}_z\mathbf{O}_z $$
```
trevordblack, line 2347: $$ D = n \cdot O $$
  $$ D = \mathbf{n}_x\mathbf{O}_x + \mathbf{n}_y\mathbf{O}_y + \mathbf{n}_z\mathbf{O}_z $$

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+We can now use these two values $\mathbf{n}$ and $D$ to find the point of intersection with a given
+ray and the plane containing the quadrilateral.
+
+
+Orienting Points on The Plane
+------------------------------
+At this stage, the intersection point is on the plane that contains the quadrilateral, but it could
+be _anywhere_ on the plane, and may lie inside or outside the quadrilateral. We need to determine if
```
trevordblack, line 2356: be _anywhere_ on the plane, and since the quadrilateral may be a small little section of the plane, the intersection may lie _inside_ or _outside_ the quadrilateral.

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+the intersection lies inside (hits) the quadrilateral, and reject points that lie outside (miss) the
+quad. To determine where a point lies relative to the quad, and to assign texture coordinates to the
+point of intersection, we need to somehow orient the intersection point on the plane.
+
+To do this, we'll construct a _coordinate frame_ for the plane -- a way of orienting points located
+on the plane. We've already been using a coordinate frame for our 3D space -- this is defined by an
+origin point $\mathbf{O}$ and three basis vectors $\mathbf{x}$, $\mathbf{y}$, and $\mathbf{z}$.
```
trevordblack, line 2363: and three basis vectors $\mathbf{x}$ ($(1,0,0)$), $\mathbf{y}$ ($(0,1,0)$), and $\mathbf{z}$ ($(0,0,1)$).

trevordblack: These could be confused with the x,y,z from the implicit formula, so best to make them explicit.

books/RayTracingTheNextWeek.html
trevordblack: I see you mention coordinate frame below, that may be enough

books/RayTracingTheNextWeek.html
trevordblack: Oh, wait. The origin point in this sentence is (0,0,0) and is not necessarily the same origin as O above. This needs to be made explicit.

Both the coordinate frame and the global frame will have a distinct origin (and the coordinate frame's origin may not overlap with the global frame's origin if the coordinate frame's orgin is projected into the global frame).

Uuuuhhh, maybe just skip the second part.

trevordblack: \bfmath{O_p} and \bfmath{O_g} for plane and global? I don't the nonstandard notation, but it feels like it might be worth the distinction?

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+Since a plane is a 2D construct, we just need a plane origin point $\mathbf{O}$ and _two_ basis
+vectors: $\mathbf{u}$ and $\mathbf{v}$. Now normally, you see axes that are perpendicular to each
+other, so they all meet at 90° angles. However, this doesn't need to be the case in order to span
+the entire space -- you just can't have two axes that are parallel to each other.
```
trevordblack, line 2368: However, this doesn't need to be the case in order to span the entire space -- you just need two axes that are not parallel to each other.

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+Consider figure [ray-plane] as an example. Ray $\mathbf{R}$ intersects the plane, yielding
+intersection point $\mathbf{P}$. Measuring against plane vectors $\mathbf{u}$ and $\mathbf{v}$, the
+intersection point $\mathbf{P}$ is at $\mathbf{O} + (1)\mathbf{u} + (\frac{1}{2})\mathbf{v}$. In
+other words, the plane coordinates of intersection point $\mathbf{P}$ is $(1,\frac{1}{2})$.
+
+Generally, given some arbitrary point $\mathbf{P}$, we seek two scalar values $\alpha$ and $\beta$,
+so that
+
+  $$ \mathbf{P} = \mathbf{O} + \alpha \mathbf{u} + \beta \mathbf{v} $$
```
trevordblack, line 2380: $$ \mathbf{P} = \mathbf{O} + \alpha \mathbf{u} + \beta \mathbf{v} = (\alpha, \beta} $$

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+
+If $\mathbf{u}$ and $\mathbf{v}$ were guaranteed to be orthogonal to each other (forming a 90°
+angle between them), then this would be a simple matter of projecting $\mathbf{P}$ onto the two
+basis vectors $\mathbf{u}$ and $\mathbf{v}$. However, since they can form any angle (other than
```
trevordblack, line 2384: However, since we are not restricting u and v to be perpendicular, and allowing them to form any angle ...

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+   &= \alpha(\mathbf{v} \times \mathbf{u}) + \beta(\mathbf{v} \times \mathbf{v})
+   \end{align*} $$
+
+Since any vector crossed with itself yields zero, these equations simplify to
+
+$$ \mathbf{u} \times \mathbf{p} = \beta(\mathbf{u} \times \mathbf{v}) $$
+$$ \mathbf{v} \times \mathbf{p} = \alpha(\mathbf{v} \times \mathbf{u}) $$
+
+To solve for the coefficients $\alpha$ and $\beta$, we can take the dot product of both sides of the
+above equations with the plane normal $\mathbf{n} = \mathbf{u} \times \mathbf{v}$, reducing both
```
trevordblack, line 2414: Are we restricting u and v to unit length? Because we normalized the normal vector above, so this equation isn't strictly true unless the cross and u and v just happens to be unit length.

trevordblack:
I was getting an error in clang:

```
In file included from /mnt/c/Users/Trevor/raytracing.github.io/src/TheRestOfYourLife/main.cc:19:
/mnt/c/Users/Trevor/raytracing.github.io/src/TheRestOfYourLife/quad.h:94:18: error: use of undeclared identifier
      'plane_origin'
        auto p = plane_origin + (random_double() * axis_A) + (random_double() * axis_B);
                 ^
/mnt/c/Users/Trevor/raytracing.github.io/src/TheRestOfYourLife/quad.h:94:52: error: use of undeclared identifier
      'axis_A'
        auto p = plane_origin + (random_double() * axis_A) + (random_double() * axis_B);
                                                   ^
/mnt/c/Users/Trevor/raytracing.github.io/src/TheRestOfYourLife/quad.h:94:81: error: use of undeclared identifier
      'axis_B'
        auto p = plane_origin + (random_double() * axis_A) + (random_double() * axis_B);
                                                                                ^
```

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+            rec.p = intersection;
+            rec.mat = mat;
+            rec.set_face_normal(r, normal);
+
+            return true;
+        }
+
+      protected:
+        point3 corner;
+        vec3 side_A, side_B;
```
trevordblack, line 2543: decided against using u and v?

trevordblack: and then using corner instead of origin

Which feels alright to me. The problem is that linear algebra 100% doesn't care what frame something is in and can be defined against any

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+More Planar Primitives
+-----------------------
+You may have noticed that if you have the 2D planar coordinates of a hit point, you can implement
+arbitrary 2D shapes -- not just parallelograms. In fact, you'd only need to change the `hit_ab`
+method above to determine if the (a,b) coordinates lie inside your 2D primitive, and set the texture
+UV coordinates if so.
+
+We'll leave these additional 2D primitives as an exercise to the reader, depending on your desire to
+explore. Consider triangles, disks, and rings. You could even create cut-out stencils based on the
+texture map! We've included the solution for triangle, ellipse, and annulus in the commit history
```
trevordblack, line 2566: We've included the solution for triangle, ellipse, and annulus ...

Consider cutting. The book != the commit history

books/RayTracingTheNextWeek.html
```diff
@@ -2256,7 +2256,319 @@
...
+            // Return false if the hit point parameter t is outside the ray interval.
+            auto t = (-D - dot(normal, r.origin())) / denom;
+            if (!ray_t.contains(t))
+                return false;
+
+            // Determine the hit point lies within the planar shape using its plane coordinates.
+            auto intersection = r.at(t);
+            vec3 planar_hitpt_vector = intersection - corner;
+            auto a = dot(w, cross(planar_hitpt_vector, side_B));
+            auto b = dot(w, cross(side_A, planar_hitpt_vector));
```
trevordblack, line 2527: auto alpha ..
auto beta...

books/RayTracingTheNextWeek.html
```diff
@@ -2714,56 +2850,54 @@
 make an axis-aligned block primitive that holds 6 rectangles:
```
trevordblack, line 2850: The Cornell Box usually has two blocks in it. These are rotated relative to the walls. First, let’s create a function that returns a box, by creating a `hittable_list` that holds 6 rectangles:

trevordblack:
I haven't dug through the code deeply enough to look for problems, but I'm really happy with what I see so far

books/RayTracingTheNextWeek.html
hollasch: As a last-minute change in the code, I renamed this from `origin` to `corner`, for the same reasons. I'm not completely convinced that it's a good term, but at least it addresses the origin-origin confusion. However, you're correct that I didn't review the text for the same issue.

books/RayTracingTheNextWeek.html
hollasch: (Maybe I should rename these (C, u, v) for corner.)

books/RayTracingTheNextWeek.html
hollasch: Ok. I'll leave it as an Easter egg.

hollasch:
These comments are gold -- _thank you_ for reviewing so diligently. Will start playing with them tonight.

books/RayTracingTheNextWeek.html
hollasch: Done.

books/RayTracingTheNextWeek.html
hollasch: Reworked.

books/RayTracingTheNextWeek.html
hollasch: Done.

books/RayTracingTheNextWeek.html
hollasch: Done.

books/RayTracingTheNextWeek.html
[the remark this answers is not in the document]
hollasch: Good point calling out the zero denominator. I don't think we need to comment about negative _t_ values. For one, it's actually _t_ values _less_ than some given value (not necessarily zero), and two, this is already consistent with our hit parameter handling for all other values.

Ah, what the heck, I'll add a sentence.

books/RayTracingTheNextWeek.html
hollasch: Fixed.

books/RayTracingTheNextWeek.html
hollasch: oopsie

hollasch: Better.

books/RayTracingTheNextWeek.html
hollasch: Hmmm, I like the idea here, but am not really comfortable relying on `(` / `)` to indicate coordinate in a different space. Let me think about this.

books/RayTracingTheNextWeek.html
hollasch: Nice.

books/RayTracingTheNextWeek.html
hollasch: Reworded.

books/RayTracingTheNextWeek.html
hollasch: Cut.

hollasch:
First round of changes. More later.

books/RayTracingTheNextWeek.html
hollasch: Renamed in the book from O to Q. Also updated member variable names to Q, u, v.

books/RayTracingTheNextWeek.html
hollasch: Done.

books/RayTracingTheNextWeek.html
hollasch: Actually, I think it reads better in the opposite order, so I did that.

books/RayTracingTheNextWeek.html
hollasch: Done.

books/RayTracingTheNextWeek.html
hollasch: I think this is more clear now with the rename off the planar basis frame to use *Q* as the planar frame origin (quad corner point).

books/RayTracingTheNextWeek.html
hollasch: I think I'll skip this suggestion; I think it confuses things more.

books/RayTracingTheNextWeek.html
hollasch: Fixed (?) in latest rename to Q, u, v.

books/RayTracingTheNextWeek.html
hollasch: Done.

books/RayTracingTheNextWeek.html
hollasch: No, *u* and *v* are not restricted to unit length, and the normal vector *n* is merely *u* ⨯ *v*. No normalization needed; we're just cancelling terms.

hollasch: Of course, correct me if this is not wrong, but I do remember verifying on paper with non-unit vectors.

hollasch:
I think that's it.

books/RayTracingTheRestOfYourLife.html
```diff
@@ -2377,47 +2381,6 @@
     [Listing [emitted-directional]: <kbd>[material.h]</kbd> Material emission, directional]
 </div>
 
-<div class='together'>
-We also need to flip the light so its normals point in the $-y$ direction:
-
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
```
trevordblack, line 2383: Why was this section removed? Are lights now double sided?

src/TheRestOfYourLife/quad.h
```diff
@@ -26,6 +26,8 @@ class quad : public hittable {
         D = -dot(normal, corner);
         w = n / dot(n,n);
 
+        area = n.length();
+
```
trevordblack, line 30: I don't see these changes in the text?

Same with area below.

books/RayTracingTheNextWeek.html
```diff
@@ -2273,6 +2273,7 @@
 
   $$ Ax + By + Cz + D = 0 $$
 
+where $A,B,C,D$ are just numbers, and $x,y,z$ are the values of a point $\mathbf{P} = (x,y,z)$.
```
trevordblack, line 2276: Sorry. This was my bad. I think maybe we should change to

where $A,B,C,D$ are just constants, and $x,y,z$ are the values of a point $\mathbf{P} = (x,y,z)$.

trevordblack:
```
quadrilateral. In fact, we can use this approach to define any planar primitive, like triangles and disks (more later). 
```
to
```
quadrilateral. In fact, we can use this approach to define any planar primitive, like triangles and disks (more on that later). 
```

books/RayTracingTheNextWeek.html
trevordblack: u and v also need bolding

books/RayTracingTheNextWeek.html
```diff
@@ -2326,13 +2327,15 @@
...
+  1. $\mathbf{Q}$, the lower-left corner of the quadrilateral.
+  2. $\mathbf{u}$, a vector representing the first side of the quadrilateral.
+     $\mathbf{Q} + \mathbf{u}$ gives the lower right corner.
+  3. $\mathbf{v}$, a vector representing the second side of the quadrilateral.
+     $\mathbf{Q} + \mathbf{v}$ gives the upper left corner.
 
-These values are three-dimensional, even though a quad itself is a two-dimensional object.
+These values are three-dimensional, even though a quad itself is a two-dimensional object. For
+example, a quad with corner at the origin and extending two units in the Z direction and one unit in
+the Y direction would have $Q = (0,0,0), u = (0,0,2), \text{and } v = (0,1,0)$.
```
trevordblack, line 2338: bold Q, u, and v

trevordblack: also bold the u and v in
```
To get this, we just use the cross product of the two side vectors u and v: 
```

trevordblack:
Why are we taking the unit_vector of u cross v ?

books/RayTracingTheNextWeek.html
```diff
@@ -2350,9 +2350,9 @@
   $$ \mathbf{n} = \text{unit_vector}(\mathbf{u} \times \mathbf{v}) $$
 
 The plane is defined as all points $(x,y,z)$ that satisfy the equation $Ax + By + Cz = D$. Well, we
-know that $\mathbf{O}$ lies on the plane, so we can use that to find $D$:
+know that $\mathbf{Q}$ lies on the plane, so we can use that to find $D$:
 
-  $$ D = \mathbf{n}_x\mathbf{O}_x + \mathbf{n}_y\mathbf{O}_y + \mathbf{n}_z\mathbf{O}_z $$
+  $$ D = \mathbf{n}_x\mathbf{Q}_x + \mathbf{n}_y\mathbf{Q}_y + \mathbf{n}_z\mathbf{Q}_z $$
```
trevordblack: What's our policy on bolding vector components? Do we have one?

trevordblack:
Sorry, leaving comments now. It's really annoying tracking individual changes to specific commits

n \cdot \mathbf{Q}

n needs to be bold

trevordblack:
Hmm. Consider:

```
Since any vector crossed with itself yields zero, these equations simplify to

  $$ \mathbf{u} \times \mathbf{p} = \beta(\mathbf{u} \times \mathbf{v}) $$
  $$ \mathbf{v} \times \mathbf{p} = \alpha(\mathbf{v} \times \mathbf{u}) $$

To solve for the coefficients $\alpha$ and $\beta$, we can take the dot product of both sides of the
above equations with the plane normal $\mathbf{n} = \mathbf{u} \times \mathbf{v}$, reducing both
sides to scalars.
```

My brain immediately goes to try and divide out the u cross v and the v cross u.

We might want to stop a spend a sentence to say that you can't do that because you can't divide by a vector, and that we use a dot product to produce a scalar, something that we _can_ divide by.

trevordblack:
I don't believe that the source for `aabb::pad` is anywhere in the text. Needs to be outlined before it's use in `quad`

books/RayTracingTheNextWeek.html
trevordblack: I think that including this would actually be a _bad_ change. And you were right not to include it

books/RayTracingTheRestOfYourLife.html
hollasch: We had to flip XZ rects, because they always point up. With quads, you can specify them in either orientation.

trevordblack: Should we maybe call out the fact that quad lights are one sided?

src/TheRestOfYourLife/quad.h
hollasch: Wow, good catch.

books/RayTracingTheNextWeek.html
hollasch: Oooh, excellent question. Despite the above, I'd say points and vectors are bolded, scalars are not. And "Qz" is a scalar, so should be unbolded. Will fix.

hollasch: Sigh. I suppose I should give a darn and review the entire text...

hollasch: Ok, that wasn't too hard. Turns out that we're consistent everywhere else. Excellent catch.

hollasch:
Actually, I prefer the annoying individual changes comments; makes things easier to track.

hollasch:
I like your explanation about dividing by a vector. Will do.

books/RayTracingTheRestOfYourLife.html
trevordblack: Ehhh, no need. We already discuss sidedness in chapter 6 of book 1.

books/RayTracingTheNextWeek.html
hollasch: Done.

PR #913, opened 2021-08-21
Added world space and object space explanation

### LINKED ISSUE #544: Next week: Instance rotation on y axis does not match the formula in hit function

trevordblack:
Resolves #544

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2975,14 +2975,14 @@ This gives:
 
 <div class='together'>
 Now that we have boxes, we need to rotate them a bit to have them match the _real_ Cornell box. In
-ray tracing, this is usually done with an _instance_. An instance is a copy of a  geometric
+ray tracing, this is usually done with an _instance_. An instance is a copy of a geometric
 primitive that has been placed into the scene. This instance is entirely independent of the other
-copies of the primitive and can be moved or rotated around. In this case our geometric primitive is
-our box hittable, and we want to rotate it somehow. This is especially easy in ray tracing because
-we don’t move anything; instead we move the rays in the opposite direction. For example, consider a
-_translation_ (often called a _move_). We could take the pink box at the origin and add 2 to all its
-x components, or (as we almost always do in ray tracing) leave the box where it is, but in its hit
-routine subtract 2 off the x-component of the ray origin.
+copies of the primitive and can be moved or rotated. In this case, our geometric primitive is our
+hittable `box` object, and we want to rotate it. This is especially easy in ray tracing because we
+don’t actually need to move objects in the scene; instead we move the rays in the opposite
+direction. For example, consider a _translation_ (often called a _move_). We could take the pink box
+at the origin and add two to all its x components, or (as we almost always do in ray tracing) leave
+the box where it is, but in its hit routine subtract two off the x-component of the ray origin.
 
   ![Figure [ray-box]: Ray-box intersection with moved ray vs box](../images/fig-2.08-ray-box.jpg)
 
@@ -2996,11 +2996,10 @@ Whether you think of this as a move or a change of coordinates is up to you. The
 this is to think of moving the incident ray backwards the offset amount, determining if an
 intersection occurs, and then moving that intersection point forward the offset amount.
 
-If we have an intersection along the offset ray, then we know that the incident ray will hit the
-translated object. However, our offset ray is not the original ray, so if we just take the
-intersection we found along the offset ray then our intersection will actually be wrong by the
-offset amount, so we have to shift the intersection point back onto the original ray by the offset
-amount. Let's add the code to make this happen.
+We need to move the intersection point forward the offset amount so that the intersection is
+actually in the path of the incident ray. If we forgot to move the intersection point forward then
+the intersection would be in the path of the offset ray, which isn't correct. Let's add the code to
+make this happen.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class translate : public hittable {
@@ -3067,9 +3066,9 @@ amount. Let's add the code to make this happen.
     [Listing [translate-class]: <kbd>[hittable.h]</kbd> Hittable translation class]
 </div>
 
-We need to remember to offset the bounding box, otherwise the incident ray might be looking in the
-wrong place and trivially reject the intersection. The expression `object->bounding_box() + offset`
-above requires some additional support.
+We also need to remember to offset the bounding box, otherwise the incident ray might be looking in
+the wrong place and trivially reject the intersection. The expression
+`object->bounding_box() + offset` above requires some additional support.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class aabb {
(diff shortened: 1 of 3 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -2904,24 +2951,112 @@
...
 
   $$ y' = \cos(\theta) \cdot y - \sin(\theta) \cdot z $$
   $$ z' = \sin(\theta) \cdot y + \cos(\theta) \cdot z $$
 </div>
 
-Unlike the situation with translations, the surface normal vector also changes, so we need to
-transform directions too if we get a hit. Fortunately for rotations, the same formulas apply. If you
-add scales, things get more complicated. See the web page https://in1weekend.blogspot.com/ for links
-to that.
+For a simple operation like a translation, thinking of the operation as a simple movement of the
```
hollasch, line 3155: "Thinking of translation as a simple movement …"

books/RayTracingTheNextWeek.html
```diff
@@ -2904,24 +2951,112 @@
...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+    bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
+        // Move the ray backwards by the offset
+
+        // Determine where (if any) an intersection occurs along the offset ray
+
+        // Move the intersection point forwards by the offset
+    }
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [translate-hit-move]: <kbd>[hittable.h]</kbd> The moving translate hit code]
```
hollasch, line 3170: Let's replace this and the code/pseudo-code below with simple numbered lists.

books/RayTracingTheNextWeek.html
```diff
@@ -2904,24 +2951,112 @@
...
+
+        // Change the intersection point from object space to world space
+    }
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [translate-hit-coord]: <kbd>[hittable.h]</kbd> The change of coordinates translate hit code]
 
 <div class='together'>
+Rotating an object will not only change the point of intersection, but will also change the surface
+normal vector, which will change the direction of reflections and refractions. So we need to change the
+normal as well. Fortuneatly, we can just treat the normal like a vector, and so use the same
```
hollasch, line 3188: `*` Fortunately

hollasch: "we can just treat the normal like a vector" — this is a dangerous statement to make, as it's really a co-vector, and as such responds differently to scaling and shearing transforms. In fact, like my statement above about the common mistaken idea of rotations being about an axis, this can lead to confounding and confusing render bugs that in most cases will never be caught.

I'd be more comfortable saying that normal vectors respond to rotations in the same way that all vectors do, "so use the same formulas as above".

I don't think it makes sense to really get into spatial transforms here, and stuff that works only in 3D is fine, but I also want to be correct. We will be forced to address this later when it comes to scaling transforms. (Now I'm wondering if we implemented normal transforms correctly. I'm on the edge of my seat.)

books/RayTracingTheNextWeek.html
```diff
@@ -2904,24 +2951,112 @@
...
+            auto origin = r.origin();
+            auto direction = r.direction();
+
+            origin[0] = cos_theta*r.origin()[0] - sin_theta*r.origin()[2];
+            origin[2] = sin_theta*r.origin()[0] + cos_theta*r.origin()[2];
+
+            direction[0] = cos_theta*r.direction()[0] - sin_theta*r.direction()[2];
+            direction[2] = sin_theta*r.direction()[0] + cos_theta*r.direction()[2];
+
+            ray rotated_r(origin, direction, r.time());
```
hollasch, line 3209: Getting more convinced than ever that we should strongly think about implementing ray and vector transforms. This function is littered with minutia about vector, point and ray transforms that fuzzes the higher-level operations happening here. I believe this would make the codebase more clear and more compact.

books/RayTracingTheNextWeek.html
```diff
@@ -2904,24 +2951,112 @@
...
+            return true;
+        }
+
+        ...
+    };
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [rot-y-hit]: <kbd>[hittable.h]</kbd> Hittable rotate-Y hit function]
+
+</div>
+The code for solving from world space to object space might at first appear incorrect because we are
```
hollasch, line 3237: I think a better strategy would be to flip θ up front, with an explanatory comment, and then proceed with a normal XZ rotation. This seems better than baking the inversion into the various equations, and then explaining why the equations seem off.

books/RayTracingTheNextWeek.html
hollasch: Ok, came back after looking at our code, which apparently just has XZ rotation and translate. We've left a booby trap for the reader who in the future wants to implement scaling transforms, and then has to deal with the transform/inverse-transform of normal (co-)vectors. Ugh, I don't really know what to do about this. I guess we can just wash our hands of this for now, or leave a cryptic comment about scales and general transforms requiring a different strategy, or an appendix article in the future, or deal with it in a subsequent work that really dives into transforms. This is pretty unsatisfying, but I think I'll just have to live with the boundary here.

books/RayTracingTheNextWeek.html
hollasch: BTW, with our point/vector duality inherent in the `vec3` class, we'd need

- translate_as_point()
- translate_as_vector() (or just omit this entirely as it's a no-op)
- scale()
- scale_as_normal()
- rotate()

And I'd almost rather have a `rotate(const vec3& axis, double degrees)` method than `rotate_x()`, `rotate_y()`, `rotate_z()` methods, but `¯\_(ツ)_/¯`.

hollasch: Thankfully, the `ray` class would just have `translate()`, `rotate()` and `scale()` methods.

src/TheNextWeek/hittable.h
```diff
@@ -55,12 +55,16 @@ class translate : public hittable {
     }
 
     bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
-        ray moved_r(r.origin() - offset, r.direction(), r.time());
-        if (!object->hit(moved_r, ray_t, rec))
+        // Move the ray backwards by the offset
```
hollasch, line 58: Applause for the commentary. Very useful here.

hollasch (changes requested):

books/RayTracingTheNextWeek.html
hollasch: How about we just resolve this for now with "Fortunately, for rotation transforms we can just treat the normal like a vector, and so use …"

books/RayTracingTheNextWeek.html
hollasch: Nothing actionable in this PR.

books/RayTracingTheNextWeek.html
trevordblack: I cannot believe this didn't occur to me.

books/RayTracingTheNextWeek.html
```diff
@@ -2778,12 +2778,14 @@
 
 <div class='together'>
 Now that we have boxes, we need to rotate them a bit to have them match the _real_ Cornell box. In
-ray tracing, this is usually done with an _instance_. An instance is a geometric primitive that has
-been moved or rotated somehow. This is especially easy in ray tracing because we don’t move
-anything; instead we move the rays in the opposite direction. For example, consider a _translation_
-(often called a _move_). We could take the pink box at the origin and add 2 to all its x components,
-or (as we almost always do in ray tracing) leave the box where it is, but in its hit routine
-subtract 2 off the x-component of the ray origin.
+ray tracing, this is usually done with an _instance_. An instance is a copy of a  geometric
```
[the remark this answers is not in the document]
trevordblack: Done

books/RayTracingTheNextWeek.html
```diff
@@ -2778,12 +2778,14 @@
...
+primitive that has been placed into the scene. This instance is entirely independent of the other
+copies of the primitive and can be moved or rotated around. In this case our geometric primitive is
```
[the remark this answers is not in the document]
trevordblack: Done

books/RayTracingTheNextWeek.html
```diff
@@ -2778,12 +2778,14 @@
...
+our box hittable, and we want to rotate it somehow. This is especially easy in ray tracing because
+we don’t move anything; instead we move the rays in the opposite direction. For example, consider a
```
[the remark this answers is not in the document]
trevordblack: Done.

books/RayTracingTheNextWeek.html
```diff
@@ -2778,12 +2778,14 @@
...
+_translation_ (often called a _move_). We could take the pink box at the origin and add 2 to all its
+x components, or (as we almost always do in ray tracing) leave the box where it is, but in its hit
+routine subtract 2 off the x-component of the ray origin.
```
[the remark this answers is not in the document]
trevordblack: substituted 2.0 instead of 2 or two

books/RayTracingTheNextWeek.html
```diff
@@ -2793,41 +2795,86 @@
...
-move any underlying hittable is a _translate_ instance.
+Whether you think of this as a move or a change of coordinates is up to you. The way to reason about
+this is to think of moving the incident ray backwards the offset amount, determining if an
+intersection occurs, and then moving that intersection point forward the offset amount.
+
+If we have an intersection along the offset ray, then we know that the incident ray will hit the
+translated object. However, our offset ray is not the original ray, so if we just take the
+intersection we found along the offset ray then our intersection will actually be wrong by the
+offset amount, so we have to shift the intersection point back onto the original ray by the offset
+amount. Let's add the code to make this happen.
```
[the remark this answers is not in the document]
trevordblack: The two paragraphs have two subtly different functions. I'll rewrite for clarity.

books/RayTracingTheNextWeek.html
```diff
@@ -2793,41 +2795,86 @@
...
         vec3 offset;
         aabb bbox;
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     };
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     [Listing [translate-class]: <kbd>[hittable.h]</kbd> Hittable translation class]
 </div>
 
-The expression `object->bounding_box() + offset` above requires some additional support.
+We need to remember to offset the bounding box, otherwise the incident ray might be looking in the
```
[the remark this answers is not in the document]
trevordblack: Done

books/RayTracingTheNextWeek.html
trevordblack: Done

trevordblack:
Which do you prefer of these two:
```
2. Determine where (if any) an intersection occurs along the offset ray
```
OR
```
2. Determine where (if anywhere) an intersection occurs along the offset ray
```

books/RayTracingTheNextWeek.html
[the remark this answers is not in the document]
trevordblack: Went with your suggestion on later review.

books/RayTracingTheNextWeek.html
trevordblack: Done

trevordblack:
Please check through edits made

D-K-E:
I would say the first one is better, because the phrasing fits more correctly to the output of the `hit` function. `hit` does not output a location (which would be a point in space), it outputs a boolean (a yes/no answer) which fits better to `if any intersection occurs` in my opinion.  

How about something like: `Determine (if any) an intersection occurs somewhere along the offset ray` ?

trevordblack:
Done

hollasch:
My computer is mostly alive again. Sorry for the lag.

hollasch (approved):

Issue #839, opened 2020-12-19
Math cleanup pass


hollasch:
- Use consistent indentation levels for our math blocks
- Use equation alignment where appropriate

ymherklotz:
In case this is useful, I just noticed that the alignment of italic variables in math blocks didn't use `\textit`, which means the spacing looks a bit strange sometimes as they are basically independent letters.

I have made the changes for all variables like that in Book3 (https://github.com/ymherklotz/raytracing.github.io/tree/textit-in-mathblocks), and if this is useful and a wanted changed, I could make the same changes in Book1 and Book2.

hollasch:
Thank you! In most cases, this has only a very slight effect on kerning, but there are definitely cases where it's significant. Here's a good example:

Before:
![pSurface-before](https://user-images.githubusercontent.com/313871/131231061-38b38e6a-7fa3-428e-87f5-1fb7a6028e42.jpg)

After:
![pSurface-after](https://user-images.githubusercontent.com/313871/131231072-a0ea459f-8b64-4af0-a091-b7b1c9246db3.jpg)

It's also proper markup, which would be enough to convince me.

Thanks again for the help -- converting the other two books would be most appreciated, based on the `dev-major` branch. Also, please add yourself to `acknowledgements.html`, and add a new v4.0.0 entry in changelog.md.

hollasch:
Done.

PR #915, opened 2021-08-29
Add \mathit for multiletter math variables

ymherklotz:
Based on the discussion in #839, I have added `\mathit` to all the multi-letter variable names in math blocks in the three books.  This is to fix some slight kerning issues when type-setting the maths.

- I used `\mathit` instead of `\textit` because it seems to give similar results, but should be more correct for math blocks.
- Book2 did not actually have any multi-letter variable names, so no changes were necessary.

In addition to adding `\mathit` around those variable names, I also formatted the paragraphs to fix the length of the lines so that it was consistent with the rest of the text.


hollasch (approved):
Thank you!

PR #916, opened 2021-08-29
Use mathbf{} instead of textbf{}

hollasch:
This notation crept in at some locations. Generally, mathbf{} is for
bold math text, and textbf{} is just general bold LaTeX text.


trevordblack:
Smart change

trevordblack (approved):

PR #918, opened 2021-09-01
Modifying Figure 1.13 for clarity

### LINKED ISSUE #542: Figure notation update: fig-1.13-refraction - refracted ray geometry
- Text talks about rays, but really is only concerned about vector components.  Should we resolve this?
- Vectors/rays should clearly indicate direction
- Text talks about normals **n** and **n'**. These are missing from the figure

`fig.ray-refract.jpg`

trevordblack:
Resolves #542


trevordblack:
Made the explicit decision to not annotate normals.

As I thought the figure did a decent enough job as is of showing the normals, and that would have made the figure excessively busy.

hollasch (approved):

PR #925, opened 2021-09-07
Removed add/sub of small floats to/from large ones

ahcox:
A vector was added to a point and then the point was immediately subtracted to recover the original vector, but with some possible precision loss as the vector can be very small.


trevordblack:
The loss of precision doesn't bother me so much as subtracting with a point.

Good catch. Will need to go through the code for arithmetic on pointers

hollasch:
Excellent change; thanks.

@trevordblack — The point/vector math looks correct to me both ways; `random_in_hemisphere()` returns a vector.

PR #984, opened 2022-01-30
Least mental model breaking way to add a color scattered variable

### LINKED ISSUE #981: Rename ray scattered and color scattered

trevordblack:
Fixes #981

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -1747,7 +1747,6 @@ a noisier image:
             color color_from_scatter =
                 (attenuation * scattering_pdf * ray_color(scattered, depth-1)) / pdf;
 
-
             return color_from_emission + color_from_scatter;
         }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


books/RayTracingTheRestOfYourLife.html
```diff
@@ -3171,9 +3171,11 @@
             ray scattered = ray(rec.p, p.generate(), r.time());
             auto pdf_val = p.value(scattered.direction());
 
-            return emitted
-                + srec.attenuation * rec.mat->scattering_pdf(r, rec, scattered)
-                                   * ray_color(scattered, depth-1) / pdf_val;
+            color scattered_color = (srec.attenuation
+                * rec.mat->scattering_pdf(r, rec, scattered)
+                * ray_color(scattered, depth-1)) / pdf_val;
```
hollasch, line 3176: What would you think about something like the following?

```
auto sample_color = ray_color(scattered, depth-1);
auto scattering_pdf = rec.mat->scattering_pdf(r, rec, scattered);
color scattered_color = (srec.attenuation * scattering_pdf * sample_color) / pdf_val;
```

hollasch (changes requested):
My one comment is more a suggestion than a request.

trevordblack:
I went through a used a model that was I was comfortable with.

I also went through and removed a lot of inconsistencies, and spun a bug #988

trevordblack:
@hollasch Please take a look

books/RayTracingTheRestOfYourLife.html
trevordblack: Let me think about this one. I've created a separate issue #989  to think about it.

hollasch:
Feels good, man.

PR #1001, opened 2022-04-12
Book 3 Section 12.1 Listing 33: fix typos in the book

shadow-lr:
url is https://raytracing.github.io/books/RayTracingTheRestOfYourLife.html#cleaninguppdfmanagement/diffuseversusspecular
image

The correct way to write it should be:
```cpp
srec.pdf_ptr = make_shared<cosine_pdf>(rec.normal);
```
image


trevordblack:
can you check against the `dev-major` branch.
Can you open an issue?

shadow-lr:
> 你能检查一下`dev-major`分支吗？ 你能打开一个问题吗？
I just read that the corresponding content in the `dev-major` branch is right, ok I'll open an issue

Issue #992, opened 2022-02-10
Rewrite book 1 diffuse in more straightforward way


trevordblack:
...

> In think that when implementing recursive bouncing for the first time, sampling a direction uniformly is one of the first things to try. 

This is actually how I would rewrite the chapter. Start with the "naive" approach, and then correct toward cosine law. We start on cos^2 today (I literally cannot remember if it's cos^3 or cos^2) because that was the outcome of the algorithm in the initial version of the book, and so we say upfront that that way of doing it was wrong. With enough time past since the initial publication of v1, it might be worth rearranging them.

_Originally posted by @trevordblack in https://github.com/RayTracing/raytracing.github.io/issues/924#issuecomment-913991770_

Issue #980, opened 2022-01-09
Polish Gamma function text and source


trevordblack:
Move the gamma function--which is currently a square root--into it's own distinct function.

Simultaneously, consider correcting to a bt709 function in book 1 or 2

trevordblack:
Closing this out. Will correct to bt709 for book 2

Issue #696, opened 2020-08-14
Book1 Diffuse text is confusing


yig:
The text around the correct cos(𝜙) distribution is confusing. Sections 8.6 "An Alternative Diffuse Formulation" and 8.5 "True Lambertian Reflection" both seem to me to be correcting Section 8.1 "A Simple Diffuse Material" by claiming to describe a correct cos(𝜙) distribution. However, the pictures aren't the same, so I must be misunderstanding.

Moreover, in Section 8.6 "True Lambertian Reflection", the text above code listing 37 says, "This is achieved by picking points on the surface of the unit sphere, offset along the surface normal. Picking points on the sphere can be achieved by picking points in the unit ball, and then normalizing those." Based on this, I would have expected code listing 37 to make use of "random_in_unit_sphere()". Instead, it does something else entirely that is not explained. (It generates a z value and an xy-plane angle, and then chooses x and y such that they have the angle and the point is on the unit sphere).

trevordblack:
Closing this out with the landing of #1010

PR #1010, opened 2022-04-26
Walk through book 1 and found lots of edits

trevordblack:
A _LOT_ of small touches to the first half of book 1.
A couple of _big_ touches to chapter 8 of book 1

There are many issues which I documented into separate issues as I went along.

This is more of a rough draft than a first draft. Comments are appreciated.

Affects:
#992, #696, #980


books/RayTracingInOneWeekend.html
```diff
@@ -596,29 +604,28 @@
...
 </div>
 
 <div class='together'>
-In graphics, you almost always want your formulas to be in terms of vectors so all the x/y/z stuff
-is under the hood in the `vec3` class. You might note that the vector from center
-$\mathbf{C} = (C_x,C_y,C_z)$ to point $\mathbf{P} = (x,y,z)$ is $(\mathbf{P} - \mathbf{C})$, and
-therefore
+In graphics, you almost always want your formulas to be in terms of vectors so that all the
+$x$/$y$/$z$ stuff can be simply represented using a `vec3` class. You might note that the vector
+from center $\mathbf{C} = (c_x, c_y, c_z)$ to point $\mathbf{P} = (x,y,z)$ is
```
hollasch, line 621: I think we should continue with Cx, Cy and Cz, the components of *C* — that is, points capitalized and bold, components normal weight but subscripted from the capitalized letter.

hollasch (changes requested):

books/RayTracingInOneWeekend.html
```diff
@@ -1640,34 +1674,23 @@
...
-normal of the surface. The sphere with a center at $(\mathbf{P} - \mathbf{n})$ is considered
-_inside_ the surface, whereas the sphere with center $(\mathbf{P} + \mathbf{n})$ is considered
-_outside_ the surface. Select the tangent unit radius sphere that is on the same side of the surface
-as the ray origin. Pick a random point $\mathbf{S}$ inside this unit radius sphere and send a ray
-from the hit point $\mathbf{P}$ to the random point $\mathbf{S}$ (this is the vector
-$(\mathbf{S}-\mathbf{P})$):
+They might also be absorbed rather than reflected. The darker the surface, the more likely
+the ray is absorbed (that’s why it's dark!). Really any algorithm that randomizes direction will
+produce surfaces that look matte. Let's start with the most intuitive: a surface that randomly
+bounces a ray equally in all directions. For this material any ray that hits the surface has an
```
hollasch, line 1681: "material, ..."

books/RayTracingInOneWeekend.html
```diff
@@ -1696,16 +1742,53 @@
...
+    [Listing [random-unit-vec]: <kbd>[vec3.h]</kbd> Random vector on the unit sphere]
+
+And now that we have a random vector on the surface of the unit sphere, we can determine if it is on
+the correct hemisphere by comparing against the surface normal:
+
+  ![Figure [normal-hor]: The normal vector tells us which hemisphere we need
+  ](../images/fig-1.12-surface-normal.jpg)
+
+We can take the dot product of the normal and our random vector to determine if it's in the correct
+hemisphere: If the dot product is positive then the vector is in the correct hemisphere, if
```
hollasch, line 1765: "If the dot product is positive, then ... . If the dot product is negative, then ... ."

books/RayTracingInOneWeekend.html
```diff
@@ -1696,16 +1742,53 @@
...
+            return on_unit_sphere;
+        else
+            return -on_unit_sphere;
+    }
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [random-in-hemisphere]: <kbd>[vec3.h]</kbd> The random_in_hemisphere() function]
+</div>
+
+<div class='together'>
+We'll set up our diffuse material to only return 50% of the color from a bounce and add it to the
```
hollasch, line 1781: "only return 50%" — a quick comment why would be nice.

books/RayTracingInOneWeekend.html
```diff
@@ -1849,9 +1957,13 @@
 Fixing Shadow Acne
 -------------------
 <div class='together'>
-There’s also a subtle bug in there. Some of the reflected rays hit the object they are reflecting
-off of not at exactly $t=0$, but instead at $t=-0.0000001$ or $t=0.00000001$ or whatever floating
-point approximation the sphere intersector gives us. So we need to ignore hits very near zero:
+There’s also a subtle bug that we need to address. A ray reflecting off of a surface might have a
+floating point round off error when calculating for its origin. This means that the origin of a
+reflected ray is unlikely to be perfectly flush with the surface that it is reflecting off of. It
```
hollasch, line 1962: "off of" → "from".

books/RayTracingInOneWeekend.html
```diff
@@ -1849,9 +1957,13 @@
...
+might be just above the surface, or it might be just below the surface. If the ray's origin is just
+below the surface then it could intersect with that surface again. Which means that it will find
+the nearest surface at $t=0.00000001$ or whatever floating point approximation the hit function
+gives us. The simplest hack to address this is just to ignore hits that very near zero:
```
hollasch, line 1966: "very near zero" → "very close to the original intersection point".

books/RayTracingInOneWeekend.html
```diff
@@ -1884,37 +1996,31 @@
...
-
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-    vec3 random_unit_vector() {
-        return unit_vector(random_in_unit_sphere());
-    }
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [random-unit-vector]: <kbd>[vec3.h]</kbd> The random_unit_vector() function]
+There are two unit radius spheres tangent to the hit point $p$ of a surface. These two spheres have
+a center of $(\mathbf{P} + \mathbf{n})$ and $(\mathbf{P} - \mathbf{n})$, where $\mathbf{n}$ is the
+normal of the surface. The sphere with a center at $(\mathbf{P} - \mathbf{n})$ is considered
```
hollasch, line 2010: How about this instead:
“Since a sphere's normal vector points away from the sphere center, the inside tangent sphere's center is ... and the outside tangent sphere's center is ... ”

books/RayTracingInOneWeekend.html
```diff
@@ -1942,99 +2048,29 @@
...
-
-An Alternative Diffuse Formulation
------------------------------------
-<div class='together'>
-The initial hack presented in this book lasted a long time before it was proven to be an incorrect
-approximation of ideal Lambertian diffuse. A big reason that the error persisted for so long is
-that it can be difficult to:
+Both of these changes are due to the less uniform scattering of the light rays, more rays are
+scattering toward the normal. This means that for diffuse objects, they will appear _darker_
+because fewer light bounces toward the camera. For the shadows, more light bounces straight-up, so
```
hollasch, line 2062: fewer → less

books/RayTracingInOneWeekend.html
```diff
@@ -1942,99 +2048,29 @@
...
+the parts of the larger sphere directly underneath the smaller sphere are darker.
```
hollasch, line 2063: → so the area underneath the sphere is darker

books/RayTracingInOneWeekend.html
```diff
@@ -1942,99 +2048,29 @@
...
-        vec3 in_unit_sphere = random_in_unit_sphere();
-        if (dot(in_unit_sphere, normal) > 0.0) // In the same hemisphere as the normal
-            return in_unit_sphere;
-        else
-            return -in_unit_sphere;
-    }
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [random-in-hemisphere]: <kbd>[vec3.h]</kbd> The random_in_hemisphere(normal) function]
-
+objects behave under light can be poorly formed. Scenes will become more complicated over the course
```
hollasch, line 2067: “As scenes become more complicated over the course of the book, you are encouraged ...”

books/RayTracingInOneWeekend.html
```diff
@@ -1942,99 +2048,29 @@
...
+of the book. You are encouraged to switch between the different diffuse renderers presented here.
+Most scenes of interest will contain a disproportionate amount of diffuse materials. You can gain
```
hollasch, line 2069: disproportionate → large

hollasch (changes requested):
(https://github.com/RayTracing/raytracing.github.io/pull/1010/commits/400845d754612e8d23a883133e26718a08cc1d25#diff-70246b31c62a7a5b08bf2e571e7f003804c4d2ada421b002076340b70553e18d) can be misleading since it looks like the reflecting volume is slightly more than a hemisphere. I wonder if a smaller object or a larger hemisphere would allow us to better illustrate that we're only doing a hemisphere from the hit point.

books/RayTracingInOneWeekend.html
trevordblack: ugh. Wow I hate it.
I understand why AABB has it set up the way that it is, but I can't help but look at that and see a problem.

Basically, we don't have a set syntax for the names of vectors.

I don't disagree with your comment, even though I would vastly prefer lower case here, so I'll be making the change as per your recommendation

trevordblack:
Ugh. Git nonsense.

I've done a right job fouling this up.

I'll address the comments within and spin up a new branch PR

PR #1047, opened 2022-08-05
scene: rename loop variable in pixel sampling loop

### LINKED ISSUE #1043: Rename pixel sample loop iterator from 's' to 'sample'
The single letter variable is confused with the desired `s`/`t` image coordinate parameters.

hollasch:
The old name 's' is easily confused with the image s/t coordinates.

Fixes #1043


trevordblack:
Why the switch from
0 -> samples_per_pixel-1
to
samples_per_pixel -> 1

At least for my brain, i think of samples as counting up either starting at 0 or 1

the zeroth sample, then the first, then the second ...
(or the first sample, then the second, then the third ...)

hollasch:
Good call. When I started I had a vague idea that this would simplify the loop, but the end result is isn't any better. Will revert that.

PR #1051, opened 2022-08-11
Wip/scene render

hollasch:


books/RayTracingInOneWeekend.html
```diff
@@ -484,6 +484,7 @@
 often, so I’ll use a non-square image. For now we'll use a 16:9 aspect ratio, since that's so
 common.
 
+
```
hollasch, line 490: Oops. Gets fixed later.

books/RayTracingInOneWeekend.html
```diff
@@ -90,7 +90,7 @@
 
         std::cout << "P3\n" << image_width << ' ' << image_height << "\n255\n";
 
-        for (int j = image_height-1; j >= 0; --j) {
```
trevordblack, line 93: Why the change?

I don't have a preference either way, but I'm curious the motivation

trevordblack:
Book 1 Figure 3 fig-1.03-cam-geom.jpg needs to be updated for s,t

books/RayTracingInOneWeekend.html
trevordblack: Especially since it's flipping the first rendered texture

books/RayTracingInOneWeekend.html
```diff
@@ -547,13 +548,13 @@
...
-            std::clog << "\rScanlines remaining: " << j << ' ' << std::flush;
+        for (int j = 0; j < image_height; ++j) {
+            std::clog << "\rScanlines remaining: " << (image_height - j) << ' ' << std::flush;
             for (int i = 0; i < image_width; ++i) {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-                auto u = double(i) / (image_width-1);
-                auto v = double(j) / (image_height-1);
-                ray r(origin, lower_left_corner + u*horizontal + v*vertical - origin);
```
trevordblack, line 556: If I understand correctly, flipping the count of `j` is the cause of the `(1-t)` in `(1-t)*vertical`

If you're going to stick with this updated convention, I'm going to ask you to justify the `(1-t)` in the text

books/RayTracingInOneWeekend.html
```diff
@@ -1489,7 +1490,11 @@
         }
 
         ray get_ray(double s, double t) const {
-            return ray(origin, lower_left_corner + s*horizontal + t*vertical - origin);
+            // Return the ray from the projection point to the indicated pixel. Coordinates s,t are
```
trevordblack, line 1525: This needs to be in text. Not comments.

books/RayTracingInOneWeekend.html
```diff
@@ -1590,15 +1595,16 @@
...
+
             for (int i = 0; i < image_width; ++i) {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
                 color pixel_color(0, 0, 0);
                 for (int sample = 0; sample < samples_per_pixel; ++sample) {
-                    auto u = (i + random_double()) / (image_width-1);
-                    auto v = (j + random_double()) / (image_height-1);
-                    ray r = cam.get_ray(u, v);
+                    auto s = (i + random_double()) / (image_width-1);
+                    ray r = cam.get_ray(s, t);
```
trevordblack, line 1642: Did you drop the definition of t?

trevordblack: Oh, I see it's up above.
Hmm.

It's more clear when s and t are defined together. But it's a waste to reassign j every iteration of i.

I have a mild preference for how it was previously for clarity.

books/RayTracingInOneWeekend.html
```diff
@@ -473,31 +473,38 @@
...
 ----------------------------
-Now we are ready to turn the corner and make a ray tracer. At the core, the ray tracer sends rays
-through pixels and computes the color seen in the direction of those rays. The involved steps are
-(1) calculate the ray from the eye to the pixel, (2) determine which objects the ray intersects, and
-(3) compute a color for that intersection point. When first developing a ray tracer, I always do a
-simple camera for getting the code up and running. I also make a simple `ray_color(ray)` function
-that returns the color of the background (a simple gradient).
+Now we are ready to turn the corner and make a ray tracer.
+At the core, the ray tracer sends rays through pixels and computes the color seen in the direction
+of those rays.
```
trevordblack, line 478: Should this be on one line?

books/RayTracingInOneWeekend.html
```diff
@@ -501,6 +501,25 @@
...
 negative z-axis.
+
+Now the inevitable tricky part.
+While our 3D space has the conventions above, this conflicts with the standard _image_ coordinates,
+where the Y axis is inverted.
+In image coordinates, Y=0 denotes the top of the image, and Y=1 denotes the bottom.
+
+(This is very handy if your pixels are read out from a scanline buffer to feed an electron beam that
+scans top to bottom, left to right, exciting screen phosphors as it goes.
+Your display doesn't work like that?
```
trevordblack, line 512: I think a ton of our audience is going to mistake this joke.

Consider rewriting for obviousness, or remove

books/RayTracingInOneWeekend.html
```diff
@@ -509,8 +528,8 @@
   ![Figure [cam-geom]: Camera geometry](../images/fig-1.03-cam-geom.jpg)
 
 <div class='together'>
-Below in code, the ray `r` goes to approximately the pixel centers (I won’t worry about exactness
-for now because we’ll add antialiasing later):
+Below in code, the ray `r` goes to the current image pixel (I won’t worry about exactness for now
+because we’ll add antialiasing later):
```
trevordblack, line 532: The rewrite makes things less clear.

In saying "I won't worry about exactness for now" it sounds like we might be accidentally hitting nearby pixels. Which is not the intent of the qualifier, rather that we might not peg the exact pixel center.

Consider reverting, qualifying the exactness comment, or removing the exactness comment

books/RayTracingInOneWeekend.html
```diff
@@ -581,11 +600,12 @@
 gradient to the color in addition to the vertical gradient.
 
 I then did a standard graphics trick of scaling that to $0.0 ≤ t ≤ 1.0$.
-When $t = 1.0$ I want blue.
-When $t = 0.0$ I want white.
+(In this context, the parameter $t$ is different from the image coordinate $t$ above.)
```
trevordblack, line 603: Please consider changing the variable from `t`. I'm unhappy with aliasing this early in the series

books/RayTracingInOneWeekend.html
trevordblack: It looks like you do in a later commit?

trevordblack:
I'll need to go through this and do a rather hairy merge with my book 1 changes as well.

So, don't worry about being absolutely perfect with this one. Just land with your take on comments I gave above, and then I'll read back through everything after I figure out my merge.

hollasch:
Yes, that's going to be updated after all of the changes to the `camera` class are in.

books/RayTracingInOneWeekend.html
hollasch: Pixels have coordinates in 3D space which trivially mirror to the XY axis in world 3-space, but that only applies for the initial camera position & orientation and the placement of the image plane. As always, there's conflict with the Y axis. Generally, I prefer to keep spaces consistent instead of having to remember what my context is in the different space.

In this case, I want the following: the `scene` class is responsible for interrogating pixels, based on their image coordinates (which are Y down). The `camera` class is positioned and oriented in 3-space, but uses image coordinate conventions when interrogating specific pixels. Thus the `camera.get_ray()` method takes in pixel coordinates, which are now consistently in image coordinates.

The flipped rendered texture is no big deal, since there's no correct orientation. I plan to re-render this image during the v4.0.0 prep phase.

books/RayTracingInOneWeekend.html
hollasch: Yup, this is addressed in a subsequent pass. I just did code changes first.

books/RayTracingInOneWeekend.html
hollasch: I'd actually rather we did better commenting the code, instead of telling folks to "read the book". Particularly in cases here where we have an implicit change of coordinate systems. Before the change was sort of smeared across the call stack, but now at least we have a crisp point where the flip happens.

books/RayTracingInOneWeekend.html
hollasch: I waffled a bit on this one as well. One point in favor of the constant hoisting is that it's a situation that comes up often in raytracing, where you have scanline loops and pixel loops.

Sort of an uncommented teaching moment, but I also see the advantage of keeping things simpler but a bit slower.

books/RayTracingInOneWeekend.html
hollasch: Yes. It all renders that way, and will be a question that we can revisit when it comes time to prep for release. I'm fine with reflowing everything at that point, but it's a bit nicer to review text changes this way. If I'm wrong on that, I can reflow again at any point.

books/RayTracingInOneWeekend.html
hollasch: Yeah, I'll drop.

books/RayTracingInOneWeekend.html
hollasch: I was a bit puzzled about the exactness comment as well. I don't think it's needed; will drop.

books/RayTracingInOneWeekend.html
hollasch: This is one problem with adopting _s_,_t_ as image coordinates, because _t_ as a parametric variable comes up often. I'm certainly open to changing this particular case, but it's elsewhere, I think.

hollasch: Hmmm. When lerping, this is frequently done with α. I'll at least change this case to use _a_.

hollasch: Heh. And then immediately below, the ray-sphere intersection code uses _t_ as the ray-sphere intersection parameter. `¯\_(ツ)_/¯`. I'll leave that one.

hollasch:
Oh, reading your last comment, looks like you want me to go ahead and merge. Will rebase first. Let me know if you still want changes after it's merged.

hollasch:
Reworked PR as #1053. Closing this one.

PR #1052, opened 2022-08-12
Add missing interval from intervals constructor

### LINKED ISSUE #1014: merging intervals is not defined in the books
Book1 and Book2 don't include following Interval constructor: 

https://github.com/RayTracing/raytracing.github.io/blob/book1-review/src/common/interval.h#L18

It's first used in Listing 14 of Book2

hollasch:
Resolves #1014


trevordblack (approved):
ship it

PR #1061, opened 2022-08-22
Separate out ternary results with space

hollasch:
Formatting improvement only


trevordblack (approved):
An unconditional approve

Issue #682, opened 2020-07-27
Camera UV coordinates — handle images with width or height of one


ghost:
I'm not quite sure how the current equation was derived. But the case clearly fails for the case that the width and height are both equal to one, for example.

```cxx
int x = 0;
int y = 0;
int width = 1;
int height = 1;
float u = (x / (width - 1)); // 0 / (1 - 1) = 0 / 0 = NaN
float v = (y / (height - 1));
```

Should be changed to:

```cxx
float u = (x + 0.5f) / width;
float v = (y + 0.5f) / height;
```

The result can be seen clearly when you can a 2x2 image as an example.

The UV coordinates previously would be:

```
(0, 0) (1, 0)
(0, 1) (1, 1)
```

The correct result would be:

```
(0.25, 0.25) (0.75, 0.25)
(0.25, 0.75) (0.75, 0.75)
```

hollasch:
Will add a special case to handle single-pixel images. I definitely do *not* want to promote the idea that a pixel is square, or has any specific shape, or indeed even that it has a finite kernel.

I'll also revisit the text to ensure that we at least hint that square pixels are a poor (but cheap) model.

trevordblack:
I forget where I read it, but the correct uv should be:
```
float u = floor((x + 0.5f) / width);
float v = floor((y + 0.5f) / height);
```

tay10r:
@hollasch

This was my issue from a previous account. I wasn't proposing to tell readers that a pixel is a square, just offering a more simpler and commonly used expression of UV coordinates that does not require any special corner cases. You're obviously free to reject it, that's totally fine.

tay10r:
@trevordblack

Thanks for the suggestion! This:

```cpp
float u = floor((x + 0.5f) / width);
float v = floor((y + 0.5f) / height);
```

Is not right because it will round to either 0 or 1 for all UV coordinates. So unless you have a 2x2 image, then you would get incorrect results. Using `floor` is more likely to be useful when converting from UV coordinates to image coordinates. For example:

```cpp
int x = floor(uv_x * width);
int y = floor(uv_y * height);
```
But you can just use native floating-point to integer conversion instructions, which performs the truncation along with the conversion.

```cpp
int x = uv_x * width;
int y = uv_y * height;
```

hollasch:
Fixed in code in PR #1154 (text update pending).

PR #1068, opened 2022-08-27
Wip/pixel jitter

hollasch:
Draft for review before making the corresponding changes to the text.

### DIFF
--- a/src/common/camera.h
+++ b/src/common/camera.h
@@ -16,8 +16,8 @@
 
 class camera {
   public:
-    void initialize(double image_width, double image_height) {
-        auto aspect_ratio = image_width / image_height;
+    void initialize(int image_width, int image_height) {
+        auto aspect_ratio = static_cast<double>(image_width) / image_height;
         auto theta = degrees_to_radians(vfov);
         auto h = tan(theta/2);
 
@@ -39,14 +39,9 @@ class camera {
         lens_radius = aperture / 2;
     }
 
-    ray get_ray(double s, double t) const {
-        // Return the ray from the projection point to the indicated pixel. Coordinates s,t are
-        // the normalized image-based coordinates of the pixel. Image left is s=0, image right
-        // is s=1, image top is t=0, image bottom is t=1.
-
-        // Jitter the sample location of the pixel.
-        s += random_double(-0.5, 0.5) * pixel_delta_x;
-        t += random_double(-0.5, 0.5) * pixel_delta_y;
+    ray get_ray(int i, int j) const {
+        auto s = (i + random_double(-0.5, 0.5)) * pixel_delta_x;
+        auto t = (j + random_double(-0.5, 0.5)) * pixel_delta_y;
 
         vec3 rd = lens_radius * random_in_unit_disk();
         vec3 offset = u * rd.x() + v * rd.y();
--- a/src/common/camera.h
+++ b/src/common/camera.h
@@ -16,11 +16,16 @@
 
 class camera {
   public:
-    void initialize(double aspect_ratio = 1.0) {
+    void initialize(double image_width, double image_height) {
+        auto aspect_ratio = image_width / image_height;
         auto theta = degrees_to_radians(vfov);
         auto h = tan(theta/2);
+
         auto viewport_height = 2.0 * h;
         auto viewport_width = aspect_ratio * viewport_height;
+
+        pixel_delta_x = 1.0 / (image_width-1);
+        pixel_delta_y = 1.0 / (image_height-1);
 
         w = unit_vector(lookfrom - lookat);
         u = unit_vector(cross(vup, w));
@@ -39,8 +44,13 @@ class camera {
         // the normalized image-based coordinates of the pixel. Image left is s=0, image right
         // is s=1, image top is t=0, image bottom is t=1.
 
+        // Jitter the sample location of the pixel.
+        s += random_double(-0.5, 0.5) * pixel_delta_x;
+        t += random_double(-0.5, 0.5) * pixel_delta_y;
+
         vec3 rd = lens_radius * random_in_unit_disk();
         vec3 offset = u * rd.x() + v * rd.y();
+
         const auto ray_time = random_double(0.0, 1.0);
 
         return ray(
@@ -60,6 +70,7 @@ class camera {
     vec3   vup      = vec3(0,1,0);
 
   private:
+    double pixel_delta_x, pixel_delta_y;
     point3 origin;
     point3 lower_left_corner;
     vec3 horizontal;
--- a/src/InOneWeekend/scene.h
+++ b/src/InOneWeekend/scene.h
@@ -22,18 +22,18 @@ class scene {
     void render() {
         const int image_height = static_cast<int>(image_width / aspect_ratio);
 
-        cam.initialize(aspect_ratio);
+        cam.initialize(image_width, image_height);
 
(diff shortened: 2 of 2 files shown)


src/common/camera.h
```diff
@@ -16,12 +16,17 @@
 
 class camera {
   public:
-    void initialize(double aspect_ratio = 1.0) {
+    void initialize(double image_width, double image_height) {
```
trevordblack: Do we want to set parameter values for image_width and image_height?

I think no

src/common/camera.h
```diff
@@ -16,12 +16,17 @@
...
+        auto aspect_ratio = image_width / image_height;
         auto theta = degrees_to_radians(vfov);
         auto h = tan(theta/2);
+
         auto viewport_height = 2.0 * h;
         auto viewport_width = aspect_ratio * viewport_height;
 
+        pixel_delta_x = 1.0 / (image_width-1);
+        pixel_delta_y = 1.0 / (image_height-1);
```
trevordblack, line 28: shouldn't this just be 1.0 / image_height ?

In the case where there are two pixels, shouldn't delta be 1/2?

src/common/camera.h
```diff
@@ -39,8 +44,13 @@ class camera {
         // the normalized image-based coordinates of the pixel. Image left is s=0, image right
         // is s=1, image top is t=0, image bottom is t=1.
 
+        // Jitter the sample location of the pixel.
+        s += random_double(-0.5, 0.5) * pixel_delta_x;
+        t += random_double(-0.5, 0.5) * pixel_delta_y;
```
trevordblack: This isn't the best place to put this comment, but it's expediant.

Should we be thinking about centering the samples at the center of the pixel? rather than as it is currently at at the edges of pixels?

i.e.
```
        s += 0.5 * pixel_delta_x + random_double(-0.5, 0.5) * pixel_delta_x;
        t += 0.5 * pixel_delta_y + random_double(-0.5, 0.5) * pixel_delta_y;
```
It's strictly more "correct", but it might be additional unnecessary confusion

src/common/camera.h
```diff
@@ -39,8 +44,13 @@ class camera {
         // the normalized image-based coordinates of the pixel. Image left is s=0, image right
         // is s=1, image top is t=0, image bottom is t=1.
 
+        // Jitter the sample location of the pixel.
```
trevordblack: Jitter is not quite correct here.

Jitter is the colloquial term given for the placement of samples within a _stratified sampling_.

That's not exactly what we're doing. We're just randomly sampling within a pixel. It's not stratified within a pixel, so it's not jitter.

Replace with something like
// Randomly place the pixel sample location

src/common/camera.h
```diff
@@ -16,8 +16,8 @@
 
 class camera {
   public:
-    void initialize(double image_width, double image_height) {
-        auto aspect_ratio = image_width / image_height;
+    void initialize(int image_width, int image_height) {
+        auto aspect_ratio = static_cast<double>(image_width) / image_height;
```
trevordblack, line 20: good catch

trevordblack:
We should take about the pixel delta and sampling.

This is something that #682 is kind of getting at, but i'll elaborate here.

Right now, for the case where there is 1 pixel by 1 pixel, divide by zero. Dead.

In the case where we are sampling 2 pixels by 2 pixels it's also problematic (although less so):
top-left pixel
s [-0.5, 0.5]
t [-0.5, 0.5]
top-right
s [0.5, 1.5]
t [-0.5, 0.5]
bottom-left
s [-0.5, 0.5]
t [0.5, 1.5]
bottom-right
s [0.5, 1.5]
t [0.5, 1.5]

Which should be fine, except that we explicitly declared the top left of the image (0,0), and said that s and t are bound by 0 and 1. That's not correct at the moment. also, only half of the samples will be within s,t [0.0, 1.0]

For n x m pixels,
s will be bound within [0 -  1/(2(n - 1)) , 1 + 1/(2(n - 1)) ]
t will be bound within [0 -  1/(2(m - 1)) , 1 + 1/(2(m - 1)) ]

Which, yes, tends to zero (and is effectively zero for 1000s of n,m) but it's still wrong.
It took pains to make sure this was correct in book 3. I think that we should do the same here.

src/common/camera.h
trevordblack: See distinct comment in the main PR chain

trevordblack:
For my own thoroughness/sanity, the above comment would need the following

```
        pixel_delta_x = 1.0 / image_width;
        pixel_delta_y = 1.0 / image_height;
        
        ...
        
        auto pixel_center_x = (i + 0.5) * pixel_delta_x;
        auto pixel_center_y = (j + 0.5) * pixel_delta_y;
        
        auto s = pixel_center_x + random_double(-0.5, 0.5) * pixel_delta_x;
        auto t = pixel_center_y + random_double(-0.5, 0.5) * pixel_delta_y;
```

This is two changes from the current code:
1. today there is an implicit `pixel_corner_*` value
2. the delta value is a tiny bit too big today (n+1/n)

This is what we're doing now:
```
        pixel_delta_x = 1.0 / (image_width - 1);
        pixel_delta_y = 1.0 / (image_height - 1);
        
        ...
        
        auto pixel_corner_x = i * pixel_delta_x;
        auto pixel_corner_y = j * pixel_delta_y;
        
        auto s = pixel_corner_x + random_double(-0.5, 0.5) * pixel_delta_x;
        auto t = pixel_corner_y + random_double(-0.5, 0.5) * pixel_delta_y;
```

We could change both of the above two (sampling from pixel corner or correcting delta value) using my code above.

Or we could sample from the top-left corner of a pixel:
```
        pixel_delta_x = 1.0 / image_width;
        pixel_delta_y = 1.0 / image_height;
        
        ...
        
        auto pixel_corner_x = i * pixel_delta_x;
        auto pixel_corner_y = j * pixel_delta_y;
        
        auto s = pixel_corner_x + random_double() * pixel_delta_x;
        auto t = pixel_corner_y + random_double() * pixel_delta_y;
```

trevordblack:
Why did we go with
```
        pixel_delta_x = 1.0 / (image_width - 1);
        pixel_delta_y = 1.0 / (image_height - 1);
```
I do not remember

src/common/camera.h
hollasch: I'm not sure I understand — are you raising the possibility of default values? If so, I'd also fall back to no.

trevordblack:
Where does this stand currently?

hollasch:
I just cleared out all of the random branches on this, starting anew based on feedback from the prior PR. Finally working on it again.

trevordblack:
Oh wonderful

src/common/camera.h
trevordblack: Sorry, yes, default values. So no.

src/common/camera.h
hollasch: Here, the pixel delta is the stride from one pixel location to the next. The intent is to have a delta that can be multiplied by the pixel index, and added to the image origin to get the location of the indexed pixel.

src/common/camera.h
hollasch: Excellent point. Will do.

src/common/camera.h
hollasch: This will get addressed when I rework the pixel math. To recap from our (now lost) Slack conversation, I plan to focus on code that deals with pixel _locations_, and separate code that takes that location and produces pixel samples, not necessarily from a box around the pixel location.

PR #1079, opened 2022-10-07
Update RayTracingTheNextWeek.html

zhouhang95:
remove unuseful code


hollasch:
Please read CONTRIBUTING.md before creating pull requests.

This change is would modify the master branch, which is not how we work. In addition, this code (and the books) have been significantly changed in the current development branch, `dev-major`, so the proposed changes no longer apply.

Issue #1065, opened 2022-08-27
Book 1 Chapter 7.2 `dev_major`: `render` part of code has bug (sphere shape is unstable)


LollipopFt:
with this code:
https://github.com/RayTracing/raytracing.github.io/blob/806ab499cc220d044f4abff57fe01612bd9a6e82/books/RayTracingInOneWeekend.html#L1626-L1633

image output is unstable:
`1st run`
![image](https://user-images.githubusercontent.com/62802897/187018469-6d528fc2-3e74-48ba-aa0b-728e609b704d.png)
`2nd run`
![image](https://user-images.githubusercontent.com/62802897/187018477-cd8f3da7-ce6c-4432-aeae-5c12e3c003ef.png)
`3rd run`
![image](https://user-images.githubusercontent.com/62802897/187018489-792ac6fb-b126-48bf-9ef4-df614e2dfc24.png)

and many more. notice how unstable the bottom and top of the sphere is. however, with the original code in `master`, a stable image is produced:
`1st run`
![image](https://user-images.githubusercontent.com/62802897/187018548-5cf2dcc2-761e-499c-be48-e144faa03373.png)
`2nd run`
![image](https://user-images.githubusercontent.com/62802897/187018560-e9c435e6-ec0c-4691-b381-fe505577fb89.png)
`3rd run`
![image](https://user-images.githubusercontent.com/62802897/187018570-ab55b981-b54c-4bc6-9ae0-e7794e83b99d.png)

xw901103:
put the line under row(height) for loop `auto t = (j + random_double()) / (image_height-1);` inside antialiasing sampling for loop. your code should looks like following.

for (int j = 0; j < image_height; ++j) {
        std::clog << "\rScanlines remaining: " << (image_height - j) << ' ' << std::flush;

        for (int i = 0; i < image_width; ++i) {
            color pixel_color(0, 0, 0);
            for (int sample = 0; sample < samples_per_pixel; ++sample) {
                auto t = (j + random_double()) / (image_height-1);
                auto s = (i + random_double()) / (image_width-1);
                std::clog << t << std::endl;
                ray r = cam.get_ray(s, t);
                pixel_color += ray_color(r, world, max_depth);
            }
            write_color(std::cout, pixel_color, samples_per_pixel);
        }
    }

if you leave that t value only changes every time you go to next scanline, you will have horizontal sampling insufficient which looks like the results you posted above.

LollipopFt:
[...]

PR #1097, opened 2022-12-20
Fix jitter sample bug

hollasch:
Fixes a bug I introduced with the camera rewrite, where I thought I was optimizing things.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -1620,15 +1620,12 @@ $t$ coordinate out of the inner loop, since it is the same value for the entire 
 
         for (int j = 0; j < image_height; ++j) {
             std::clog << "\rScanlines remaining: " << (image_height - j) << ' ' << std::flush;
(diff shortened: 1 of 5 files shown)


trevordblack:
I might not be able to get around to this until the weekend. Is that alright?

hollasch:
No rush.

trevordblack:
I'm not sure what you're fixing here, `auto t` is inside of the outer for loop on lines 1624, 1804, 2850, etc.

I'm approving this commit to unblock you

trevordblack (approved):

hollasch:
The bug here is that we're generating `samples_per_pixel` random s,t coordinates, but the result of my earlier "optimization" (moving the t-calculation outside the innermost loop) is that all samples will be along a constant t line, instead of within in the sampling box.

hollasch:
This showed up as odd/significant aliasing for horizontally-oriented edges.

hollasch:
This fixed #1065.

PR #1112, opened 2023-05-10
Scene changes

hollasch:
This is a draft PR, before I update the book text to match. What do you think about these changes?

### DIFF
--- a/src/InOneWeekend/main.cc
+++ b/src/InOneWeekend/main.cc
@@ -20,10 +20,7 @@
 
 
 void random_spheres(scene& scene_desc) {
-    scene_desc.image_width       = 1200;
-    scene_desc.aspect_ratio      = 16.0 / 9.0;
-    scene_desc.samples_per_pixel = 10;
-
+    scene_desc.aspect_ratio   = 16.0 / 9.0;
     scene_desc.cam.vfov       = 20;
     scene_desc.cam.focus_dist = 10.0;
     scene_desc.cam.aperture   = 0.1;
@@ -78,5 +75,5 @@ void random_spheres(scene& scene_desc) {
 int main() {
     scene scene_desc;
     random_spheres(scene_desc);
-    scene_desc.render();
+    scene_desc.render(1200, 10);
 }
--- a/src/InOneWeekend/scene.h
+++ b/src/InOneWeekend/scene.h
@@ -19,20 +19,20 @@
 
 class scene {
   public:
-    void render() {
-        const int image_height = static_cast<int>(image_width / aspect_ratio);
+    void render(int resolution_x, int samples_per_pixel) {
+        const int resolution_y = static_cast<int>(resolution_x / aspect_ratio);
 
         cam.initialize(aspect_ratio);
 
-        std::cout << "P3\n" << image_width << ' ' << image_height << "\n255\n";
+        std::cout << "P3\n" << resolution_x << ' ' << resolution_y << "\n255\n";
 
-        for (int j = 0; j < image_height; ++j) {
-            std::clog << "\rScanlines remaining: " << (image_height - j) << ' ' << std::flush;
-            for (int i = 0; i < image_width; ++i) {
+        for (int j = 0; j < resolution_y; ++j) {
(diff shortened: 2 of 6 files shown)


hollasch:
This is preliminary to the camera changes, but not required. Since the `scene` class is new to v4.0.0, I don't feel much resistance to tweaking it. Still, it might be a bit fussy. Won't hurt my feelings if you don't like it.

trevordblack (commented):
Two questions.

1. Why did we decide to go with:
```
    scene scene_desc;
    random_spheres(scene_desc);
    scene_desc.render();
```
Instead of:
```
    scene scene_desc = random_spheres()
    scene_desc.render();
```

2. Why are you keeping aspect ratio part of scene? That really feels like something that should be part of camera?
...
Indeed, it feels like maybe max_depth should also be a camera thing?
```
pixel_color += ray_color(r, cam.max_depth);
```
...
Wait, sorry, it's been literally years since I walked this code. What is our mental model for what a scene is? Is it everything that is needed to create a picture, or is it just the collection of geometry?

hollasch:
Yeah, honestly it's still somewhat awkward.

_Why `random_spheres(scene_desc)` instead of `scene_desc = random_spheres()`?_

This allows the program to set scene defaults, and then the scene generator can decide to override these for "important" aspects. We could adopt the suggested approach; we'd just need to repeat the three defaults (background color, camera up, and camera focus-distance) across all ten scenes. Not terrible.

_Why is `aspect_ratio`  part of the scene instead an aspect of the camera?_

I was thinking that a scene had a definite preference for a particular aspect ratio (for example, a landscape scene versus the Cornell box), but it's a weak argument. It certainly makes sense as a camera property.

_Maybe `max_depth` should also be a camera thing._

Max depth kind of depends on the scene. For example, if my scene is a bunch of glass and mirror balls, a high max depth is called for. If it's a Cornell box with all diffuse surfaces, then a low max depth is appropriate. Again, this is a fairly fuzzy idea, because it's really a rendering parameter than a construction of the scene.

_What's our mental model?_

Very hand-wavy. Ideally, the scene describes the geometry and lighting, independent of how it's viewed, and the camera is tasked with everything needed to construct an image of that scene. And way, I've clearly violated this model in several prickly ways.

Your questions are spot on, and make me more and more uncomfortable with both how things are and some of my suggestions. I don't think we need to be slaves to the mental model, and our code is often very pragmatic, but I'm also nervous about confusing the reader with the handful of exceptions.

trevordblack:
My concern is making the changes above is sacred cows effectively.

I checked and the last time I really dug into the code was more than 2 years ago.

The changes I proposed above were kind of obvious concerns for me, but I'm not sure why we haven't made these changes before.

So, I'm either a different programmer than I was before or we've already discussed this and decided on the code in the book. I genuinely don't know.

I also don't know what I think is best, my current preference aligns with the questions above, but I would need to chew on it to come with my ultimate opinion.

As said in the other issue, I've only really got time for smell-tests over the next two weeks.

If this is blocking something else then just use best judgement

hollasch:
Just took this meta conversation to Slack for a tactical review.

For this specific change, I'm going to reverse direction and issue a new PR:

* Camera owns the image/rendering parameters. This will now include aspect ratio, image width, samples per pixel, and maximum ray depth.
* The scene object is thus reduced to a tuple: <world, camera>. In the end, I think this has a simplicity that's clearly superior to all of the fussiness of the above approach.
* The scene generators, as you noted, currently take in a `scene` reference. This allowed me to set common default properties for all scenes, then have the generator functions fill in the rest. In reviewing issues and discussions, I ran across one case where a reader couldn't figure out why his renderings didn't have our background color, and it was because of the fact that we set some scene parameters in one place, and some in another. Again, I think we should remove this complexity, and have the generator functions responsible for _all_ aspects of the scene.
* Regarding your comment about whether the generator functions should _return_ the scene, the second reason for the main loop owning the scene object and passing a reference for the scene generator to fill out is that it makes the memory management trivial, as a naturally destructed stack object, instead of using `std::unique_ptr` (a new concept) or just using `std::shared_ptr` or something. It also opens the doors for readers to chain scene generators, or have functions to tweak cameras (like an animation loop) or whatever. Bottom line is that I still think the simplicity is a good thing.

PR #1116, opened 2023-05-10
Add hints for dealing with bad PPMs on Windows

### LINKED ISSUE #636: Can't open created ppm file (PPM must be ASCII or UTF-8)
I've just started reading/implementing Ray Tracing in One Weekend and already stumbled upon this problem. This is my c++ code:

```
int main() {
  const int image_width = 256;
  const int image_height = 256;

  std::cout << "P3\n" << image_width << ' ' << image_height << "\n255\n";

  for (int y = image_height-1; y >= 0; --y) {
    std::cerr << "\rScanlines remaining: " << y << ' ' << std::flush;
    for (int x = 0; x < image_width; ++x) {
      auto r = double(x) / (image_width-1);
      auto g = double(y) / (image_height-1);
      auto b = 0.25;

      int ir = static_cast<int>(255.999 * r);
      int ig = static_cast<int>(255.999 * g);
      int ib = static_cast<int>(255.999 * b);

      std::cout << ir << ' ' << ig << ' ' << ib << '\n';
    }
  }
}
```

And this is the output in the .ppm file:

P3
256 256
255
0 255 63
1 255 63
2 255 63
3 255 63
4 255 63
5 255 63
6 255 63
7 255 63
8 255 63
9 255 63
10 255 63
...

Tried opening this with Gimp, XNView and IrfanView but they all can't seem to read the file.

hollasch:
Resolves #636


books/RayTracingInOneWeekend.html
```diff
@@ -171,11 +171,19 @@
     [Listing [first-img]: First image output]
 </div>
 
-If it doesn’t, then you probably just have some newlines or something similar that is confusing the
-image reader.
-
-If you want to produce more image types than PPM, I am a fan of `stb_image.h`, a header-only image
-library available on GitHub at https://github.com/nothings/stb.
+If it doesn’t, then you may just have some line-ending differences or something similar that is
```
trevordblack, line 174: There is a grammatical mistake here.

```
It should start something like this:
...
If it doesn't, then you may just have some line-ending differences
```

Should look like this -> you may have line-ending differences

This is incorrect.

If your ppm doesn't look like this, the problem is that you formatted the ppm wrong.

If your ppm _does_ look like this, the problem _may_ be line-endings or the like

```
It should start something like this:
...
If the output doesn't look like this you probably have a mistake in how you are printing out.

If the output does like look the above but you're _not_ getting an image, then you may just have some line-ending differences... "
```

trevordblack (changes requested):
Fix grammatical mistake and looks ok to shape

books/RayTracingInOneWeekend.html
hollasch: Changed to
> If your PPM file doesn't look like this, then double-check our formatting code. If it _does_ look like this, then you may have line-ending differences or something similar that is confusing the image reader.

PR #1118, opened 2023-05-10
Changelog: added bvh_node usage in random_spheres

### LINKED ISSUE #715: Book 2: Add an example for Chapter 3
I really like the style of Book 1, where in each chapter we have some code to test the content of that chapter and see some concrete results. However, when I finished Chapter 3 of Book 2 (Bounding Volume Hierarchies), I realized that although we have lots of new codes, they are **NOT** used in the main program. So I tried to change the main program a little bit to use the codes in Chapter 3, and it turns out very simple changes can greatly improve the performance, which is the reason why we use BVH.

We only need to change the world part from
image
to
image
This simple change demonstrated the contents in Chapter 3 very well. Although the generated image is the same as before, the speed is much faster.

hollasch:
Resolves #715


hollasch:
I'm just going to go ahead and merge this one, since it's pretty trivial. As always, feedback never closes.

PR #1120, opened 2023-05-11
aabb: fix performance puzzle with aabb::hit()

### LINKED ISSUE #817: aabb::hit() has wildly different performance based on the two presented implementations
For now, I have both versions coded up in `aabb.h`, with a preprocessor block that selects the fast (prior) one. We'll need to resolve this before we ship v4.0.0.

hollasch:
It turns out that the slow version, preserved in the original code, had a bug in that it failed to tighten the `ray_t` interval with each successive iteration through the X, Y and Z axes. Instead, it was modifying temporary variables.

This change cleans up the two versions of the `hit()` method, and aligns the text's version with our codebase.

Resolves #817

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -682,14 +682,19 @@ as my go-to method:
         ...
         bool hit(const ray& r, interval ray_t) const {
             for (int a = 0; a < 3; a++) {
-                auto invD = 1.0f / r.direction()[a];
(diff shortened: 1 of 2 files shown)


trevordblack:
Checking this requires a diligence that I'm realistically not going to be able to provide for 2 weeks.

Does this block other issues?

hollasch:
I'm going to merge this for now, as it implements the same fix reported in two separate issues, and because we'll be verifying the render results in the v4.0.0-rel milestone. Added this to the list of async PRs.

PR #1122, opened 2023-05-11
book1: Fix include-guard issue with vec3.h listing

### LINKED ISSUE #1073: Syntax issue on vec3.h in book one
This goes back to #958, the `#endif` should appear at the bottom of the 2nd snippet.

Being where it currently is makes it yield code that will not compile if you just add the first and the second part. That means there's an error somewhere in the text.  The user could assume that the error is in the text "Listing 5: [vec3.h]" presuming that was meant to say "Listing 5: [vec3.cpp]" and thus still get invalid code.

Seeing a newline followed by an `#endif` is basically industry standard for meaning "This is the end of the file" so saying "The second part of the header file" doesn't really make any sense without some type of change.

screen

hollasch:
Resolves #1073

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -227,11 +227,9 @@ In spite of this, we do declare two aliases for `vec3`: `point3` and `color`. Si
 are just aliases for `vec3`, you won't get warnings if you pass a `color` to a function expecting a
 `point3`, for example. We use them only to clarify intent and use.
 
-
-Variables and Methods
-----------------------
 <div class='together'>
-Here’s the top part of my `vec3` class:
+We define the `vec3` class in the top half of a new `vec3.h` header file, and a set of useful vector
+utility functions in the bottom half:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     #ifndef VEC3_H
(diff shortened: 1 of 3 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -289,7 +289,7 @@
     using point3 = vec3;   // 3D point
     using color = vec3;    // RGB color
 
-    #endif
+    ...
```
trevordblack, line 292: Keep the #endif where it is

books/RayTracingInOneWeekend.html
```diff
@@ -305,6 +305,8 @@
 The second part of the header file contains vector utility functions:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+    ...
```
trevordblack, line 308: Continue with a snippet from above

```
    ...
    
    // Type aliases for vec3
    using point3 = vec3;   // 3D point
    using color = vec3;    // RGB color
```

trevordblack: Use highlighting on the changes

hollasch:
Hmmm, generally we do not use highlighting when showing whole new sections added, otherwise the entire section is highlighted, which really isn't helpful.

hollasch:
Actually, the whole "first part immediately followed by the second part" is somewhat clumsy. Perhaps I should just present the entire thing at once, and then adjust the text accordingly?

hollasch:
Ok, took a second pass. I like this one a lot more.

trevordblack (approved):
Looks Good

Issue #1012, opened 2022-04-27
Infinite loop in book 1, simple diffusion material


genmeblog:
I go through the book 1 (book1-review branch) reimplementing it in Clojure.

Listing 40 introduces recursive ray casting which is infinite. When `world.hit` hits a sphere, `rec.p` contains point on sphere. Next hit check always returns the same `rec.p`.
The reason is that `ray_t.contains(...)` in listing 28, checks closed interval (listing 25) instead of open one (as it was defined in listing 15)

genmeblog:
Fixing shadow acne solves this problem but is introduced later (listing 43)

hollasch:
Oof; good catch -- thanks.

hollasch:
See related issue #875 .

Issue #875, opened 2021-03-21
Book 1: Image 8 contains incorrect rendering after Section 8.3 


ronnieholm:
The image rendered at the end of Section 8.3 appears to be the result of fixing shadow acne in Section 8.4, and should probably be placed at the end of Section 8.4.

Placing the image after Section 8.3 causes one to go hunt through the code for bugs because the image is too dark.

hollasch:
We'll double check this during the v4.0.0 sanity pass.

johannmeyer:
I am unable to reproduce image 7 in sec 8.21(The first render of a diffuse sphere) and image 8 in sec 8.22 (gamma correction). My images are much darker than the images shown but removing the shadow acne by setting `t_min = 1e-3` corrects it mostly but the produced images are then lighter than image 7 or 8. Image 9 is reproducible though.

luizfnetto:
> The image rendered at the end of Section 8.3 appears to be the result of fixing shadow acne in Section 8.4, and should probably be placed at the end of Section 8.4.
> 
> Placing the image after Section 8.3 causes one to go hunt through the code for bugs because the image is too dark.

Thanks for your comment. It solved for me. I began to bug hunt and came here before doing a thorough check.

hollasch:
Thank you for the additional information!

webanck:
I concur, image 8 (section 8.3) is much closer to the result I get after taking care of shadow acne (section 8.4), even with the gamma correction.

ISchenhaiqing:
I fix this bug.
```
if (root < t_min || t_max < root) {
        root = (-half_b + sqrtd) / a;
        if (root < t_min || t_max < root)
            return false;
    }
```
**[sphere.h] Ray-sphere intersection code (now)**

```
if (root <= t_min || t_max <= root) {
        root = (-half_b + sqrtd) / a;
        if (root <= t_min || t_max <= root)
            return false;
    }
```
**[sphere.h] Ray-sphere intersection code (after fixing)**
The inconsistency between the 2020 version and the 2016 version results in a different energy transfer that makes the current result darker than the previous one.

tysonliddell:
Confirming @ISchenhaiqing's fix yields an image that seems to match Image 7 in the book much more closely.

#### Before the fix
![before_fix](https://user-images.githubusercontent.com/20079016/224584259-89160f2c-a221-4381-896c-a9bd4d7d7d18.jpg)

#### After the fix
[...]

PR #1123, opened 2023-05-11
Tweak sphere::hit() method to address dark shading

hollasch:
This is in response to issues raised in issue #875, where the sphere rendering image appears darker than it used to. The issue is that points that lie exactly on the ray interval used to be excluded, but then were not. The issue was compounded when we started to use the `interval::contains()` method, because this subtlety was one step removed.

This should fix the issue as raised, but we will verify in the v4.0.0 release milestone, when we regenerate all of the images.


hollasch:
This change should also fix #1012 .

hollasch:
Superseded by PR #1126 .

PR #1124, opened 2023-05-11
Book1: Fix hit_sphere() discriminant test

hollasch:
Earlier in the text, we assert that a positive discriminant indicates two real solutions to the quadratic formula, a negative discriminant indicates no real solutions, and a zero discriminant indicates a single real solution.

This change fixes the hit_sphere() function to properly reflect this.

Resolves #1057


trevordblack (approved):
...

HOW did I miss this???

Issue #1125, opened 2023-05-11
The moving_sphere class is mostly redundant with sphere, and lacks UV coordinates


hollasch:
It should be a simple matter of adding a second constructor to the `sphere` class to add the ability to be a moving sphere.

PR #1128, opened 2023-05-11
Update the sphere class to allow movement

hollasch:
This change deprecates the moving_sphere class, and instead extends the existing sphere class to include this functionality.

Note that in TheRestOfYourLife, the PDF functions ignore the moving aspect, and assume a static sphere (which preserves existing behavior).

See issue #1125

This is the code portion only, shared for feedback.


trevordblack (approved):
I think I'm okay with this.

It's not a complete refactoring with full support for time and SRTs.

But, it's not a massive increase in understanding for the user,
AND it's just objectively better than what exists at the moment.

I think the only potential gotcha is that previously a "moving sphere" represented a completely different geometric object than a regular sphere. Here, there is no true distinction between the two. A "moving sphere" is just a sphere.

This is definitely a change in mental model. I think it's an improvement. But it's a change in mental model and I might need a few days to simmer to make sure.

Feel free to ship, but I _may_ come back with concerns

hollasch:
Final work is complete in the `moving-sphere` branch. This will be rebased once the `classes-cleanup` PR is approved and merged.

Issue #1076, opened 2022-09-28
Book 1 Chapter 13.2 `dev-major`: `Generating Sample Rays`


hollasch:
### Discussed in https://github.com/RayTracing/raytracing.github.io/discussions/1075

<div type='discussions-op-text'>

<sup>Originally posted by **LollipopFt** September 25, 2022</sup>
https://github.com/RayTracing/raytracing.github.io/blob/dd6660d0f0309c178618695e78c929c1ddcd65f9/books/RayTracingInOneWeekend.html#L3214
was changed to this:
https://github.com/RayTracing/raytracing.github.io/blob/dd6660d0f0309c178618695e78c929c1ddcd65f9/books/RayTracingInOneWeekend.html#L3387-L3390

this flipped the image on the vertical axis such that it is upside down. changing the code to
```diff
            return ray(
                origin + offset,
-                lower_left_corner + s*horizontal + t*vertical - origin - offset
+                lower_left_corner + s*horizontal + (1-t)*vertical - origin - offset
            );
```
fixed it.
is it just me?</div>

hollasch:
Fixed in #1144

Issue #1042, opened 2022-08-05
camera::get_ray() function should perform all sample jittering


hollasch:
Currently the `camera::get_ray()` function performs depth-of-field lens jittering, and the `scene::render()` loops perform the pixel sample jittering. Both of these should be done in the `camera::get_ray()` function.

Issue #858, opened 2021-02-07
Improve `camera.h` naming, commentary


hollasch:
The `camera` class is moderately complex, with lots of member and intermediate variables. Member variables `u, v, w` could use a comment indicating what they are and that they're unit length, and some of the intermediate expressions like `lower_left_corner + s*horizontal + t*vertical - origin - offset` could be either broken into named descriptive components or at least given a descriptive name.

In addition to this general pass, it may be appropriate to split this into one specific to book1, and duplicates specific to books 2 & 3.

trevordblack:
The camera class has member variables `s`/`t`
the main loop has camera coordinates `u`/`v`

From what I can remember in other literature, `s`/`t` are typically used for pixel coordinates and `u`/`v` are used for texture coordinates.

We use `u`/`v`for the member variables of textures in book 2.

Maybe we should consider renaming the main loop variables to `s`/`t`?

trevordblack:
Can I get an update on this?

hollasch:
I'll move to `s`/`t` where possible and unique.

hollasch:
Fixed in code in PR #1154 (text updates pending).

Issue #546, opened 2020-05-16
Confusion surrounding use of the word "aperture"


trevordblack:
So, I have a question. How do we want to deal with "aperture"?

In the book as is, the aperture is the diameter of the lens element (the virtual one, anyway).
In photography, the "aperture stop" or "f-stop" is (focal_length/entrance_pupil) (Where the entrance_pupil is the diameter of the lens element as perceived when looking through the lens. This is not the same as just the diameter of the aperture).
Unfortunately, the aperture-stop is basically never referred to as the "aperture-stop", and is colloquially shortened to just "aperture".

And so we have an overloading of a term here.

Photographers' use of the word "aperture" here is (((and I'm a photographer!!!)))... Stupid. The term "aperture" correlates with an inverse of the lens aperture ::face palm::
Whereas, we use the word correctly, we're using the word "aperture" to mean... the diameter of the aperture. 

So, for photographers, an "increase" in "aperture" is associated with lesser defocus blur.
Whereas in the book an "increase" in "aperture" is associated with greater defocus blur.

(So, for photographers, an "increase" in "aperture" is associated with greater depth-of-field.
Whereas in the book an "increase" in "aperture" is associated with lesser depth-of-field.)

And, then, there is the matter of the utility of an f-stop. Any decent photographer will be able to map what a specific f-stop looks like.
By scaling with focal length, the defocus blur of any lens will be largely determined by f-stop. All f1.4 lens will have similar defocus blur (same w f2.0 and f2.8, and ....)

And, ultimately, most photographers have a patchwork of wrong understandings on why these things work the way they do (see entrance_pupil above). So it wouldn't hurt to tread lightly here.
I don't know if it's a good idea to have big rewrites for any photographers reading the books (and this may have only collided for me because of my exp in photography) but if a reader wants to continue down the path of filmic RT, then they're going to have to eventually understand f-stop and focal length as defined by photography (a good understanding of _T-Stop_ might also be good). I could probably write a ~500 word appendix converting the lens to 35mm sensor equivalent.

hollasch:
I go back and forth with formalizing our rendering setup through a `camera` class (and possibly a `renderer` class). In the end, I keep abandoning this as it feels like a net increase in complexity, and because we're aiming to illustrate a minimal -- rather than good -- raytracer. _Leaving some meat on the bone_, as it were.

[...]

PR #1132, opened 2023-05-31
Move image width & aspect ratio from scene to cam

hollasch:
This is a draft PR before I dive into the associated text changes.

--

This change does several things. Primarily, it's a refinement of the division between the scene class and the camera class.

The scene class holds the world description (which includes all scene geometry and lighting) and the camera. It is responsible for using the camera to interrogate the world, using the main render loop (all samples for all pixels).

The camera class is responsible for all 2D rendered image parameters, and uses these to generate rays one at a time.

The net effect of these changes are:

1. The image_width and aspect_ratio member variables move from the scene class to the camera class.

2. camera::initialize() no longer needs to take the aspect ratio argument.

3. camera::get_image_height() is a new public function that returns the image height, computed from the image width and the desired aspect ratio.

4. `camera::aperture` is now `camera::defocus_diameter`. `camera::lens_radius` is now `camera::defocus_radius`.

5. The scene and camera parameter assignments have been reordered to logical groupings based on the updated mental model.

6. In TheNextWeek/main.cc, we assigned background color, camera up, and focus distance before handing the scene off to the scene generators. There's been at least one case where a reader was confused about the state of the scene object because values were set in two different locations. This saved repeating some lines of code, but the simplicity of assigning everything in one place for each scene is better.

These two changes solidify the responsibilities of the two classes in preparation for future changes to the camera class, addressing the following issues:

- #546 Confusion surrounding use of the word "aperture"
- #682 Camera UV coordinates - handle images with width or height of one
- #858 Improve camera.h naming, commentary
- #1042 camera::get_ray() function should perform all sample jittering
- #1076 Book 1 Chapter 13.2 dev-major: Generating Sample Rays

One question left is that now that everything's assigned in the scene generator functions, why not just have them allocate and return the generated scene? I've decided to leave the current approach of mutating the scene -- passed by reference -- purely because it's simpler code. In my opinion, there's not a whole lot of advantage to the returned value approach.

### DIFF
--- a/src/InOneWeekend/main.cc
+++ b/src/InOneWeekend/main.cc
@@ -20,16 +20,19 @@
 
 
 void random_spheres(scene& scene_desc) {
-    scene_desc.image_width       = 1200;
-    scene_desc.aspect_ratio      = 16.0 / 9.0;
-    scene_desc.samples_per_pixel = 10;
+    scene_desc.cam.lookfrom = vec3(13,2,3);
+    scene_desc.cam.lookat   = vec3(0,0,0);
+    scene_desc.cam.vup      = vec3(0,1,0);
+    scene_desc.cam.vfov     = 20;
 
-    scene_desc.cam.vfov       = 20;
-    scene_desc.cam.focus_dist = 10.0;
-    scene_desc.cam.aperture   = 0.1;
-    scene_desc.cam.lookfrom   = vec3(13,2,3);
-    scene_desc.cam.lookat     = vec3(0,0,0);
-    scene_desc.cam.vup        = vec3(0,1,0);
+    scene_desc.cam.aspect_ratio = 16.0 / 9.0;
+    scene_desc.cam.image_width  = 1200;
+
+    scene_desc.cam.defocus_diameter = 0.1;
+    scene_desc.cam.focus_dist       = 10;
+
+    scene_desc.samples_per_pixel = 10;
+    scene_desc.max_depth         = 20;
 
     hittable_list& world = scene_desc.world;
 
--- a/src/common/camera.h
+++ b/src/common/camera.h
@@ -16,11 +16,14 @@
 
 class camera {
   public:
-    void initialize(double aspect_ratio = 1.0) {
+    void initialize() {
         auto theta = degrees_to_radians(vfov);
         auto h = tan(theta/2);
         auto viewport_height = 2.0 * h;
-        auto viewport_width = aspect_ratio * viewport_height;
+
+        // Calculate the actual aspect ratio of the integer image width and height.
+        auto actual_aspect_ratio = static_cast<double>(image_width) / get_image_height();
+        auto viewport_width = actual_aspect_ratio * viewport_height;
 
         w = unit_vector(lookfrom - lookat);
         u = unit_vector(cross(vup, w));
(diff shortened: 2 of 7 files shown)


src/InOneWeekend/main.cc
```diff
@@ -20,16 +20,19 @@
...
-    scene_desc.cam.vfov       = 20;
-    scene_desc.cam.focus_dist = 10.0;
-    scene_desc.cam.aperture   = 0.1;
-    scene_desc.cam.lookfrom   = vec3(13,2,3);
-    scene_desc.cam.lookat     = vec3(0,0,0);
-    scene_desc.cam.vup        = vec3(0,1,0);
+    scene_desc.cam.defocus_diameter = 0.1;
+    scene_desc.cam.focus_dist       = 10;
+
+    scene_desc.samples_per_pixel = 10;
```
trevordblack, line 34: My mental model has samples_per_pixel in camera rather than scene, but I'm happy with either

trevordblack: Hmm. 

I'm trying to think what be the more consistent mental model for max_depth would be.

If we moved samples_per_pixel over to camera would it make sense to keep max_depth in the scene class or in the camera class....

The number of bounces is pretty much solely dependent on geo+lighting, which is very much contained in scene. But it _feels_ like it would be coupled with samples_per_pixel.

I'm fine either way. Just things to consider

src/common/camera.h
```diff
@@ -16,11 +16,14 @@
...
   public:
-    void initialize(double aspect_ratio = 1.0) {
+    void initialize() {
         auto theta = degrees_to_radians(vfov);
         auto h = tan(theta/2);
         auto viewport_height = 2.0 * h;
-        auto viewport_width = aspect_ratio * viewport_height;
+
+        // Calculate the actual aspect ratio of the integer image width and height.
+        auto actual_aspect_ratio = static_cast<double>(image_width) / get_image_height();
```
trevordblack, line 25: Not sure how I feel about this.

You're correct that the image_height could be rounded in such a way that the "actual" aspect ratio is different from the parameter

But this looks really messy.

I think it's fine because this is only called once, and it's to solve a possible technical round-off error

I would like this to be neatly documented to the associated text (which I assume you already plan to write).

Do you think that the roundoff error possibility should be added into a comment?

trevordblack (approved):
Some small nitpicks, but I'm otherwise really happy with this change (and this direction)

hollasch:
Ugh. Such simple comments, but you've triggered a lot of rethink for me over the past several days. My knee-jerk response was going to be "yeah, it's a bit odd, but that's how it works out". Then a longer response about how the camera isn't a great metaphor, and the `camera` class actually is responsible for the production of a single randomized pixel sample, and the scene is responsible for the primary render loop over the entire image. Then I kind of realized that's its only real justification, besides the extraordinarily weak value-add of holding a camera and the world geometry. Then realizing that it's a bit cruddy that the client has to reach into the scene to get the camera to tweak its parameters. Finally, that bubbled up enough for me to question the way this is all used at the main program.

Sigh.

I think we (I) should kill the `scene` class. I sketched out what things would look like from the top if instead of functions that construct a `scene` so that the `main` function could then just trivially call the `scene::render()` function, we just dispatch to functions that render the desired scene, using a `camera` and the world geometry. And it's just a lot simpler and cleaner. Which makes be feel a bit foolish, but that's where it is.

hollasch:
See #1133

PR #1133, opened 2023-06-03
SKETCH: eliminate the scene class

hollasch:
The long and winding road.

### DIFF
--- a/src/TheNextWeek/main.cc
+++ b/src/TheNextWeek/main.cc
@@ -25,18 +25,21 @@
 
 
 void random_spheres(scene& scene_desc) {
-    scene_desc.aspect_ratio      = 16.0 / 9.0;
-    scene_desc.image_width       = 400;
-    scene_desc.samples_per_pixel = 100;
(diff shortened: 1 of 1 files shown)


trevordblack:
Do we want to encapsulate all of the stuff in a render?

I think there's value in having a scene class, as it reinforces the idea that all of the geo+lighting+camera(?) Is packaged together and neatly reinforces the idea that a render should be treated like an artificial thing. Like a scene on Broadway or in Hollywood.

The previous encapsulation is kind of thin though. It does sort of act like an immutable struct after construction (and some initial parameter setting). That you have to reach into struct to change the camera is kind of not great

We've wittled down the mental model and structure of every other part of the code to a place that we're both pretty happy with. This part of the code is just a fountain for slightly misaligned mental models

Which, I mean, makes sense as this is part of most rendering engines that is the most poorly defined. There is a lot of play in terms of how you want a camera and scene represented. Whereas there is shockingly few in terms of the actual assets

PR #1138, opened 2023-06-06
Moving sphere

hollasch:


trevordblack:
I don't have the time to go through and check the body of book 2 for flow and if the english does a good job of explaining the mental model

BUT
I'm happy with the changes here.

And I'm going to be checking for flow and what not in book 2 eventually, so that's an issue for another day.

Approved.

trevordblack (approved):

PR #1147, opened 2023-06-23
Fix render ray construction in text

### LINKED ISSUE #1094: Render ray construction is inconsistent
The code in listing 24 in the book is as follows.

```
ray r(origin, lower_left_corner + u*horizontal + v*vertical);
```

But from the previous ray generation implementation (Listing 9) it should still subtract the origin coordinates, even if at this stage the origin is still (0, 0, 0).

```
ray r(origin, lower_left_corner + u*horizontal + v*vertical - origin);
```

The book goes back to subtracting the origin (Listing 27) when implementing the camera class in order to get the accurate ray if the camera is to move.

hollasch:
Resolves #1094


trevordblack (approved):
Ship as is

PR #1148, opened 2023-06-23
Change camera::get_ray() function to int params

hollasch:
Text changes to follow after this is finished.

### DIFF
--- a/src/InOneWeekend/camera.h
+++ b/src/InOneWeekend/camera.h
@@ -47,9 +47,7 @@ class camera {
             for (int i = 0; i < image_width; ++i) {
                 color pixel_color(0,0,0);
                 for (int sample = 0; sample < samples_per_pixel; ++sample) {
-                    auto s = (i + random_double()) / (image_width-1);
-                    auto t = (j + random_double()) / (image_height-1);
-                    ray r = get_ray(s, t);
+                    ray r = get_ray(i, j);
                     pixel_color += ray_color(r, max_depth, world);
                 }
                 write_color(std::cout, pixel_color, samples_per_pixel);
@@ -60,6 +58,7 @@ class camera {
     }
 
   private:
+    int image_height;
     point3 origin;
     vec3 horizontal;
     vec3 vertical;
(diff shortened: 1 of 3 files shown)


src/InOneWeekend/camera.h
```diff
@@ -60,6 +56,7 @@ class camera {
     }
 
   private:
+    int image_height;
```
trevordblack, line 59: it's kind of weird having image_width be public and image_height be private?

Consider having image_height be public and just solve for any of the image_height, image_width, aspect ratio that are not explicitly defined? ...

No. That doesn't really work. If they're all public it's hard to know which were explicitly set unless we create constructors. But that's not a fantastic idea if we let main set aspect_ratio.

What was the decision behind having aspect ratio and width be set by main? I don't remember.

We shouldn't do getters/setters (we decided no a while ago).

WDYT?

src/InOneWeekend/camera.h
```diff
@@ -85,10 +84,9 @@ class camera {
         lens_radius = aperture / 2;
     }
 
-    ray get_ray(double s, double t) const {
-        // Return the ray from the projection point to the indicated pixel. Coordinates s,t are
-        // the normalized image-based coordinates of the pixel. Image left is s=0, image right
-        // is s=1, image top is t=0, image bottom is t=1.
```
trevordblack, line 91: What was the decision behind removing the comments?

trevordblack (approved):
The fundamental change here (_i.e._ get_ray() to ints) is _fine_.

I'm not entirely sure the motivating factor behind the change. If I had to guess it's easier to express the move from 1spp to multiple with this change? If that's the case then yes, I think this is probably valuable. Not entirely certain though. If you think this is worth it then feel free to ship.

With that said, the private image_height thing has me scratching my head. Maybe we should just set width and height in main and calculate aspect? What was the reason for the way it's set up now? Goddamn we've been on this too long.

src/InOneWeekend/camera.h
hollasch: Three variables, two unknowns. The main idea is that while developing, people will typically alter parameters to adjust speed of rendering, image dimensions being one aspect. Once the render aspect ratio is done, it's typically fixed. So they can just adjust a single intuitive variable (width) or two variables plus at least a mental calculation. The former is a nicer interface. Also, if you want to compare renders without regard to size, then you want a fixed aspect ratio.

src/InOneWeekend/camera.h
hollasch: This is one of several increments to follow, so if there will be a comment, it will come after the dust settles.

PR #1152, opened 2023-06-26
Add clarification about unit-length normal vectors

### LINKED ISSUE #1062: Resolve whether hit_record::set_face_normal expects unit vectors or not
If not, then the color computation in early versions of `ray_color()` need to ensure that the normal vector is unit length.

Ref discussion #1060.

hollasch:
Resolves #1062

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -1046,9 +1046,10 @@ literature.
 
 We add the `front_face` bool to the `hit_record` class.
 We'll also add a function to solve this calculation for us: `set_face_normal()`.
-Note that the vector passed to the new `set_face_normal` function is expected to be of unit length.
-Of course, we could always normalize the parameter explicitly, but it's more efficient if the
-  geometry code does this, as it's usually easier when you know more about the specific geometry.
+For convenience we will assume that the vector passed to the new `set_face_normal()` function is of
+  unit length.
+We could always normalize the parameter explicitly, but it's more efficient if the geometry code
+  does this, as it's usually easier when you know more about the specific geometry.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class hit_record {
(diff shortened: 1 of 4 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -1044,8 +1044,11 @@
 at geometry time. This is simply a matter of preference, and you'll see both implementations in the
 literature.
 
-We add the `front_face` bool to the `hit_record` class. We'll also add a function to solve this
-calculation for us.
+We add the `front_face` bool to the `hit_record` class.
+We'll also add a function to solve this calculation for us: `set_face_normal()`.
+Note that the vector passed to the new `set_face_normal` function is expected to be of unit length.
+Of course, we could always normalize the parameter explicitly, but it's more efficient if the
+  geometry code does this, as it's usually easier when you know more about the specific geometry.
```
trevordblack, line 1051: "Note that the vector passed to the new `set_face_normal` function is expected to be of unit length. Of course, we could always normalize the parameter explicitly, but it's more efficient if the geometry code does this, as it's usually easier when you know more about the specific geometry."
-> something like
"For convenience we will assume that the vector passed to the new `set_face_normal` function is of unit length. We could always normalize the parameter within the body of the `set_face_normal` function, ...

books/RayTracingInOneWeekend.html
```diff
@@ -1057,6 +1060,9 @@
         bool front_face;
 
         void set_face_normal(const ray& r, const vec3& outward_normal) {
+            // Sets the hit record normal vector.
+            // NOTE: the parameter `outward_normal` must be of unit length.
```
trevordblack, line 1064: Replace "must" with "is assumed to"

There is no code enforcement, a reader might read this and assume from MUST that there is some code of enforcement (e.g. their code won't compile, it'll crash).

We have no guarantees that a non-unit length normal will be problematic, and so I don't think MUST is appropriate

trevordblack (changes requested):
Please consider changes requested

books/RayTracingInOneWeekend.html
hollasch: I like it.

hollasch:
Took all suggestions. Will squash and merge.

PR #1154, opened 2023-06-29
Camera v2 code

hollasch:
- Switch from aperture to defocus angle
- Refactored pixel sample functions
- Add disk pixel sample function
- Updated camera components & construction
- Handle images with width/height of 1
- Clamp image width/height to minimum of 1

### DIFF
--- a/src/InOneWeekend/camera.h
+++ b/src/InOneWeekend/camera.h
@@ -22,18 +22,18 @@
 
 class camera {
   public:
-    double aspect_ratio      = 1.0;
-    int    image_width       = 100;
-    int    samples_per_pixel = 10;
-    int    max_depth         = 10;
+    double aspect_ratio      = 1.0;  // Ratio of image width over height
+    int    image_width       = 100;  // Rendered image width in pixel count
+    int    samples_per_pixel = 10;   // Count of random samples for each pixel
+    int    max_depth         = 10;   // Maximum number of ray bounces into scene
 
-    double vfov     = 40;
-    point3 lookfrom = point3(0,0,-1);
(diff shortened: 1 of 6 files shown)


src/InOneWeekend/camera.h
```diff
@@ -56,54 +56,90 @@ class camera {
...
+    vec3   pixel_delta_u;   // Offset to pixel to the right
+    vec3   pixel_delta_v;   // Offset to pixel below
+    vec3   viewport_u;      // Vector across viewport horizontal edge
+    vec3   viewport_v;      // Vector down viewport vertical edge
+    vec3   u, v, w;         // Camera frame basis vectors
+    vec3   defocus_disk_u;  // Defocus disk horizontal radius
+    vec3   defocus_disk_v;  // Defocus disk vertical radius
 
     void initialize() {
+        image_width  = (image_width < 1) ? 1 : image_width;
```
trevordblack, line 71: This feels unnecessary?

trevordblack: If I understand correctly that this is to stop the person from set the width to 0 or lower, but that feels like the kind of thing that the user should just not do? And if they accidentally do it, they should be able to debug

trevordblack (approved):

src/InOneWeekend/camera.h
hollasch: We must do this for `image_height`, since a width of 1 may yield a height of 0 for some aspect ratios. I put this here for symmetry, but you're right that it's strictly unnecessary given our current style of "here's a gun, there's your foot, have a great day".

I'll pull this line.

Issue #1071, opened 2022-09-01
Book 1: Camera terms used without being introduced


kristopolous:
There's a number of things that for me, need to be cleared up:

> ...For now we'll use a 16:9 aspect ratio, since that's so common.

> In addition to setting up the pixel dimensions for the rendered image, we also need to set up a virtual viewport through which to pass our scene rays. For the standard square pixel spacing, the viewport's aspect ratio should be the same as our rendered image. We'll just pick a viewport two units in height. 

This suggests a width of 16/9 * 2 units, right?

> We'll also set the distance between the projection plane and the projection point to be one unit. 

"Projection plane" and "projection point" aren't defined here. They are technical terms with functional meanings. This is the first time they are introduced.

> This is referred to as the “focal length”, not to be confused with “focus distance”, which we'll present later.

Without a meaningful functional definition of "projection plane" and "projection point", the term "focal length" and what "distance" means in this context is unknown and opaque. It is effectively the same as this:

> We'll also set the ████ between the ████ and the ████ to be one unit. This is referred to as the “focal length”, not to be confused with “focus distance”, which we'll present later.

ShijianXu:
I agree. I took a lot of time to figure out what the `viewport` mean. I think some terms should be explained in more details. Like why we set the viewport size to be `2x3.5555` or actually any size that keeps the ratio `9:16` makes sense.

hollasch:
We'll roll this into the new camera updates.

Issue #464, opened 2020-04-11
InOneWeekend: clarify Defocus Blur


shaunplee:
I'm having trouble developing an intuition on how defocus blur works.

In trying to understand the chapter, I drew, for myself, a slightly modified version of Figure 17 to add an "out-of-focus plane":

![Limnu_2020_4_11](https://user-images.githubusercontent.com/1076792/79056888-f5e4fd00-7c0f-11ea-9d37-d912b72e22eb.png)

        ray get_ray(double s, double t) {
            vec3 rd = lens_radius * random_in_unit_disk();
            vec3 offset = u * rd.x() + v * rd.y();

            return ray(
                origin + offset,
                lower_left_corner + s*horizontal + t*vertical - origin - offset
            );

My understanding is that the modification of `get_ray` from Listing 63 to be a ray to `lower_left_corner + s*horizontal + t*vertical - origin - offset` causes the rays from `origin + offset` to converge at the same point because the `- offset` translates the focus plane to compensate for the random `offset` at the lens, but this compensation is only perfect at the focus plane. Based on the shape of the rays, at other distances, color samples are taken from different parts of the scene, and the rays are farther apart the farther you are from the focus plane. A smaller aperture results in a tighter set of rays and therefore less blur.

However, I'm not sure if my understanding is correct.

My suggestion is to add some additional explanation of what is happening at the focus plane.

I also note that the aperture is specified in the more natural units of a diameter, and not as an _f_-stop. I experimented with setting the aperture to 16.0, unthinkingly assuming it would make everything sharp and got an extremely blurry image. I probably made the assumption because Listing 64 suggests "Using a big aperture" and sets the aperture to "2.0" and, as it happens, _f_/2.0 is a pretty large aperture. I don't know if other readers will make the same assumption.

dafhi:
> A smaller aperture results in a tighter set of rays and therefore less blur.

looks like it contradicts your last paragraph.  I'm not necessarily "love," so feel free to rip me a new one

shaunplee:
It's very likely I'm being self-contradictory, as I'm pretty confused about this whole thing. :)

A smaller aperture resulting in less blur is in line with my experience with using a camera, in that higher _f_-stops give me less blur (more depth of field).

The last paragraph from my original post seems to me to be consistent with that. 2.0 is a "large" aperture. Listing 65 for creating the "final image" uses a more reasonable aperture of 0.1, which still causes the defocus blur effect to be apparent in the image.

[...]

PR #1158, opened 2023-07-12
Text changes to all three books for camera updates

hollasch:
An ugly monster of a PR. Probably best to pull this down and review locally.

This should also address #546, #1071, and #464.


trevordblack:
"and this is what is often called a half-line or ray."
>>
"and this is what is often called a half-line or a ray."

trevordblack:
"that I call ray::at(t):"
>>
"that we'll call ray::at(t):"

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,93 @@
 
 Sending Rays Into the Scene
 ----------------------------
-Now we are ready to turn the corner and make a ray tracer. At its core, a ray tracer sends rays
-through pixels and computes the color seen in the direction of those rays. The involved steps are
+Now we are ready to turn the corner and make a ray tracer.
+At its core, a ray tracer sends rays through pixels and computes the color seen in the direction of
+  those rays.
```
trevordblack, line 507: ?

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,93 @@
...
 When first developing a ray tracer, I always do a simple camera for getting the code up and running.
 I also make a simple `ray_color(ray)` function that returns the color of the background (a simple
-gradient).
+  gradient).
 
 I’ve often gotten into trouble using square images for debugging because I transpose $x$ and $y$ too
-often, so I’ll use a non-square image.
+  often, so I’ll use a non-square image.
 For now we'll use a 16:9 aspect ratio, since that's so common.
+A 16:9 aspect ratio means that the image width over image height = 16/9.
```
trevordblack, line 521: "A square image has a 1:1 aspect ratio, because it's width is the same as its height. Since we don't want to use a square image, we'll need to choose a different aspect ratio. We'll choose 16:9 because it's so common. A 16:9 aspect ratio means that the ratio of image width to image height is 16:9. As another example, an image 800 pixels wide by 400 pixels high has a 2:1 aspect ratio"


Consider using latex shenanigans for 16:9 and other aspect ratios

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,93 @@
...
+For example, as an image 800 pixels wide by 400 pixels high has a 2:1 aspect ratio.
+
+We will drive the image (and hence viewport) size from the desired aspect ratio and the image width.
+That way, you can increase or decrease your image size by changing a single value, instead of doing
+  a calculation and changing two values, which you'd have to do if you had to specify both the image
+  width and image height for some target aspect ratio.
+Also, in calculating the image height from the image width and the desired aspect ratio, we'll need
+  to ensure that the resulting height is at least 1.
```
trevordblack, line 529: "The image's aspect ratio can be determined from the ratio of its height to its width. But since we have a given aspect ratio in mind, it's easier to set the image's width and the aspect ratio, and then using this to calculate for its height. This way, we can scale up or down the image by changing the image width, and it won't throw off our desired aspect ratio. We do have to make sure that when we solve for the image height the resulting height is at least 1."

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,93 @@
...
 
 In addition to setting up the pixel dimensions for the rendered image, we also need to set up a
-virtual viewport through which to pass our scene rays. For the standard pixel spacing (equally
-spaced in both the horizontal and vertical direction), the viewport's aspect ratio should be the
-same as our rendered image. We'll just pick a viewport two units in height. We'll also set the
-distance between the projection plane and the projection point to be one unit. This is referred to
-as the _focal length_, not to be confused with the _focus distance_, which we'll present later.
+  virtual _viewport_ through which to pass our scene rays.
+The viewport is a rectangle in the 3D world that contains the grid of image pixel locations.
```
trevordblack, line 533: virtual rectangle

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,93 @@
...
+For the standard pixel spacing (equally spaced in both the horizontal and vertical direction), the
+  viewport will have the same aspect ratio as the rendered image.
```
trevordblack, line 535: If the width is the same as the height of a pixel, the viewport will have the same aspect ratio as the rendered image. The distance between two adjacent pixels is called the pixel spacing, and square pixels is the standard.

trevordblack:
Lots of red ink.

Feel free to liberally interpret my comments

trevordblack:
"To start things off, we'll arbitrarily choose a viewport height of 2.0,"
>>
"To start things off, we'll choose an arbitrary viewport height of 2.0,"

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,94 @@
...
+
+To start things off, we'll arbitrarily choose a viewport height of 2.0, and scale the viewport width
```
trevordblack, line 537: we'll choose an arbitrary viewport height of 2.0

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,94 @@
...
+  to give us the desired aspect ratio.
```
trevordblack, line 538: we'll choose an arbitrary viewport height of 2.0

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,94 @@
...
+Here's a sketch of what this code will look like:
```
trevordblack, line 539: Don't use sketch. Conjures image of drawing.

Go with "example" or "snippet" or your choice

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,94 @@
...
+
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+    auto aspect_ratio = 16.0 / 9.0;
+    int image_width = 400;
+
+    // Calculate the image height, and ensure that it's at least 1.
+    int image_height = static_cast<int>(image_width / aspect_ratio);
+    image_height = (image_height < 1) ? 1 : image_height;
+
+    auto viewport_height = 2.0;
```
trevordblack, line 549: Consider adding a comment here.:

"// Calculate the viewport height, we don't need to check for 1 because it's a double.

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,94 @@
...
+    auto viewport_width = viewport_height * (static_cast<double>(image_width)/image_height);
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [image-setup]: Rendered image setup]
+
+(If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
+because the _actual_ aspect ratio is the integer-based ratio of image width over image height.)
```
trevordblack, line 555: ... and the image width was rounded down to the nearest integer, rather than a perfect fraction

books/RayTracingInOneWeekend.html
hollasch: ? Do you mean "use the Unicode ratio glyph"? ( `:` vs `∶`)?

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,94 @@
...
+
+Next we will define a _projection point_ in 3D space, from which all scene rays will originate (this
```
trevordblack, line 557: In the literature, this is called the "focal point":
https://en.wikipedia.org/wiki/Focal_point
https://en.wikipedia.org/wiki/Focus_(optics)
https://en.wikipedia.org/wiki/Thin_lens

The projection point is the point on a plane nearest to a point.

I'm not sure if projection point is wrong, but it's not what I would go with.

YES, I am aware that focal point is confusing with focal plane, which is a different thing:
https://en.wikipedia.org/wiki/Cardinal_point_(optics)#Focal_planes

Sorry

books/RayTracingInOneWeekend.html
hollasch: How about this?

> If pixels are spaced the same distance horizontally as they are vertically, the viewport will have the same aspect ratio as the rendered image.
> The distance between two adjacent pixels is called the pixel spacing, and square pixels is the standard.

books/RayTracingInOneWeekend.html
hollasch: Can you elaborate?

books/RayTracingInOneWeekend.html
hollasch: I like it.

books/RayTracingInOneWeekend.html
hollasch: Better:
> If pixels are spaced the same distance horizontally as they are vertically, the viewport will have the same aspect ratio as the rendered image.
> The distance between two adjacent pixels is called the pixel spacing, and square pixels is the standard.

books/RayTracingInOneWeekend.html
hollasch: Done.

books/RayTracingInOneWeekend.html
hollasch: Did this:
```
    // Calculate the viewport height. Since it's real valued, we don't check for values less then 1.
```

books/RayTracingInOneWeekend.html
hollasch: Updated to
> (If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
because the _actual_ aspect ratio is the integer-based ratio of image width over image height, which
will yield zero for values less than one.)

books/RayTracingInOneWeekend.html
trevordblack: Why is there `  ` sprinkled about at the beginning of lines?

books/RayTracingInOneWeekend.html
trevordblack: Sorry to nitpick. I prefer
```
// Calculate the viewport height. We don't check for values less than 1 since it's real valued
```

No passive voice in comments

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+
+    // Calculate the viewport height. Since it's real valued, we don't check for values less then 1.
+    auto viewport_height = 2.0;
+    auto viewport_width = viewport_height * (static_cast<double>(image_width)/image_height);
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [image-setup]: Rendered image setup]
+
+(If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
+because the _actual_ aspect ratio is the integer-based ratio of image width over image height, which
+will yield zero for values less than one.)
```
trevordblack, line 564: Are you sure that's correct?

My understanding is that the reason we don't use the `aspect_ratio` is because it's just a user suggestion, and what really matters is the actual ratio of the image dimensions. I don't think that's quite what you're saying here?

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+
+I’ll put the camera center at $(0,0,0)$.
 I will have the y-axis go up, and the x-axis to the right.
-In order to respect the convention of a right handed coordinate system, into the screen is the
-negative z-axis.
+Since we use a right-handed coordinate system, the vector into the screen is the negative z-axis.
+Note that I do not make the ray direction a unit length vector because I think not doing that makes
+  for simpler and slightly faster code.
+
+  ![Figure [camera-geom]: Camera geometry](../images/fig-1.03-cam-geom.jpg)
```
trevordblack, line 578: Two requests.

1. Can you have a right angle symbol at the intersection of the vector from camera center to  viewport
2. Can you either
  a. Add an intermediate figure between 3 and 4 which is an askew viewport with a grid of pixels on it OR
   b. Add a grid of pixels to the viewport in 3

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+    auto viewport_width = viewport_height * (static_cast<double>(image_width)/image_height);
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    [Listing [image-setup]: Rendered image setup]
+
+(If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
+because the _actual_ aspect ratio is the integer-based ratio of image width over image height, which
+will yield zero for values less than one.)
 
-I’ll put the “eye” (or camera center if you think of a camera) at $(0,0,0)$.
+Next we will define the camera center: a point in 3D space from which all scene rays will originate
```
trevordblack, line 566: the _camera center_ ?

trevordblack:
Most of the `I`s and `I'll`s should be `We`s and `We'll`s

If it's opinion -> I
If it's something that we're going to do -> We

Should I spin up a separate issue for this?

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+  (this is also commonly referred to as the _eye point_).
+The vector from the camera center to the viewport center will be orthogonal to the viewport.
+We'll initially set the distance between the viewport and the camera center point to be one unit.
+This distance is often referred to as the _focal length_.
+
+I’ll put the camera center at $(0,0,0)$.
```
trevordblack, line 572: "For simplicity we'll start with the camera center at $(0,0,0)$"

books/RayTracingInOneWeekend.html
hollasch: Yeah, it hurts my head when I try to parse this. Which doesn't really matter because it's completely bogus. Let me take another swipe.

books/RayTracingInOneWeekend.html
hollasch: Ok.

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
 I will have the y-axis go up, and the x-axis to the right.
-In order to respect the convention of a right handed coordinate system, into the screen is the
-negative z-axis.
+Since we use a right-handed coordinate system, the vector into the screen is the negative z-axis.
```
trevordblack, line 574: Couple of things

1. I don't think that we need to explain why we're doing y up versus z up, but it's something to consider explaining
2. This is the first time we're mentioned that we're using a right-handed coordinate system. There was no previous mention of what that means, nor was there a previous mention that we're doing this. This reads kind of confusingly. "Since we are doing X"... wait we are? Since when. You need to rephrase this to state that we are in right-handed coordinate systems, and we should also probably say the choice is an arbitrary one

books/RayTracingInOneWeekend.html
hollasch: Got it.

books/RayTracingInOneWeekend.html
hollasch: I was trying to continue the tone of the original, but it's definitely all over the map throughout the book now.

books/RayTracingInOneWeekend.html
hollasch: I would stay far away from the Y-up/Z-up debate. Just tell them what our convention is and move along.

Will reword with respect to RH coordinates.

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+Note that I do not make the ray direction a unit length vector because I think not doing that makes
+  for simpler and slightly faster code.
```
trevordblack, line 576: It's early to state this. We haven't seen any code yet

hollasch:
This is going to be _such_ merge heck.

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
-$s=0$ is image left, $s=1$ is image right, $t=0$ is image top, and $t=1$ is image bottom.
-That means that we get our vertical axis flip at the right point by computing $t = 1 - y$, where $y$
-is the vertical component of our 3D coordinate system, and $t$ is the flipped vertical component of
-our image coordinate system.
-
-I will traverse the screen from the upper left hand corner, and use two offset vectors along the
-screen sides to move the ray endpoint across the screen.
-Note that I do not make the ray direction a unit length vector because I think not doing that makes
-for simpler and slightly faster code.
+  where the Y axis is inverted (increasing going down).
```
trevordblack, line 582: "Now the inevitable tricky part. While our 3D space has the conventions above, this conflicts with our image coordinates, where we want to print out the zeroth pixel in the top-left and work our way down to the last pixel at the bottom-right, meaning that our image coordinates are inverted: An increase in the Y axis goes down. "

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+
+As we scan our image, we will start at the upper left pixel (pixel 0,0), scan left-to-right across
+  the row, and then scan the next row, top-to-bottom (reading order for most Latin languages).
```
trevordblack, line 585: As we can our image, we will start at the upper-left pixel (pixel \math(0,0)) scan left-to-right across the row, and then scan one row after another from the top to the bottom

I would remove (reading order for most Latin languages)
I'm not sure it's helpful and it's off topic

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+To help navigate the pixel grid, we'll use a vector from the left edge to the right edge
+  (`viewport_u`), and a vector from the upper edge to the lower edge (`viewport_v`).
 
-  ![Figure [cam-geom]: Camera geometry](../images/fig-1.03-cam-geom.jpg)
+Our pixel grid will be inset from the viewport edges by half the pixel-to-pixel distance, so each
+  pixel will gather light from the area around it.
+Here's what our viewport and pixel grid look like:
+
+  ![Figure [pixel-grid]: Viewport and pixel grid](../images/fig-1.04-pixel-grid.jpg)
```
trevordblack, line 593: The figure should have the pixel annotated P_0,0

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
-I will traverse the screen from the upper left hand corner, and use two offset vectors along the
-screen sides to move the ray endpoint across the screen.
-Note that I do not make the ray direction a unit length vector because I think not doing that makes
-for simpler and slightly faster code.
+  where the Y axis is inverted (increasing going down).
+
+As we scan our image, we will start at the upper left pixel (pixel 0,0), scan left-to-right across
+  the row, and then scan the next row, top-to-bottom (reading order for most Latin languages).
+To help navigate the pixel grid, we'll use a vector from the left edge to the right edge
+  (`viewport_u`), and a vector from the upper edge to the lower edge (`viewport_v`).
```
trevordblack, line 587: have these be math instead of code

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
 
-  ![Figure [cam-geom]: Camera geometry](../images/fig-1.03-cam-geom.jpg)
+Our pixel grid will be inset from the viewport edges by half the pixel-to-pixel distance, so each
+  pixel will gather light from the area around it.
```
trevordblack, line 590: "Each Pixel will gather light from the area around it. If pixels were at the grid edge they would gather light from outside of the pixel grid. To avoid this, we will inset the pixels from the viewport edges by half the pixel-to-pixel distance"

trevordblack: Alternively:

""Each Pixel will gather light from the area around it. If pixels were at the grid edge they would gather light from outside of the pixel grid. To avoid this, we will inset the pixels from the viewport edges"

Unless you acknowledge that the sampling area is a pixel in width, saying the inset length may not be needed (or helpful)

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+Here's what our viewport and pixel grid look like:
+
+  ![Figure [pixel-grid]: Viewport and pixel grid](../images/fig-1.04-pixel-grid.jpg)
+
+In this figure, we have the viewport, the pixel grid for a 7×5 resolution image, the viewport upper
```
trevordblack, line 595: consider \cross instead of ×
if you haven't already (i assume you have, but...)

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+  left corner $\mathbf{Q}$, the pixel $\mathbf{P_{0,0}}$ location, the `viewport_u` vector
+  $\mathbf{V_u}$, the `viewport_v` vector $\mathbf{V_v}$, and the pixel delta vectors
+  $\mathbf{\Delta_u}$ and $\mathbf{\Delta_v}$.
```
trevordblack, line 598: $\mathbf{\Delta u}$ and $\mathbf{\Delta v}$.

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
-  ![Figure [cam-geom]: Camera geometry](../images/fig-1.03-cam-geom.jpg)
+Our pixel grid will be inset from the viewport edges by half the pixel-to-pixel distance, so each
+  pixel will gather light from the area around it.
+Here's what our viewport and pixel grid look like:
+
+  ![Figure [pixel-grid]: Viewport and pixel grid](../images/fig-1.04-pixel-grid.jpg)
+
+In this figure, we have the viewport, the pixel grid for a 7×5 resolution image, the viewport upper
+  left corner $\mathbf{Q}$, the pixel $\mathbf{P_{0,0}}$ location, the `viewport_u` vector
+  $\mathbf{V_u}$, the `viewport_v` vector $\mathbf{V_v}$, and the pixel delta vectors
```
trevordblack, line 597:
```
"  left corner $\mathbf{Q}$, the pixel $\mathbf{P_{0,0}}$ location, the viewport vector
  $\mathbf{V_u}$ (`viewport_v`), the viewport vector $\mathbf{V_v}$ (`viewport_u`), and the pixel delta vectors"
```

books/RayTracingInOneWeekend.html
```diff
@@ -577,18 +626,29 @@
...
+        image_height = (image_height < 1) ? 1 : image_height;
 
         // Camera
 
-        auto viewport_height = 2.0;
-        auto viewport_width = aspect_ratio * viewport_height;
         auto focal_length = 1.0;
+        auto viewport_height = 2.0;
+        auto viewport_width = viewport_height * (static_cast<double>(image_width)/image_height);
+        auto center = point3(0, 0, 0);  // Camera center
```
trevordblack, line 639: why not just name it `camera_center`?

books/RayTracingInOneWeekend.html
```diff
@@ -502,54 +502,103 @@
...
+At its core, a ray tracer sends rays through pixels and computes the color seen in the direction of
+  those rays.
+The involved steps are
 
-    1. Calculate the ray from the eye to the pixel,
+    1. Calculate the ray from the “eye” through the pixel,
     2. Determine which objects the ray intersects, and
     3. Compute a color for that intersection point.
 
 When first developing a ray tracer, I always do a simple camera for getting the code up and running.
```
trevordblack, line 514: Move line 514 down to line 679

books/RayTracingInOneWeekend.html
```diff
@@ -616,9 +677,9 @@
 
 <div class='together'>
 The `ray_color(ray)` function linearly blends white and blue depending on the height of the $y$
```
trevordblack, line 679: Don't have the `ray_color` function in the above code block. There's already a lot going on in that.

We also completely dropped the motivation for adding a `ray_color` since we just got inundated with camera stuff. Move line 514-515 to here, and then make a code block for `ray_color`

books/RayTracingInOneWeekend.html
hollasch: I'm authoring using semantic linefeeds. Wrapped sentences get indented. This makes it easier to diff text changes  without having to track whole-paragraph reflow. Unfortunately, the reflow issue happens once when going from prose style to semantic linefeeds.

books/RayTracingInOneWeekend.html
hollasch: Done.

books/RayTracingInOneWeekend.html
hollasch: I don't see any passive voice. It's a combination of command and active. Are you referring to the predicate "since it's real valued"?

Anyway, latest iteration is `// Viewport widths less than one are ok since they are real valued.`

books/RayTracingInOneWeekend.html
hollasch: Converted to "camera center", per Slack conversation.

books/RayTracingInOneWeekend.html
hollasch: Moved to follow the code listing.

books/RayTracingInOneWeekend.html
hollasch: Changed to:
> Now the inevitable tricky part.
> While our 3D space has the conventions above, this conflicts with our image coordinates,
>   where we want to have the zeroth pixel in the top-left and work our way down to the last pixel at
>   the bottom right.
> This means that our image coordinate Y-axis is inverted: Y increases going down the image.

books/RayTracingInOneWeekend.html
hollasch: Changed to:
> As we scan our image, we will start at the upper left pixel (pixel $0,0$), scan left-to-right across
>   each row, and then scan row-by-row, top-to-bottom.

books/RayTracingInOneWeekend.html
hollasch: There's nothing wrong with sampling light outside the pixel grid, or even the viewport. For example, the `sinc` image sampler (one of the best samplers available, though expensive) is infinite in extent. (See https://mathworld.wolfram.com/SincFunction.html.)

But yeah, could use another tweaking. How about this?
> Our pixel grid will be inset from the viewport edges by half the pixel-to-pixel distance.
> This way, our viewport area is evenly divided into width × height identical regions.

books/RayTracingInOneWeekend.html
hollasch: This is text instead of math, so I'm using Unicode. Changed to the character entity `&times;`.

books/RayTracingInOneWeekend.html
hollasch: Done.

books/RayTracingInOneWeekend.html
hollasch: Changed to this:
> Note that `aspect_ratio` is an ideal ratio, which we approximate as best as possible with the
>   integer-based ratio of image width / image height.
> In order for our viewport proportions to exactly match our image proportions, we use the calculated
>   image aspect ratio to determine our final viewport width.

books/RayTracingInOneWeekend.html
hollasch: Just reworded to:
> I will have the y-axis go up, the x-axis to the right, and the negative z-axis pointing in the
> viewing direction. (This is commonly referred to as _right-handed coordinates_.)

books/RayTracingInOneWeekend.html
hollasch: I liked the succinctness of just `camera` in the code that uses it. And I'd have to rewrap some lines that use it. And the `sphere` class has a member `center`, which is a `<property>` of the `<thing>`.

There — that's all my ammunition.

I'm open to renaming to `camera_center` if you still want; let me know.

hollasch:
Continuing with rebased PR: #1174

books/RayTracingInOneWeekend.html
trevordblack: I prefer  "We do Y because X" then "since X, we do Y"

books/RayTracingInOneWeekend.html
trevordblack: Looks good

books/RayTracingInOneWeekend.html
trevordblack: While it's in main have it be `camera_center`. Once inside of it's own class rename to `center`?

hollasch: I like that; will do.

books/RayTracingInOneWeekend.html

books/RayTracingInOneWeekend.html
```diff
@@ -1749,18 +1876,20 @@
...
 
-To handle the multi-sampled color computation, we'll update the `write_color()` function. Rather
-than adding in a fractional contribution each time we accumulate more light to the color, just add
-the full color each iteration, and then perform a single divide at the end (by the number of
-samples) when writing out the color. In addition, we'll add a handy utility function to the
-`interval` class: `clamp(x)`, which clamps the value `x` to the interval:
+First we need to update the `write_color()` function to account for the number of samples we use.
+Rather than adding a fractional contribution each time we accumulate more light to the color, we'll
+  just add the full color from each iteration, and then finish with a single division (by the number
+  of samples) at the end, before writing out the color.
```
trevordblack, line 1889: "We need to find the average across all of the samples that we take. We'll just add the full color from each iteration, and then finish with a single division (by the number of samples) at the end, before writing out the color."

books/RayTracingInOneWeekend.html
```diff
@@ -1799,15 +1928,48 @@
...
+  overlap each other.
+(See figure 1 of [_Cell Populations of the Retina_](https://iovs.arvojournals.org/article.aspx?articleid=2188127)
+  for an example of the different photoreceptor types in the human retina. There are many more than
+  the single rod + three cone types that most people think of.)
+There's much more to say about this, but if you want to learn more about image sampling and
+  reconstruction, see the excellent technical memo
+  [_A Pixel is Not a Little Square, A Pixel is Not a Little Square, A Pixel is Not a Little Square_](https://www.researchgate.net/publication/244986797),
+  by Alvy Ray Smith, for a fantastic introduction.
+
+In spite of all this, we will proceed with the trivial square pixel model.
```
trevordblack, line 1952: From "Which begs the question: _what's the area surrounding the pixel_?" to here.

I'm not sure if the prose makes sense here.
It feels more like Rest of Your Life material than In One Weekend

It's just very deeply technical and is quite a large aside for something that we really don't have the ink to discuss in any amount of detail.

books/RayTracingInOneWeekend.html
hollasch: Ok. Well, looks like the above comment satisfies this.

books/RayTracingInOneWeekend.html
hollasch: Went with your suggestion.

books/RayTracingInOneWeekend.html
hollasch: Latest:
> For simplicity we'll start with the camera center at $(0,0,0)$.
> We'll also have the y-axis go up, the x-axis to the right, and the negative z-axis pointing in the
> viewing direction. (This is commonly referred to as _right-handed coordinates_.)

books/RayTracingInOneWeekend.html
hollasch: Done.

books/RayTracingInOneWeekend.html
trevordblack: That or $16:9$

books/RayTracingInOneWeekend.html
```diff
@@ -3085,11 +3267,13 @@
 
 Positionable Camera
 ====================================================================================================
-Cameras, like dielectrics, are a pain to debug. So I always develop mine incrementally. First, let’s
-allow an adjustable field of view (_fov_). This is the angle you see through the portal. Since our
-image is not square, the fov is different horizontally and vertically. I always use vertical fov. I
-also usually specify it in degrees and change to radians inside a constructor -- a matter of
-personal taste.
+Cameras, like dielectrics, are a pain to debug, so I always develop mine incrementally.
+First, let’s allow an adjustable field of view (_fov_).
```
trevordblack, line 3271: allow for an

hollasch: Done.

trevordblack:
Changing my tact.

I'm going to approve this.
I have always needed to do a walk through and edit for consistency, clarity, simplicity.

I'm going to skim through this for anything I find particularly objectionable and stand by.

Fundamentally the merge could be very painful. It's better to let you do that, and unblock everything else.

books/RayTracingInOneWeekend.html
hollasch: Done.

trevordblack:
A bigger thing is that a substantial portion of the problems I'm noticing are hold-overs from the original text, like 30-40%

hollasch:
Sounds good. I'm going to continue on with a revisit to the image-sampling section to see if I can tighten it up, but as a subsequent PR.

trevordblack:
Looks good for now

trevordblack:
Decide how you want to approach outstanding comments and land

trevordblack (approved):

hollasch:
*This PR is superseded by PR #1174*

Issue #1156, opened 2023-07-09
Change Metal scatter function to random_unit_vector


trevordblack:
the metal scatter function currently uses `random_in_unit_vector`:

```
        vec3 reflected = reflect(unit_vector(r_in.direction()), rec.normal);
        scattered = ray(rec.p, reflected + fuzz*random_in_unit_sphere());
        attenuation = albedo;
        return (dot(scattered.direction(), rec.normal) > 0);
```

this should probably be replaced with `random_unit_vector` for the same reasons that `lambertian` was rewritten with `random_unit_vector`.

This would require a redraw of Figure 15:
https://github.com/RayTracing/raytracing.github.io/blob/master/images/fig-1.12-reflect-fuzzy.jpg
So that the reflection is actually on the fuzzy sphere

WDYT @hollasch

hollasch:
I've been doing a couple of new figures lately, so yeah. I'll tackle this one.

hollasch:
@trevordblack — do you want me to assign this over to you?

hollasch:
New figure is parked in PR #1159.

PR #1159, opened 2023-07-12
Fuzzy reflection rays chosen from sphere surface

hollasch:
The prior code randomly chose rays from the volume of the fuzz sphere.

Addresses #1156. This is a draft, as it will need to be manually merged by me after all of the figure renames in the camera update. Let me know if this completely addresses #1156, or if there is more work to be done.


trevordblack:
Oh FFS.
More merge filthiness.

The three uses of `random_in_unit_sphere` previously were
1. Called within `random_unit_vector`
2. Called within `lambertian::scatter`
3. Called within `metal::scatter`

My previous merge should have killed all instances of (2).
This PR should kill all instances of (3)

Unfortunately there are still a bunch of (2) in the text of the books. It's thankfully no longer in the source, but book text, yes.

As far as this PR goes, the big thing is just replacing the existing image with a new one.

Question. Are you using Limnu for the new image? It looks like you aren't?
We need to either replace all figures with something new, or keep everything as Limnu.

I need to go through and clean up the diffuse chapter ASAP so I'll let you decide what of the above you want to take over

trevordblack:
Apologies, but please choose a color other than brown for the ground, it looks like a turd

hollasch:
Updated.

trevordblack (approved):

Issue #868, opened 2021-02-16
Revisit inline code CSS style


hollasch:
Perhaps change to a rust color, with much more subtle (or removed) background color change.

hollasch:
Merged.

PR #1161, opened 2023-07-13
Update static Markdeep from 1.13 to 1.15

hollasch:
Normally, our books automatically use the latest version of Markdeep, but we also include a static version for those who are offline. This change updates the static version to the latest.

See https://casual-effects.com/markdeep/#history for changes.


trevordblack:
I'm not seeing any visual artifacts, but I am seeing differences in the way that code gets colored

trevordblack:
Screenshot from 2023-07-13 16-36-50
Screenshot from 2023-07-13 16-36-58
Screenshot from 2023-07-13 16-37-11
Screenshot from 2023-07-13 16-37-18

trevordblack:
I don't think the change is necessarily bad, which do you think is better?

hollasch:
Code style will be reviewed and addressed in #868.

trevordblack (approved):
If code coloring cannot be changed I may request this gets reverted

hollasch:
Pinning ourselves to an old version for code styling would be a bad decision in my opinion. Either way, we can address in CSS.

PR #1164, opened 2023-07-15
Distribute src/common to individual book sources

hollasch:
With the exception of the invariant STB image library and associated wrapper, this change moves all of the files out from common to each book that requires each file.

In addition, common headers that included changes from subsequent books have been reverted to their state at the end of each book.

This change also fixes up some issues found during this exercise.


trevordblack:
Two things
1. {COMMON_ALL} is still in the CMakeLists.txt
2. Why do we need common at all? I feel like we should just move `rtw_stb_image.h` up into it's relevant book folders, and move `external` up into `src/external`

trevordblack (approved):

hollasch:
I'll pull the COMMON_ALL references.

I left `rtw_stb_image.h`, because it's the one source file that's _truly_ common: it should *never* change across versions. I'm inclined to leave it, but it _is_ an odd exception. I'm slightly inclined to keep this, but you definitely raise a good point. I'll continue with your suggestion unless you change your mind.

PR #1174, opened 2023-07-19
Update text to reflect new camera updates

### LINKED ISSUE #464: InOneWeekend: clarify Defocus Blur

### LINKED ISSUE #546: Confusion surrounding use of the word "aperture"

### LINKED ISSUE #1071: Book 1: Camera terms used without being introduced

hollasch:
- Clarify confusing camera nomenclature and terms
- Update progression and code listing to match current camera code
- Update camera figures

Resolves #464
Resolves #546
Resolves #1071


hollasch:
This is the updated PR, now rebased to the latest `dev-major` commit.

I'll leave the old one open (https://github.com/RayTracing/raytracing.github.io/pull/1158) so you can go to the "Conversations" tab to continue outstanding review issues.

image

Please create new review issues on this PR.

trevordblack (approved):

trevordblack:
Please hold off on closing out #546 and assign that issue to me.

I would like to go through and figure out if my concerns still hold

PR #1175, opened 2023-07-19
Change book2 to final parameterized scene

hollasch:
Instead of passing a "high quality" flag, pass in the image size, samples per pixel and max depth directly for simplicity and clarity.


trevordblack:
Wow. Yes. Change it.

trevordblack (approved):

PR #1176, opened 2023-07-19
Update book covers

### LINKED ISSUE #425: TheNextWeek final scene perlin rendering along wrong axis
The perlin noise sphere at the end of The Next Week isn't rendering any noise. It could be due to my render being low spp. But, it appears that there isn't a noise texture, and is being rendered as flat lambertian.

### LINKED ISSUE #1173: Update book covers

hollasch:
- Reflect new authors list
- Update to latest images
- Move cover images to their own folder

For the time being, I'm preserving the old covers in images/covers/old. These should be removed in the future.

Resolves #425
Resolves #1173


trevordblack:
This also resolves #425

trevordblack (approved):

PR #1179, opened 2023-07-25
Update antialiasing chapter

hollasch:
This addresses some of the criticisms of the prior text included in the camera update, which discussed square versus non-square pixels.


hollasch:
Rebased to #1181

PR #1181, opened 2023-07-25
Update antialiasing chapter

### LINKED ISSUE #695: Book 1 Antialiasing motivation is incomplete
Book 1 Antialiasing states that the purpose of antialiasing is to eliminate jagged edges. If this were the only motivation, I would skip the code complexity and render a larger image. However, I believe there is a second motivation. The following two sections introduce materials which absolutely depend on integrating multiple samples of the same pixel. I think this deserves a mention when introducing antialiasing. Something like, "Our approach to simulating the behavior of light reflecting off of a point involves randomness, so our result would be heavily speckled inside shapes if we didn't sample each pixel many times."

hollasch:
This addresses some of the criticisms of the prior text included in the camera update, which discussed square versus non-square pixels.

Resolves #695


hollasch:
This is a rebase of #1181 .

PR #1190, opened 2023-07-27
Move all v4-alpha content into alpha/ directory

hollasch:


trevordblack:
It's not clear to me why all the sidecar alpha stuff is necessary, but I have no objections to the work.

The Changelog is an extremely welcome change. Thanks Steve

trevordblack (approved):

PR #1192, opened 2023-07-30
Book1 review final alpha

trevordblack:


alpha/books/RayTracingInOneWeekend.html
```diff
@@ -561,10 +564,11 @@
 
 (If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
 because the _actual_ aspect ratio is the integer-based ratio of image width over image height, which
-will yield zero for values less than one.)
+_may_ not be the same as `aspect_ratio` because `image_height` may have off round-off error and may
+have been set to 1.)
```
hollasch, line 567: This over focuses on the case where image_height is small; it's more general than that.

Consider the case where aspect_ratio is 4/3, and we're given an image_width of 25. In this case, the corresponding image_height would be 25/(4/3) = 18, yielding a true aspect ratio of 25/18 = 1.39.

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -741,20 +745,22 @@
...
 ====================================================================================================
 Let’s add a single object to our ray tracer. People often use spheres in ray tracers because
-calculating whether a ray hits a sphere is pretty straightforward.
+calculating whether a ray hits a sphere is relatively simple.
 
 
 Ray-Sphere Intersection
 ------------------------
-Recall that the equation for a sphere of radius $r$ that is centered at the origin is:
+The equation for a sphere of radius $r$ that is centered at the origin is well known to be:
```
hollasch, line 752: Seems a bit condescending.

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -882,24 +888,27 @@
...
 First, let’s get ourselves a surface normal so we can shade. This is a vector that is perpendicular
 to the surface at the point of intersection. There are two design decisions to make for normals.
-The first is whether these normals are unit length. That is convenient for shading so I will say
-yes, but I won’t enforce that in the code. This could allow subtle bugs, so be aware this is
-personal preference as are most design decisions like that. For a sphere, the outward normal is in
-the direction of the hit point minus the center:
+The first is whether or not these normals are unit length. This is convenient for shading so I will
+say yes, but we won’t enforce that in the code. This could allow for subtle bugs, so be aware that
+this is a personal preference. Note that many of the design decisions in this book prioritize
+convenience over absolute perfection.
```
hollasch, line 893: We are definitely setting normal vectors to unit length, and are depending on that in the code.

The issue here is that if you _don't_ assume unit normals, then any code that needs it will have to always normalize the vectors, leading to cases where a normal vector is normalized multiple times, leading to worse performance. On the other end, the geometric primitives themselves have enough knowledge to generate unit-length normals in many cases. In our sphere code, we divide the normal by the sphere radius: an optimization that avoids the need for a square root. Our new quad code explicitly normalizes the normal vector because it's needed for other operations, so the cost to providing this is effectively zero.

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -1637,12 +1649,17 @@
 The new camera class will contain two public methods `initialize()` and `render()`, plus two
   private helper methods `get_ray()` and `ray_color()`.
 
-Ultimately, the camera will use a set of public variables as parameters instead of using a
-  constructor with a long sequence of arguments, or using a bunch of setter methods.
-It's crude, but simple.
-Given this, the usage pattern will be such that the main code will create a default camera,
-  then set various camera parameters (perhaps at multiple places in the code).
-After everything's been set, it will call the `render()` method.
+Ultimately, the camera will use the simplest bring-up that we could think of. The usage pattern
```
hollasch, line 1651: "bring-up"?

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -2646,11 +2663,11 @@
...
 An Abstract Class for Materials
 --------------------------------
-If we want different objects to have different materials, we have a design decision. We could have a
-universal material with lots of parameters and different material types just zero out some of those
-parameters. This is not a bad approach. Or we could have an abstract material class that
-encapsulates behavior. I am a fan of the latter approach. For our program the material needs to do
-two things:
+If we want different objects to have different materials, we have a design decision. We could have
+a universal material type with lots of parameters so each material type can just zero out the
+parameters that don't affect it. This is not a bad approach. Or we could have an abstract material
```
hollasch, line 2667: I think "ignore" would be better here than "zero out".

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -2684,9 +2701,10 @@
 A Data Structure to Describe Ray-Object Intersections
 ------------------------------------------------------
 The `hit_record` is to avoid a bunch of arguments so we can stuff whatever info we want in there.
-You can use arguments instead; it’s a matter of taste. Hittables and materials need to know each
-other so there is some circularity of the references. In C++ you just need to alert the compiler
-that the pointer is to a class, which the “class material” in the hittable class below does:
+You can use arguments instead of an encapsulated type, it’s just a matter of taste. Hittables and
+materials need to be able to reference the other's type in code so there is some circularity of the
+references. In C++ you just need to alert the compiler that the pointer is to a class, which the
+`class material;` in the hittable class below does:
```
hollasch, line 2706: Hmmm, I think this could be improved. What do you think about this?

> In C++ we add the line `class material;` to tell the compiler that `material` is a class that will be defined later.

Optionally, we could continue with

> Since we're just going to specify a pointer to the class, the compiler doesn't need to know the details of the class, solving the circular reference issue.

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -2855,7 +2872,7 @@
 
 Mirrored Light Reflection
 --------------------------
-For smooth metals the ray won’t be randomly scattered. The key math is: how does a ray get
+For a polished mirror the ray won’t be randomly scattered. The key question is: How does a ray get
```
hollasch, line 2874: To me "mirror" implies a perfect reflector, making "polished" redundant (though I think we're discussing _imperfect_ reflectors here. How about "shiny materials" or "shiny metal"?

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -3737,10 +3758,10 @@
 
 The reason we have defocus blur in real cameras is because they need a big hole (rather than just a
   pinhole) through which to gather light.
-A large hole would defocus everything, but if we stick a lens in the hole, there will be a certain
-  distance where everything is in focus.
-Objects appear blurrier the further they are from that distance (either closer or further from the
-  camera).
+A large hole would defocus everything, but if we stick a lens in front of the sensor, there will be
+a certain distance at which everything is in focus.
```
hollasch, line 3761: Oof. I hate that we mention a sensor, but never define it what that means and then continue with a model that doesn't have one anyway. I also wish we could drop the concept of a lens entirely, as we don't have that either (or any object that bends light rays), and say that having one must imply "a certain distance at which everything is in focus" as if we just proved something. Sigh; moving along.

hollasch (changes requested):

alpha/books/RayTracingInOneWeekend.html
hollasch: Should I take a stab at communicating the above? Seems useful.

alpha/books/RayTracingInOneWeekend.html
hollasch: → "implementation"?

trevordblack:
I'll try to get to this in about 30-60 mins

alpha/books/RayTracingInOneWeekend.html
trevordblack: I mean, I tried to address that, that's what "may have off round-off error" us trying to get across. I'll try again

alpha/books/RayTracingInOneWeekend.html
trevordblack: Hmm. I got someone saying that they want us to explain where we got the equation of the sphere from.

I didn't want to do that. Let me figure out another way to say this.

alpha/books/RayTracingInOneWeekend.html
trevordblack: Sure. I'll have updates in a minute.

alpha/books/RayTracingInOneWeekend.html
trevordblack: I specifically went away from "metals" in the first sentence because (even though the chapter is called "metals") most everyday metals are not mirror-like

And that's ultimately what we're modeling here.

Let me think on it

trevordblack: Going with polished metals. Tell me if you disagree with that

alpha/books/RayTracingInOneWeekend.html
trevordblack: I clocked this as well

I just dropped it as I didn't think we had time, nor was that a string I wanted to pull on.

I'll see if there's anything small worth doing

trevordblack:
Feel free to hack on this

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -561,10 +561,13 @@
...
-because the _actual_ aspect ratio is the integer-based ratio of image width over image height, which
-_may_ not be the same as `aspect_ratio` because `image_height` may have off round-off error and may
-have been set to 1.)
+If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
+because the value set to `aspect_ratio` is the ideal ratio, it may not be the _actual_ ratio
+between `image_width` and `image_height`. If `image_height` was allowed to be real valued--rather
+than just an integer--then it would fine to use `aspect_ratio`. But the _actual_ ratio
+between `image_width` and `image_height` can vary based on two parts of the code. First,
+`integer_height` is rounded down to the nearest integer, which can increase the ratio. Second, we
+don't allow `integer_height` to be less than one, which can also change the actual aspect ratio.
```
hollasch, line 570: Sweet.

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -1648,15 +1652,15 @@
 The new camera class will contain two public methods `initialize()` and `render()`, plus two
 private helper methods `get_ray()` and `ray_color()`.
 
-Ultimately, the camera will use the simplest bring-up that we could think of. The usage pattern
-will be such that the main code will call the default constructor with no arguments, the main
-function will then modify the camera's public variables through simple assignment, before finally
-everything gets initialized by a call to the `initialize()` function.
+Ultimately, the camera will use the simplest implementation that we could think of. The usage
+pattern will be such that the main code will call the default constructor with no arguments, then
+main will modify the camera's public variables through simple assignment, before everything finally
```
hollasch, line 1657: "main" → "the calling code" (this isn't actually changed in `main`).

alpha/books/RayTracingInOneWeekend.html
```diff
@@ -2702,8 +2706,9 @@
 The `hit_record` is to avoid a bunch of arguments so we can stuff whatever info we want in there.
 You can use arguments instead of an encapsulated type, it’s just a matter of taste. Hittables and
 materials need to be able to reference the other's type in code so there is some circularity of the
-references. In C++ you just need to alert the compiler that the pointer is to a class, which the
-`class material;` in the hittable class below does:
+references. In C++ we add the line `class material;` to tell the compiler that material is a class
```
hollasch, line 2709: Inline code style `material`.

hollasch (approved):

alpha/books/RayTracingInOneWeekend.html
hollasch: Looks good.

trevordblack:
Changes look good. OK with the landing

PR #1201, opened 2023-08-11
Fix refs to random_in_hemisphere

### LINKED ISSUE #1198: random_in/on_hemisphere
See also #1197 

Code listing in book 1 has a caption referencing `random_in_hemisphere`, but it should be _on_hemisphere.

hollasch:
These should be random_on_hemisphere.

Resolves #1198


books/RayTracingTheRestOfYourLife.html
```diff
@@ -1819,7 +1819,7 @@
         bool scatter(const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered)
         const override {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            auto scatter_direction = random_in_hemisphere(rec.normal);
+            auto scatter_direction = random_on_hemisphere(rec.normal);
```
hollasch, line 1822: I _think_ this is correct. If not, we need to implement `random_in_hemisphere()`. It's quickly replaced with different code in the text.

PR #1211, opened 2023-08-13
Make linear_to_gamma robust for negative inputs

### LINKED ISSUE #1202: Harden linear_to_gamma() function?
We currently have the following implementation:

    inline double linear_to_gamma(double linear_component)
    {
        return sqrt(linear_component);
    }

This will barf on negative values, returning NaN. Should we harden this like so?

    inline double linear_to_gamma(double linear_component)
    {
        if (linear_component < 0)
            return 0;

        return sqrt(linear_component);
    }

or

    inline double linear_to_gamma(double linear_component)
    {
        return sqrt(fmax(0, linear_component));
    }

hollasch:
Without this hardening, we will return NaNs for negative inputs, which can happen from time to time.

I also ran across fmin/fmax in one implementation, which we never commented on in the books. Added an explanation for these two functions for readers unfamiliar with C++.

Resolves #1202


hollasch:

PR #1212, opened 2023-08-14
Reorganize the quads chapter

### LINKED ISSUE #1204: Book 2 section "Orienting Points on The Plane" simplification
This section derives the formulas for the scalar values α and β. The derivation for these values is more detailed that similar other sections in the book, and on review I think it would improve things to cover the concepts and usage of these values, along with their formulas, and put the derivation in a section on its own, with guidance to the reader that they can skip this if they're not interested.

### LINKED ISSUE #1205: Book 2 section "Interior Testing of The Intersection Using UV Coordinates": expand on alternative primitives
The current text reads like so:

> Pause a bit here and consider that if you use the (α,β) coordinates to determine if a point lies inside a quadrilateral (parallelogram), it's not too hard to imagine using these same 2D coordinates to determine if the intersection point lies inside any other 2D (planar) primitive!
>
> We'll leave these additional 2D shape possibilities as an exercise to the reader, depending on your desire to explore. Consider triangles, disks, and rings (all of these are surprisingly easy). You could even create cut-out stencils based on the pixels of a texture map, or a Mandelbrot shape!

We should flesh this out just a tiny bit more. Perhaps move this text after the next listing, and mention that one would only need to write a different `is_interior(double a, double b, hit_record& rec)` function that returns whether a given input coordinate (a,b) is inside the primitive. We might also supply the text describing how we'd do this for disks to get folks thinking.

hollasch:
This does two things:

- Moves the planar coordinates derivation into its own skippable section.

- Moves the alternate 2D primitives discussion to its own section at the end of the quad chapter, and points to the `alternate-2D-primitves` tag for triangles, ellipses, and annuli.

Resolves #1204
Resolves #1205

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2598,10 +2598,64 @@ so that
 
   $$ \mathbf{P} = \mathbf{Q} + \alpha \mathbf{u} + \beta \mathbf{v} $$
 
-If $\mathbf{u}$ and $\mathbf{v}$ were guaranteed to be orthogonal to each other (forming a 90° angle
-between them), then this would be a simple matter of using the dot product to project $\mathbf{P}$
-onto each of the basis vectors $\mathbf{u}$ and $\mathbf{v}$. However, since we are not restricting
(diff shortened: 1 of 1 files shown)


hollasch:

PR #1216, opened 2023-08-15
BVH optimization: split on the longest bbox axis

### LINKED ISSUE #1007: Change BVH Creation algorithm
At present, the text outlines the algorithm for creating a BVH structure:

```
1. randomly choose an axis
2. sort the primitives (using std::sort)
3. put half in each subtree
```

We should consider replacing--either outright, or in a revision chapter--random axis choice with using the axis with the longest extent.

This appears to be the first "good-enough" solution that you find in the literature, once you go beyond random axis.

hollasch:
I see about an 18% speedup on my box.

Resolves #1007


hollasch:

PR #1223, opened 2023-08-18
MSVC warning-clean pass

hollasch:
Bumping up warnings to max with explicit exceptions for the Microsoft Visual C++ compiler. Found a number of issues:

- Some simple signed/unsigned mismatches
- Some unused function parameters. I've disabled this warning generally, as we often have virtual functions that may not use some parameters that other derived classes do. Still this found some functions that _never_ use some parameters.
- Found a member variable data shadowing issue in quad::pdf_value(). Renamed parameters `o` and `v` to `origin` and `direction` for better clarity and to fix the shadowing.
- Found what looks remnants of earlier code versions in cos_cubed.cc. Ensured that the book never develops the unnecessary code, so deleted the unnecessary parts.
- Fixed some minor whitespace issues.


CMakeLists.txt
```diff
@@ -95,9 +95,26 @@ include_directories(src)
...
+    # /wd #### - Disable warning
+    # /w3 #### - set warning to level 3
+    add_compile_options("/Wall")    # Enable all warnings
     add_compile_options("/w3 5038") # Data member will be initialized after [other] data member
+    add_compile_options("/we 4265") # Class has virtual functions, but its non-trivial destructor is not virtual
     add_compile_options("/we 5204") # Class has virtual functions, but its trivial destructor is not virtual
+    add_compile_options("/wd 4100") # unreferenced formal parameter
+    add_compile_options("/wd 4365") # conversion from 'size_t' to 'const int __int64', signed/unsigned mismatch
+    add_compile_options("/wd 4457") # index X range checked by comparison on this line
+    add_compile_options("/wd 4464") # feeds call on this line
```
armansito: The MSVC docs refer to this as `relative include path contains '..'`: https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4464?view=msvc-170

CMakeLists.txt
```diff
@@ -95,9 +95,26 @@ include_directories(src)
...
+    # /we #### - treat warning as error
+    # /wd #### - Disable warning
+    # /w3 #### - set warning to level 3
+    add_compile_options("/Wall")    # Enable all warnings
     add_compile_options("/w3 5038") # Data member will be initialized after [other] data member
+    add_compile_options("/we 4265") # Class has virtual functions, but its non-trivial destructor is not virtual
     add_compile_options("/we 5204") # Class has virtual functions, but its trivial destructor is not virtual
+    add_compile_options("/wd 4100") # unreferenced formal parameter
+    add_compile_options("/wd 4365") # conversion from 'size_t' to 'const int __int64', signed/unsigned mismatch
+    add_compile_options("/wd 4457") # index X range checked by comparison on this line
```
armansito: Docs say 4457 is `declaration of 'identifier' hides function parameter`: https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4457?view=msvc-170

armansito (approved):
The edits look good to me overall. I double checked the individual MSVC warning codes and found a couple of them that are documented differently on https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/ (though I could have been looking in the wrong place).

hollasch:
I have no idea what's going on, but those two warnings no longer display, so I must have looked cross-eyed at the build output or something. Anyway, I've just removed those two warnings and everything still build warning clean for both debug and release.

Additionally, I turned all warnings off and then on again one-by-one to ensure that they are still needed, and they all check out. Nice catch.

PR #1227, opened 2023-08-19
From C++ static_cast to ctor-like conversion

### LINKED ISSUE #1222: Code improvements: InOneWeekend §4.2
```c++
// Calculate the image height, and ensure that it's at least 1.
int image_height = static_cast<int>(image_width / aspect_ratio);
```
There is no need for `static_cast`. Assigning to an `int` variable will automatically do that:
```c++
// Calculate the image height, and ensure that it's at least 1.
int image_height = image_width / aspect_ratio;
```
---
```c++
// Viewport widths less than one are ok since they are real valued.
auto viewport_height = 2.0;
auto viewport_width = viewport_height * (static_cast<double>(image_width)/image_height);
```
I am not sure why the division has to be done first, but if you remove the parentheses, then `static_cast` becomes unnecessary since the first member of the expression is `double` and will cause other members to be promoted to `double`. So, you can simply write:
```c++
// Viewport widths less than one are ok since they are real valued.
auto viewport_height = 2.0;
auto viewport_width = viewport_height * image_width / image_height;
```

hollasch:
Instead of `static_cast<double>(x)` switch to `double(x)` for easier readability for non-C++ readers and simple for C++ readers alike. In the end, these are the same thing.

Resolves #1222

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -122,9 +122,9 @@ Let’s make some C++ code to output such a thing:
                 auto g = double(j) / (image_height-1);
                 auto b = 0;
 
-                int ir = static_cast<int>(255.999 * r);
-                int ig = static_cast<int>(255.999 * g);
-                int ib = static_cast<int>(255.999 * b);
+                int ir = int(255.999 * r);
+                int ig = int(255.999 * g);
+                int ib = int(255.999 * b);
 
                 std::cout << ir << ' ' << ig << ' ' << ib << '\n';
             }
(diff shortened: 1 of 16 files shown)


armansito (approved):
Thank you for the clean up, I think this is a good improvement overall.

P.S.:
Something occurred to me while looking through the diff (though I don't want this to block this PR): the code examples currently use `auto` frequently. Sometimes this is good for brevity (for example with gnarly template expansions) but there are a few spots where I think `auto` obscures the code a little bit.

For instance, from `src/TheNextWeek/perlin.h`:
```
auto u = p.x() - floor(p.x());
auto v = p.y() - floor(p.y());
auto w = p.z() - floor(p.z());
auto i = int(floor(p.x()));
auto j = int(floor(p.y()));
auto k = int(floor(p.z()));
```
The types should generally be clear if you read it carefully but I think it would be clearer with explicit type annotations, especially given the intentional type conversions involved:
```
double u = p.x() - floor(p.x());
double v = p.y() - floor(p.y());
double w = p.z() - floor(p.z());
int i = int(floor(p.x()));
int j = int(floor(p.y()));
int k = int(floor(p.z()));
```
Similarly, in `src/TheRestOfYourLife/camera.h` the assignment to `viewport_height` relies on implicit type deduction, which should be a `double` but this may not be immediately obvious upon reading. I think it would be clearer to express that line as:
```
double viewport_height = 2 * h * focus_dist;
double viewport_width = viewport_height * (double(image_width)/image_height);
```
Just food for thought.

PR #1229, opened 2023-08-20
Reverse ray-sphere direction and calculations

### LINKED ISSUE #1191: Equation for the intersection of a ray and a sphere is backwards
It's written as
(P - C)

It should be
(C - P)

The math and everything is still correct but that (-1) gets squared out, but all of the ray intersections are from the center to the ray origin. Not the other way around

hollasch:
Trevor observed that the equation for the ray-sphere intersection was backwards in that it used (P - C) [center to ray origin] instead of (C - P) [ray origin to sphere center].

That's all well and good, but it ends up rippling all the way back to the fundamental sphere equations presented in the text.

Sigh. And I couldn't resist.

In flipping everything so it flows without an odd inversion in the middle, I also discovered some annoying minus signs that could be eliminated, plus some bits that needed a tiny bit more explanation. Quite a bit more work than it originally appeared, but I'm pleased with the results.

Resolves #1191

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -560,10 +560,10 @@ to give us the desired aspect ratio. Here's a snippet of what this code will loo
 If you're wondering why we don't just use `aspect_ratio` when computing `viewport_width`, it's
 because the value set to `aspect_ratio` is the ideal ratio, it may not be the _actual_ ratio between
 `image_width` and `image_height`. If `image_height` was allowed to be real valued--rather than just
-an integer--then it would fine to use `aspect_ratio`. But the _actual_ ratio between `image_width`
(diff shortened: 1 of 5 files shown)


hollasch:
I recommend that you review the rendered books locally since the changes are so math heavy.

hollasch:
Comment from @trevordblack on Slack:
> But the code was initially h before I changed it to half_b because the reason for naming it h was not properly motivated. We just changed the variable name and the code didn't clarify either
> I might push back against the change to having it return to h if it's not thoroughly motivated

The reason for renaming to `h` is two-fold:
1. It was a bit jarring to see this same value named `h` in the text, but `half_b` in the code.
2. In the current formulation, $h = -b/2$, so it would have to be named `neg_half_b` or something. However, I'd rather this just be a variable than a recap of the equation used to compute it.

PR #1232, opened 2023-08-21
Some minor listing cleanup from book 1 progression

hollasch:


armansito (approved):

books/RayTracingTheNextWeek.html
```diff
@@ -71,9 +71,6 @@
 
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-        ray(const point3& origin, const vec3& direction) : orig(origin), dir(direction), tm(0)
```
armansito, line 74: [src/TheNextWeek/ray.h](https://github.com/RayTracing/raytracing.github.io/blob/dev/src/TheNextWeek/ray.h) still defines both constructor overloads and `time` doesn't have a default value there. Does the source code need to get updated as well?

armansito: Ah nvm, I see that the progression branch for book 2 has the change: https://github.com/RayTracing/raytracing.github.io/blob/progression/2/2023-08/src/TheNextWeek/ray.h

hollasch: Looking at the end state source of book 2, I think this is another problem.

hollasch: Yup. It is.

PR #1273, opened 2023-09-15
Use const ref params

### LINKED ISSUE #1250: [Book 1] Listing 4: use const references for parameters in utility functions
Regarding the following snippet from listing 4:
```cpp
inline vec3 operator/(vec3 v, double t) {
    return (1/t) * v;
}

/* omitted code for brevity */

inline vec3 unit_vector(vec3 v) {
    return v / v.length();
}
```
They are the only utility functions that have parameters that are not constant references, and instead use copy-by value parameter passing. I believe change those parameters to const references would be more consistent with the other utility functions, and could even help a bit with performance (even though I'm no optimization expert), as in:

```cpp
inline vec3 operator/(const vec3 &v, double t) {
    return (1/t) * v;
}

/* omitted code for brevity */

inline vec3 unit_vector(const vec3 &v) {
    return v / v.length();
}

### PR DESCRIPTION
**Pull request by hollasch**
There were many cases where we were passing vectors, points, rays and colors by value. Passing by const reference or mutable reference is more efficient.

This speeds up book1 by 13% and book2 by 3.5%.

Resolves #1250

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -2291,7 +2291,7 @@ But first, let's quickly update the `isotropic` material:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class isotropic : public material {
       public:
-        isotropic(color c) : albedo(make_shared<solid_color>(c)) {}
+        isotropic(const color& c) : albedo(make_shared<solid_color>(c)) {}
         isotropic(shared_ptr<texture> a) : albedo(a) {}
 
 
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: books/RayTracingTheRestOfYourLife.html [hunk 1]
```diff
@@ -3595,9 +3595,9 @@
             return sum;
         }
 
-        vec3 random(const vec3& o) const override {
+        vec3 random(const vec3& origin) const override {
```
armansito: The declaration in the base class (i.e. `hittable::random` and `hittable::pdf_value`) still use singleletter names. I think it makes sense to update the listing (Listing 38) and the code to match.

I also noticed that `sphere` (TheRestOfYourLife/sphere.h) also still uses the single-letter names, so I would update it to be consistent. AFAICT `sphere` is the only `hittable` subclass that needs to be updated but I recommend double checking all of them in case I missed one.

armansito (approved):
Looks good. I tried to review this as thoroughly as possible but I may not have caught every discrepancy.

books/RayTracingTheRestOfYourLife.html
hollasch: Will do.

hollasch:
Ok, feedback incorporated @armansito .

armansito (approved):
Thanks!

PR #1320, opened 2023-10-28
Better ordering of hit record in ray_color() func

hollasch:
Originally due to GitHub user @sshaplygin.


hollasch:
Original change due to GitHub user @sshaplygin.

armansito (approved):

PR #1323, opened 2023-10-31
Clarify the addition of the inverted glass sphere

### LINKED ISSUE #1278: [Book1] A hollow glass sphere
Hey, huge noob here.

I’m following the first book in rust; once I reached the chapter: « Modeling a Hollow Glass Sphere » I started to have different results than the one you show in the book:
image

The thing is, after I spent a few hours trying to find a difference between my code and yours, I decided to run your code directly and found a different result than the one in the book (and different than mine as well 😓).

image

On top is my code, and on the bottom is yours.

I don’t know where the issue is located and am a bit lost, but I’m pretty sure that my code does the same thing as the c++ version, except the two final chapters are not implemented. (and that may explain the slight difference between the two renders I made?) 

I pushed my code here if you want to check it out: https://github.com/irevoire/raytracing2. You can run it with `cargo run --release` after installing rust.

hollasch:
It turns out that several readers missed the fact that the hollow glass sphere is modeled by _adding_ a second inverted sphere, and instead mistook the change to be a _modification_ of the original glass sphere. This change should helpfully make this mistake less likely.

Resolves #1278

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -3481,7 +3481,8 @@ finite bubble inside the sphere, so the outward normal needs to point toward the
 Dividing by the (negative) radius flips the normal as we want. If you implmented your code like the
 second example above, you'll want to fix that now.
 
-Let's use this hollow sphere hack to model a sphere with a given thickness:
+Let's use this hollow sphere hack to model a sphere with a given thickness -- that is, with an
+interior surface. To do this, add a second _inverted_ glass sphere inside the original glass sphere:
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     ...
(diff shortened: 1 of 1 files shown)


hollasch:
@irevoire, @meemsbror

books/RayTracingInOneWeekend.html
```diff
@@ -3481,7 +3481,8 @@
 Dividing by the (negative) radius flips the normal as we want. If you implmented your code like the
 second example above, you'll want to fix that now.
 
-Let's use this hollow sphere hack to model a sphere with a given thickness:
+Let's use this hollow sphere hack to model a sphere with a given thickness -- that is, with an
+interior surface. To do this, add a second _inverted_ glass sphere inside the original glass sphere:
```
armansito: Slight rephrasing suggestion: `Let's use this hollow sphere hack to model the interior surface of a sphere with a given thickness. To do this, add a ...`

armansito (approved):
Thank you for the clarification! I just left a small suggestion.

books/RayTracingInOneWeekend.html

irevoire:
Nice, it helps!

PR #1325, opened 2023-11-03
Fix change highlighting for listing 30, 46

### LINKED ISSUE #1313: Listing 3.30: Inconsistent change highlighting
Benny Tsang writes:
> (Book 3) Listing 30: the highlighting is inconsistent with what is modified, e.g., `attenuation` should be `alb` to be consistent with the function argument.

hollasch:
Resolves #1313

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -2251,17 +2251,17 @@ We can rewrite our Lambertian material using this to get:
     class lambertian : public material {
       public:
         ...
-
-        bool scatter(
-            const ray& r_in, const hit_record& rec, color& alb, ray& scattered, double& pdf
-        ) const override {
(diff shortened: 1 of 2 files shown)


hollasch:
@bthtsang

armansito (commented):
I noticed that between listings 23 and 29 the `attenuation` parameter of `lambertian::scatter` also gets renamed to `alb` but I believe the change isn't highlighted.

This got me thinking though: why rename `attenuation` to `alb` at all? In the final state of the code albedo gets assigned to `scatter_record::attenuation` so the rename in between feels confusing to me. Ultimately "attenuation" is the right terminology.

hollasch:
Good changes; thanks! I've incorporated your suggestions and also made a few cosmetic tweaks.

armansito (approved):
Thanks!

PR #1347, opened 2023-12-29
Enclose free URLs with angle-bracket delimiters

hollasch:
This makes the parsing unambiguous in the presence of trailing punctuation (fixing one bug) and is more clear to read.


armansito (approved):
This seems to fix the broken links. I only ran a cursory grep to check if there are any remaining broken links beyond those you fixed and I didn't come across any but I wasn't very thorough.

Issue #1361, opened 2024-01-18
[Book 2] Listing 68: Should include "aabb::empty" and "aabb::universe" to be consistent with Listing 69


rupsis:
No real issue, more of an inconsistency. Listing 68 adds the `operator+`

```
class aabb {
    ...
};

aabb operator+(const aabb& bbox, const vec3& offset) {
    return aabb(bbox.x + offset.x(), bbox.y + offset.y(), bbox.z + offset.z());
}

aabb operator+(const vec3& offset, const aabb& bbox) {
    return bbox + offset;
}
```

but doesn't include the `aabb::empty` & `aabb::universe` in the diff. Whereas Listing 69 includes the `interval::empty` & interval::universe` in the diff. 

happy to make this change :)

hollasch:
Fixed in 1d2740a

PR #1362, opened 2024-01-18
Listing 68: Should include "aabb::empty" and "aabb::universe" to be consistent with Listing 69

rupsis:
Addresses #1361

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -3564,7 +3564,10 @@ offset` above requires some additional support.
         ...
     };
 
+    const aabb aabb::empty    = aabb(interval::empty,    interval::empty,    interval::empty);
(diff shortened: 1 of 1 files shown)


hollasch:
Looks good, but new or changed code should have change highlighting.

hollasch:
Also, note that there's a weirdness to the Markdeep code highlighting processing. If you need a blank line before highlighted code, you have to have _two_ blank lines before the opening highlight fence. See other listing source for examples.

rupsis:
> Looks good, but new or changed code should have change highlighting.

This isn't new nor changing, rather it's just a contextual clue in the diff of where the new code should fall in the `aabb` class. 

Listing 69 includes the `interval::empty` & `interval::universe` 
image

So this change is for consistency. 

Alternatively, I could remove `interval::empty` & `interval::universe`  from Listing 69, but In my opinion it's tougher to read.

hollasch (approved):

Issue #1360, opened 2024-01-18
[Book 2] image 15: is the "Perlin noise, marbled texture" image wrong?


rupsis:
Seems like the perlin marble noise image is from the old book?
![image](https://github.com/RayTracing/raytracing.github.io/blob/release/images/img-2.15-perlin-marble.png?raw=true)

Running `two_perlin_spheres()` on dev branch produces this results:
![perlin_marble](https://github.com/RayTracing/raytracing.github.io/assets/9646385/2ee6e2ab-bbc7-4683-b257-6e8f973bccc6)

Happy to update the image (or related code) :)

rupsis:
I suppose if this _is_ the case, then subsequent  images (17, 18) would also need to be updated.

hollasch:
Fixed with #1363

Issue #863, opened 2021-02-11
Validate Book 3 progression for milestone v4.0.0


hollasch:
Run through the book progression before v4 release.

Validate/update rendered images. Make sure that they are PNG format, and sized according to the rendered size specified in the code.

- [x] 3.1: Cornell box, no stratification
- [x] 3.2: Cornell box, with stratification (*see note below*)
- [x] 3.3: Cornell box, refactored
- [x] 3.4: Cornell box, with imperfect PDF
- [x] 3.5: Cornell box, with uniform hemispherical sampling (_new render_)
- [x] 3.6: Cornell box, with orthonormal basis scatter function
- [x] 3.7: Cornell box, sampling only the light, 10 samples per pixel
- [x] 3.8: Cornell box, light emitted only in the downward direction
- [x] 3.9: Cornell box with a cosine density PDF
- [x] 3.10: Cornell box, sampling a hittable light, 10 samples per pixel
- [x] 3.11: Cornell box, mixture density of cosine and light sampling
- [x] 3.12: Cornell box with arbitrary PDF functions
- [x] 3.13: Cornell box with glass sphere, using new PDF functions
- [x] 3.14: Cornell box using a mixture of glass & light PDFs
- [x] 3.15: Cornell box with anti-acne color function

Compare render 3.01 and 3.02. They _should_ be the same. If they have slightly different noise, then the text needs to be slightly altered:
```
You should get exactly the same picture.
```
→
```
You should get the same picture, hopefully with less noise
```

hollasch:
Review #900.

PR #1363, opened 2024-01-19
updating imag2 2.15, 2.17, 2.18 with v4 outputs

rupsis:
Addresses #1360 

The old V3 Perlin noise has a _slightly_ phase then V3:
```c++
// v3 noise_texture
virtual color value(double u, double v, const vec3& p) const override {
    return color(1,1,1)*0.5*(1 + sin(scale*p.z() + 10*noise.turb(p)));
  }
```

vs V4 
```c++
// v4 noise_texture
        auto s = scale * p;
        return color(1,1,1) * 0.5 * (1 + sin(s.z() + 10*noise.turb(s, 7)));
    }
```

Because of this, images 2.15, 2.17, & 2.18 don't match up with V4's output. 

Ran the steps agains V4, and updated the images


hollasch (approved):

cccccz:
Hi I know this issue is already closed but I still want to address that the final scene seems to be using the v3 perlin texture as well.
using noise_texture in v3:
image

comparing to the output using noise_texture in v4:
image

also I think it would be nice to mention that the "sweet dream" picture(higher sample version of the final scene) will take more than 40 minutes to render on a 8-core cpu(using the tile and multithreading tech mentioned on the last chapter of Book 1), which is crazy.
 @hollasch

rupsis:
@cccccz ooo, good catch. I'll need to update the final picture. 

I assumed "(sweet dreams)" was a little nod to how long the render would take. I.e, render overnight. 

But giving a little more indication of how long the render will take might be useful.

hollasch:
Two things here. First, please don't fork issues by adding separate content to existing issues. This issue is now about updating three images, then an additional one, and then a recommended update to the text about rendering time.

Second, issue #779 already tracks the complete update of _all_ images in book 2. #863 tracks the complete update of all images for book 3. These issues are slotted into the release milestones because the body and code of books two and three are still undergoing major-level changes, and we don't want to have to update the images every time we modify the code.

rupsis:
@hollasch  Sorry I should have been more clear, I was planning on updating the final image in a different issue. 

I was unaware of #779. Seems like we should hold off on image updates until all alpha-2 milestones have been completed?

hollasch:
Yeah, I'd recommend that.

Issue #1217, opened 2023-08-15
Port Pete's in1weekend.blogspot.com articles to the raytracing wiki


hollasch:
Peter originally created https://in1weekend.blogspot.com/ to serve as updates and further information for the _Ray Tracing in One Weekend_ series. Though it's a bit out of date, the articles contain useful information. (And sadly, the comments have been filled with spam.)

We'd like to port this content over to [the wiki](https://github.com/RayTracing/raytracing.github.io/wiki) to keep it close to the project and open for further updates.

hollasch:
See also #676

hollasch:
See also #407

rupsis:
Addressed in #1378

Issue #676, opened 2020-07-19
Replace link to in1weekend to the wiki page


trevordblack:
Book 3 has a line pointing the reader to the in1weekend blogspot.

This should be redirected to the aggregated list of links.

rupsis:
Addressed in #1378

PR #1378, opened 2024-02-02
Update all Books to point to Project Wiki instead of in1weekend

rupsis:
Addresses https://github.com/RayTracing/raytracing.github.io/issues/1217 (also addresses #https://github.com/RayTracing/raytracing.github.io/issues/676)

Created a [Further-Readings](https://github.com/RayTracing/raytracing.github.io/wiki/Further-Readings) section in the wiki, and used an anchor tag to link to the relevant sections (Book 1, etc). 

Unfortunately the URL length is _almost_ longer than the style guide line length... Page seems to render just fine, but it could be worth exploring an alternative (URL shortener,  relative path?, etc)


hollasch:
Hey @trevordblack — this was your project; do you want to review this?

books/RayTracingInOneWeekend.html
```diff
@@ -66,8 +66,8 @@
 Graphics_ by Steve Marschner and Peter Shirley, or _Computer Graphics: Principles and Practice_
 by J.D. Foley and Andy Van Dam.
 
-Peter maintains a site related to this book series at <https://in1weekend.blogspot.com/>, which
-includes further reading and links to resources.
+Further readings and links to resources can found at the project's
```
armansito: > ...resources can found at...

can _be_ found at

books/RayTracingInOneWeekend.html
```diff
@@ -66,8 +66,8 @@
 Graphics_ by Steve Marschner and Peter Shirley, or _Computer Graphics: Principles and Practice_
 by J.D. Foley and Andy Van Dam.
 
-Peter maintains a site related to this book series at <https://in1weekend.blogspot.com/>, which
-includes further reading and links to resources.
+Further readings and links to resources can found at the project's 
+<a href="https://github.com/RayTracing/raytracing.github.io/wiki/Further-Readings#ray-tracing-in-one-weekend">wiki</a>.
```
armansito: You can use Markdown syntax here too:

```
[wiki](https://github.com/RayTracing/raytracing.github.io/wiki/Further-Readings#ray-tracing-in-one-weekend)
```

books/RayTracingTheNextWeek.html
```diff
@@ -21,8 +21,9 @@
 
 A heuristic in ray tracing that many people--including me--believe, is that most optimizations
 complicate the code without delivering much speedup. What I will do in this mini-book is go with the
-simplest approach in each design decision I make. Check <https://in1weekend.blogspot.com/> for
-readings and references to a more sophisticated approach. However, I strongly encourage you to do no
+simplest approach in each design decision I make. Check the project's
+ <a href="https://github.com/RayTracing/raytracing.github.io/wiki/Further-Readings#ray-tracing-the-next-week">wiki</a>
```
armansito: Since the wiki gets referenced twice in this file, consider declaring this at the bottom of the file (next to the links to stb_image and the authors' accounts):

```
[wiki]: https://github.com/RayTracing/raytracing.github.io/wiki/Further-Readings#ray-tracing-the-next-week
```
Then you can reference it like so:
```
...See the project's [wiki][] for pointers...
```

You could do this for the other files too if it helps with formatting.

armansito (commented):
Thank you for the change. Just some minor suggestions.

books/RayTracingTheNextWeek.html
hollasch: Yes, Arman's right. I much prefer descriptive links instead of bare links, and using the footnote link style helps keep the bare text readable. In addition, I always like widening URLs to provide both a more descriptive link as well as a fatter click target.

I'd do something like this:

```
See [our Further Reading wiki page][wiki-further] for additional references to more sophisticated approaches.

…

[wiki-further]: https://github.com/RayTracing/raytracing.github.io/wiki/Further-Readings
```

hollasch (changes requested):

hollasch:
Oh, also a general comment — let's just point to the wiki page right now without using anchor points. My gut feeling is that our further reading section should stand on its own, independent of the books. That is, rather than organizing the references by the current status of our book outline, group them by topic within ray tracing. This is especially true because the organization of our books seems to be under constant flux, and I'd rather not worry about keeping these synchronized as we move things around. Items which are very specific to book content should appear with the books themselves (perhaps as a footnote or other inline reference).

books/RayTracingTheNextWeek.html
rupsis: I'm hesitant to add `... for additional references to more sophisticated approaches.`, because (in my opinion) it sounds like the further readings wiki is limited to _only_ things related in the books.  

`See [our Further Reading wiki page][wiki-further] for additional resources and references to more sophisticated approaches.`
 would be fine, or 
`See [our Further Reading wiki page][wiki-further] for additional project related resources.`

This is a very minor point, and certainly not a hill I'm willing to die on :)

hollasch: Either of those sound fine.

hollasch (approved):

PR #1390, opened 2024-02-17
Simplify vec3 initialization

### LINKED ISSUE #847: expected member name or ';' after declaration specifiers
vscode is giving me a few weird errors in the vec3.h file:
```
expected` '(' gcc[13,25]
expected ';' after expression gcc[13,25]
expected member name or ';' after declaration specifiers gcc [13,27]
```

That is referring to the vec3 constructor

I've tried copy & pasting the code directly and it throws the same problem. 
I'm new to c++ (I want to learn a bit) so not completely sure about the syntax. I am wondering if this is to do with the c++ task.json or launch.json files? Should I avoid using vscode for c++ in general?

A bit of a turn off to c++ as a newbie! Gimme my c# back ha

```
#ifndef VEC3_H
#define VEC3_H

#include <cmath>
#include <iostream>

using std::sqrt;

class vec3 
{
    public:
        vec3() : e{0,0,0} {}
        vec3(double e0, double e1, double e2) : e{e0, e1, e2} {}

        double x() const { return e[0]; }
        double y() const { return e[1]; }
        double z() const { return e[2]; }

        vec3 operator-() const { return vec3(-e[0], -e[1], -e[2]); }
        double operator[](int i) const { return e[i]; }
        double& operator[](int i) { return e[i]; }

        vec3& operator+=(const vec3 &v) {
            e[0] += v.e[0];
            e[1] += v.e[1];
            e[2] += v.e[2];
            return *this;
        }

        vec3& operator*=(const double t) {
            e[0] *= t;
            e[1] *= t;
            e[2] *= t;
            return *this;
        }

        vec3& operator/=(const double t) {
            return *this *= 1/t;
        }

        double length() const {
            return sqrt(length_squared());
        }

        double length_squared() const {
            return e[0]*e[0] + e[1]*e[1] + e[2]*e[2];
        }
    public:
        double e[3];

};
```

hollasch:
We were using brace initialization like so:

    vec3() : e{0,0,0} {}
    vec3(double e0, double e1, double e2) : e{e0, e1, e2} {}

Some readers reported issues on Mac using braced initialization, so likely an issue with some compiler version. It may be technically legal with C++11 (I think, or perhaps a later C++ version), but it's not really buying anything, so this change uses an older-style initialization.

Resolves #847

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -310,8 +310,15 @@ vector utility functions in the bottom half:
       public:
         double e[3];
 
-        vec3() : e{0,0,0} {}
-        vec3(double e0, double e1, double e2) : e{e0, e1, e2} {}
+        vec3() {
+            e[0] = e[1] = e[2] = 0;
+        }
+
+        vec3(double e0, double e1, double e2) {
+            e[0] = e0;
+            e[1] = e1;
+            e[2] = e2;
+        }
 
         double x() const { return e[0]; }
         double y() const { return e[1]; }
--- a/src/InOneWeekend/vec3.h
+++ b/src/InOneWeekend/vec3.h
@@ -21,8 +21,15 @@ class vec3 {
   public:
     double e[3];
 
-    vec3() : e{0,0,0} {}
-    vec3(double e0, double e1, double e2) : e{e0, e1, e2} {}
+    vec3() {
+        e[0] = e[1] = e[2] = 0;
+    }
+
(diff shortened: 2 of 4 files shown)


rupsis (approved):
Approved on the technical side of things. 

My preference is to keep the bracket initialization (more concise). But considering this is the only place we use it, I'm fine either way.

hollasch:
Ok. Let's keep the brace initialization for now, and keep our ears open for any other people reporting this as an issue.

Issue #1327, opened 2023-11-04
The Next Week: The book says the ordering of bvh_node children doesn't matter, but acts like it does


DeltaPavonis:
In **Section 3: Bounding Volume Hierarchies** of  "Ray Tracing: The Next Week", the first example given of a BVH (3.2) notes that the order of left and right children in a `bvh_node` does not matter (emphasis mine):
> Note that the blue and red bounding volumes are contained in the purple one, but they might overlap, **and they are not ordered — they are just both inside**. So **the tree shown on the right has no concept of ordering in the left and right children**; they are simply inside.

However, in the subsequent code for splitting BVH volumes (3.9), a constraint seems to be imposed where the left child of a `bvh_node` always contains the objects which have axis value _less_ than the right child. Specifically, in the code

```cpp
        } else if (object_span == 2) {
            if (comparator(objects[start], objects[start+1])) {
                left = objects[start];
                right = objects[start+1];
            } else {
                left = objects[start+1];
                right = objects[start];
            }
        } else {
```
When the current `objects` list has only 2 objects, the given code always places the object with the lesser coordinate along the chosen axis in the left child. However, why can't we just always set `left = objects[start];` and `right = objects[start + 1];`, saving a call to `comparator`? Doesn't the ordering of the left and right children not matter, as stated before? If it does, then why does the ordering matter now?

hollasch:
Hmmm, well, you're right and you're wrong. Technically you can just arbitrarily divide the objects into two groups, and everything will still work. However, an arbitrary partition will yield bounding volumes that will likely overlap quite a bit. That's why in the next section it says that "… we need to make good divisions, rather than bad ones, …". But this needs to be reworded because we "make good divisions" in the interest of _speed_, as "bad" divisions will still yield functional results. So it looks like we need to clarify the text here.

**_But_**, the case with only two objects is special. With many objects, ordering them in one dimension will yield a tighter cluster, and thus smaller and minimally overlapping boxes for the two groups. With only _two_ objects, though, you're absolutely correct — dividing is trivial, and the two boxes will be as small as can be. Order doesn't matter, as you point out. The code should be simplified exactly as you say.

hollasch:
The _code_ portion of this issue should be included in PR #1391.

The rewriting work I mention above will be included in a separate change.

hollasch:
@DeltaPavonis — Nice catch, by the way!

hollasch:
Done.

Issue #1388, opened 2024-02-13
Faster and simpler BVH construction


relief7:
Hi!

first of all, I wanted to say a huge thank you to Peter Shirley for creating this fantastic tutorial series. It is amazing how intuitive and simple things are in his explanations and it is so much fun to follow. It is also fascinating what amazing results you get with so little code. I have been reading the series with great enthusiasm and already learned so much from it!

Also thank you to everyone here on github for all the maintenance work and valuable improvements to the code and the documentation. I can imagine it is hard work to keep everything updated and in sync so it is really greatly appreciated! :+1:

The reason why I am writing ... (I actually just signed up at github to try to comment here :smile:)

I have followed part 1 and 2 of the series, then spent some time adding a few additional features on my own (such as triangle mesh hittables, a different material etc.). Then my goal was to render the Stanford dragon test model. It is a triangle mesh and has roughly 900k triangles and that is when I suddenly hit a pretty severe bottleneck in the code. It was related to the bvh construction that it was impossible to even start rendering any pixels.

The problem seems that in every recursion step of the bvh construction, we make a copy of the entire vector of hittable pointers, even though we only need to process a handful of objects of those in the actual recursion step. So in my example, the pointers for all 900k triangles are copied in every split which led to the bvh construction taking forever (I cancelled it after a couple of hours). 

After taking a closer look at the bvh construction code, I believe there is a way of optimizing the current code but at the same time making it easier to understand for beginners as well. This is why I am writing here. Because I think there is an actual benefit to review the current bvh code and make some changes. It's nothing crazy, only changing a few lines and you gain not only a huge speedup for potential large amounts of hittables but also it is a bit more intuitive as well (in my humble opinion). In the case of my dragon model, the bvh construction was reduced from many hours to ~10 seconds, yay! :smile:

The proposed change also eliminates the tracking of the start and end indices and divides the input array before each recursion level. This means that only the really needed objects are passed to the next level. In a true "divide and conquer" fashion. This avoids copying data that is not really needed, the array gets smaller and smaller on each level and the speedup is huge.

As an example, let's say you have 24 objects in your scene. In the current code, all 24 objects get passed down in every step down to the leaf construction level:

![bvh_01](https://github.com/RayTracing/raytracing.github.io/assets/159066453/ff510e7f-e280-4f4f-956c-b3b48035875d)

In the proposed new code, we would process 24, then split the array, then process 2x12, split, then process 4x6, split, process 8x3 etc.

[...]

PR #1391, opened 2024-02-19
Proposal for new bvh split method (approach Nr. 3) - divide and conquer with std::sort()

relief7:
Hi Steve!

please find proposal for "new bvh split method (approach Nr. 3) - divide and conquer with std::sort()" based on latest dev branch in this pull request (see further info in issue reply).

Thanks,
Markus

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -347,9 +347,8 @@ location.
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
             sqrt_spp = int(sqrt(samples_per_pixel));
             recip_sqrt_spp = 1.0 / sqrt_spp;
-
-
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+
             // Calculate the u,v,w unit basis vectors for the camera coordinate frame.
             w = unit_vector(lookfrom - lookat);
             ...


src/TheNextWeek/bvh.h
```diff
@@ -22,12 +22,15 @@
...
-    bvh_node(const std::vector<shared_ptr<hittable>>& src_objects, size_t start, size_t end) {
+    bvh_node(const std::vector<shared_ptr<hittable>>& src_objects) {
         // Build the bounding box of the span of source objects.
         bbox = aabb::empty;
-        for (size_t object_index=start; object_index < end; object_index++)
+
+        size_t size = src_objects.size ();
+
+        for (size_t object_index=0; object_index < size; object_index++)
             bbox = aabb(bbox, src_objects[object_index]->bounding_box());
```
hollasch: Simplify to:

    for (auto& obj : src_objects)
        bbox = aabb(bbox, obj->bounding_box());

src/TheNextWeek/bvh.h
```diff
@@ -22,12 +22,15 @@
...
-    bvh_node(const hittable_list& list) : bvh_node(list.objects, 0, list.objects.size()) {}
+    bvh_node(const hittable_list& list) : bvh_node(list.objects) {}
 
-    bvh_node(const std::vector<shared_ptr<hittable>>& src_objects, size_t start, size_t end) {
+    bvh_node(const std::vector<shared_ptr<hittable>>& src_objects) {
         // Build the bounding box of the span of source objects.
         bbox = aabb::empty;
-        for (size_t object_index=start; object_index < end; object_index++)
+
+        size_t size = src_objects.size ();
```
hollasch: Move `size` definition down below the line:

    auto objects = src_objects; // A modifiable array of the source scene objects

src/TheNextWeek/bvh.h
```diff
@@ -38,24 +41,25 @@ class bvh_node : public hittable {
...
+        } else if (size == 2) {
+            if (comparator(objects[0], objects[1])) {
+                left = objects[0];
+                right = objects[1];
             } else {
-                left = objects[start+1];
-                right = objects[start];
+                left = objects[1];
+                right = objects[0];
             }
```
hollasch: Since we're rewriting this, let's fix issue #1327 in the same commit. Please simplify this case to the following:

    ...
    } else if (size == 2) {
        left = objects[0];
        right = objects[1];
    } else {
    ...

(Ordering the two child objects in this case accomplishes nothing.)

hollasch (changes requested):
If you agree to these changes, then we also need to do the following work:
- Make the same changes to bvh.h in `src/TheRestOfYourLife`.
- Update the text and any listings for books 2 & 3
- Add an entry to the CHANGELOG.md file.
- Please add your name to `books/acknowledgments.md`.

relief7:
Hi Steve!
Thank you for the great feedback! I totally forgot about the code in the other chapters and the book! 🤣 
Will try to do another round and re-submit! After all these changes, the bvh construction will be so easy that even a pre-schooler can understand it. 😄  
Thanks,
Markus

wlbksy:
You can improve further by writing a helper function to do the job.
1. `bvh_node(const vector& objs)` first sort all the objects.
2. call the helper function to do the split job so that you don't have to sort every level down the tree.

This reduce the current $O(n\lg^2 n)$ to $O(n\lg n)$

relief7:
Hi @wlbksy!
Thank you for your suggestion! I am not sure if I can follow though... Don't we need to sort in every level because we are randomly switching up the axis of sorting in each level?

```
int axis = random_int(0,2);
auto comparator = (axis == 0) ? box_x_compare
                : (axis == 1) ? box_y_compare
                              : box_z_compare;
```

So a good sort for x in the beginning might not be an optimal sort for z a few levels further down, for example. Or am I misunderstanding this issue?

armansito:
> Hi @wlbksy! Thank you for your suggestion! I am not sure if I can follow though... Don't we need to sort in every level because we are randomly switching up the axis of sorting in each level?
> 
> ```
> int axis = random_int(0,2);
> auto comparator = (axis == 0) ? box_x_compare
>                 : (axis == 1) ? box_y_compare
>                               : box_z_compare;
> ```
> 
> So a good sort for x in the beginning might not be an optimal sort for z a few levels further down, for example. Or am I misunderstanding this issue?

I agree, you need to know the spatial subdivision axis and position at each level  and the BVH construction code is already handling this.

As an alternative, if we really wanted to eliminate all the redundant copies and vector allocations, we could allow the BVH construction to sort the object list in-place without creating any intermediate vectors. The code would look very similar to the original listing but instead remove the `const` qualifier from the `src_objects` declaration. If preserving the original order of the input objects is important, the input could be copied _once_ and the in-place sort could operate over that initial copy (but I'm not sure if that's really necessary).

@hollasch any thoughts?

relief7:
Hi @armansito 

I personally think that this is a great idea! It would probably be super fast and efficient. 

I always wondered why we are doing the indexing *and* copying the vector in the first place because doing both kind of negate each other.

My proposal works around the issue to simplify the indexing (because it would not be needed) but what you are suggesting is probably the most elegant solution with the smallest RAM and CPU footprint. 😄

wlbksy:
> Hi @wlbksy! Thank you for your suggestion! I am not sure if I can follow though... Don't we need to sort in every level because we are randomly switching up the axis of sorting in each level?
> 
> ```
> int axis = random_int(0,2);
> auto comparator = (axis == 0) ? box_x_compare
>                 : (axis == 1) ? box_y_compare
>                               : box_z_compare;
> ```
> 
> So a good sort for x in the beginning might not be an optimal sort for z a few levels further down, for example. Or am I misunderstanding this issue?

Oh, you're right. I missed this random axis switching.

relief7:
Hi! 
apologies for closing and opening a new pull request (still trying to get the hang of github 😆).

I made some tests with your suggestions, @armansito (see results in my previous bvh thread #1388 ).

IMHO, this would be the most elegant solution to most issues, so a new bvh method such as the one from my original proposal would not really be needed anymore. There is also a pull request #1409 with corresponding code and book changes, Steve @hollasch , in case you find it useful.

Thanks!

hollasch:
> @[relief7](https://github.com/relief7) relief7 [force-pushed](https://github.com/RayTracing/raytracing.github.io/compare/7e62d8e94d20007a2c1b37c5dc43149da9db9d8f..b11548c7783a6dd1a3dfdb50192b4a3959608f77) the dev branch from [7e62d8e ](https://github.com/RayTracing/raytracing.github.io/commit/7e62d8e94d20007a2c1b37c5dc43149da9db9d8f)to [b11548c ](https://github.com/RayTracing/raytracing.github.io/commit/b11548c7783a6dd1a3dfdb50192b4a3959608f77)
[7 hours ago](https://github.com/RayTracing/raytracing.github.io/pull/1391#event-11923065891)

You didn't force-push the `RayTracing:dev` branch did you?

relief7:
> > @[relief7](https://github.com/relief7) relief7 [force-pushed](https://github.com/RayTracing/raytracing.github.io/compare/7e62d8e94d20007a2c1b37c5dc43149da9db9d8f..b11548c7783a6dd1a3dfdb50192b4a3959608f77) the dev branch from [7e62d8e ](https://github.com/RayTracing/raytracing.github.io/commit/7e62d8e94d20007a2c1b37c5dc43149da9db9d8f)to [b11548c ](https://github.com/RayTracing/raytracing.github.io/commit/b11548c7783a6dd1a3dfdb50192b4a3959608f77)
> > [7 hours ago](https://github.com/RayTracing/raytracing.github.io/pull/1391#event-11923065891)
> 
> You didn't force-push the `RayTracing:dev` branch did you?

No, as you can see in the status message:

image

I am only modifying the ```dev``` branch in my own fork and submitting pr's from there. I wouldn't even have the administrative rights to commit any changes to your repository, would I?

hollasch:
_Phew!_ Thanks @relief7.

PR #1393, opened 2024-02-20
Fix noise_texture regression in call to turb()

### LINKED ISSUE #1286: [Book2] Incorrect marble pattern at the final Perlin noise section
Hi, I'm implementing this code in Rust instead of CPP so I very much hope that this isn't due to language differences.

## The issue
Anyways, in the Perlin Noise chapter at the marble pattern section the book proposes this modification to get the marble texture:
```c++
    auto s = scale * p;
    return color(1,1,1) * 0.5 * (1 + sin(s.z() + 10*noise.turb(s)));  // << this line
}
```
Here is my equivalent Rust code:
```rust
fn value(&self, _: f64, _: f64, point: &Point3) -> Color {
    let scaled_point = self.scale * (*point);
    Color::new(1.0, 1.0, 1.0) * 0.5 * (1.0 + (scaled_point.z() + 10.0 * self.noise.turbulence(scaled_point, 7)).sin())
}
```
And this is the image produced:
marble_sphere_bugged

This is doesn't really look like marble, and does not resemble the following example from the book:
from-the-book

## Potential solution
I've played around with the code, and noticed that if I pass the raw point (unscaled) to the turbulence function like so:
```rust
fn value(&self, _: f64, _: f64, point: &Point3) -> Color {
    let scaled_point = self.scale * (*point);
    Color::new(1.0, 1.0, 1.0) * 0.5 * (1.0 + (scaled_point.z() + 10.0 * self.noise.turbulence(*point, 7)).sin()) // << *point instead of scaled_point
}
```
I get the following image:
marble_sphere_correct

Which appears much closer to the intended result.  
I cannot say I understand the math behind this chapter very well, so I don't know if this is the correct solution, but wanted to share anyway.

Hope this is helpful!

hollasch:
At some point we passed the scaled point to the noise.turb() function, when we should have continued to pass the unscaled point.

Resolves #1286

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2395,8 +2395,7 @@ Here `fabs()` is the absolute value function defined in `<cmath>`.
 
         color value(double u, double v, const point3& p) const override {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            auto s = scale * p;
-            return color(1,1,1) * noise.turb(s, 7);
+            return color(1,1,1) * noise.turb(p, 7);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
         }
 
@@ -2433,8 +2432,7 @@ effect is:
 
         color value(double u, double v, const point3& p) const override {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            auto s = scale * p;
-            return color(1,1,1) * 0.5 * (1 + sin(s.z() + 10*noise.turb(s, 7)));
+            return color(1,1,1)*0.5*(1 + sin(scale * p.z() + 10*noise.turb(p, 7)));
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
         }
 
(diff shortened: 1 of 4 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -2433,8 +2432,7 @@
 
         color value(double u, double v, const point3& p) const override {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            auto s = scale * p;
-            return color(1,1,1) * 0.5 * (1 + sin(s.z() + 10*noise.turb(s, 7)));
+            return color(1,1,1)*0.5*(1 + sin(scale * p.z() + 10*noise.turb(p, 7)));
```
armansito: I personally prefer to keep the spaces between the `*` operands. I would prefer to at least stay consistent across multiplications in a compound statement like this (having `scale * p.z()` and `10*noise.turb(p, 7))` together appears a little jarring to me).

I'll defer to you on the broader style, however.

src/TheNextWeek/main.cc
```diff
@@ -364,7 +364,7 @@ void final_scene(int image_width, int samples_per_pixel, int max_depth) {
 
     auto emat = make_shared<lambertian>(make_shared<image_texture>("earthmap.jpg"));
     world.add(make_shared<sphere>(point3(400,200,400), 100, emat));
-    auto pertext = make_shared<noise_texture>(0.1);
+    auto pertext = make_shared<noise_texture>(0.20);
```
armansito: `0.2` instead of `0.20`? (The former is more consistent with the surround code).

src/TheNextWeek/texture.h
```diff
@@ -76,8 +76,7 @@ class noise_texture : public texture {
     noise_texture(double sc) : scale(sc) {}
 
     color value(double u, double v, const point3& p) const override {
-        auto s = scale * p;
-        return color(1,1,1)*0.5*(1 + sin(s.z() + 10*noise.turb(s, 7)));
+        return color(1,1,1)*0.5*(1 + sin(scale * p.z() + 10*noise.turb(p, 7)));
```
armansito: Same question about adding spaces between `*` operands.

src/TheRestOfYourLife/texture.h
```diff
@@ -76,8 +76,7 @@ class noise_texture : public texture {
     noise_texture(double sc) : scale(sc) {}
 
     color value(double u, double v, const point3& p) const override {
-        auto s = scale * p;
-        return color(1,1,1)*0.5*(1 + sin(s.z() + 10*noise.turb(s, 7)));
+        return color(1,1,1)*0.5*(1 + sin(scale * p.z() + 10*noise.turb(p, 7)));
```
armansito: Same question about adding spaces between `*` operands.

armansito (commented):
Thank you for fixing this issue. Just some minor formatting nits.

I believe images 2.13-2.15 might need to get updated to look like their previous versions? I haven't run the code changes locally but that's my understanding from the discussion on #1286.

hollasch:
Note that images will all be updated as necessary in #779.

armansito (approved):
Looks good to me. Acknowledging that the images will get fixed as part of #779

PR #1404, opened 2024-02-23
[Book 2] Fix code listings to reflect the removed `aabb::pad()` method

### LINKED ISSUE #1403: [Book 2] aabb.pad() has been renamed to pad_to_minimums in codes, but writing is not consistent to the code
`aabb.pad()` has been renamed to `pad_to_minimums` in codes, but writing is not consistent to the code.

This need some writing and listed codes change in book2 section6.1.

I'll leave this to people who has time to fix it.

armansito:
Fixed some discontinuities in the code listings regarding implicit padding during AABB construction to avoid zero volume cases, which were leftover following the refactor in #1238.

Instead of introducing `pad_to_minimums()` during the first definition of the `aabb` class, I moved it to later where `aabb::pad()` originally got introduced. I felt that this goes better with the existing narrative that motivates the padding when talking about quads.

Fixes #1403

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -2473,7 +2473,7 @@ The following figure illustrates the quadrilateral components.
 
 <div class='together'>
 Quads are flat, so their axis-aligned bounding box will have zero thickness in one dimension if the
-quad lies in the XY, YZ, or XZ plane. This can lead to numerical problems with ray intersection, but
+quad lies in the XY, YZ, or ZX plane. This can lead to numerical problems with ray intersection, but
 we can address this by padding any zero-sized dimensions of the bounding box. Padding is fine
 because we aren't changing the intersection of the quad; we're only expanding its bounding box to
 remove the possibility of numerical problems, and the bounds are just a rough approximation to the
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -592,7 +592,7 @@ given amount:
 </div>
 
 <div class='together'>
-Now we have everything we need to implment the new AABB class.
+Now we have everything we need to implement the new AABB class.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     #ifndef AABB_H
@@ -607,10 +607,7 @@ Now we have everything we need to implment the new AABB class.
         aabb() {} // The default AABB is empty, since intervals are empty by default.
 
         aabb(const interval& _x, const interval& _y, const interval& _z)
-          : x(_x), y(_y), z(_z)
-        {
-            pad_to_minimums();
-        }
+          : x(_x), y(_y), z(_z) {}
 
         aabb(const point3& a, const point3& b) {
             // Treat the two points a and b as extrema for the bounding box, so we don't require a
@@ -618,8 +615,6 @@ Now we have everything we need to implment the new AABB class.
             x = interval(fmin(a[0],b[0]), fmax(a[0],b[0]));
             y = interval(fmin(a[1],b[1]), fmax(a[1],b[1]));
             z = interval(fmin(a[2],b[2]), fmax(a[2],b[2]));
-
-            pad_to_minimums();
         }
 
         const interval& axis(int n) const {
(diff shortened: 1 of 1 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -2486,36 +2473,57 @@
...
-        aabb pad() {
-            // Return an AABB that has no side narrower than some delta, padding if necessary.
-            double delta = 0.0001;
-            interval new_x = (x.size() >= delta) ? x : x.expand(delta);
-            interval new_y = (y.size() >= delta) ? y : y.expand(delta);
-            interval new_z = (z.size() >= delta) ? z : z.expand(delta);
+            pad_to_minimums();
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+        }
+        ...
```
wlbksy, line 2508: `aabb(const aabb& box0, const aabb& box1)` should call `pad_to_minimums()` at its end too.

armansito: I intentionally left the call out of that constructor, since I think we can assume the input `aabb` parameters already have a non-zero volume and so should the result of combining their intervals. Though, I might have overlooked a corner case and I'd be happy to revisit this.

wlbksy: Second thought, you're right on this.

hollasch: Agreed with @armansito.

books/RayTracingTheNextWeek.html
```diff
@@ -607,19 +607,14 @@
         aabb() {} // The default AABB is empty, since intervals are empty by default.
 
         aabb(const interval& _x, const interval& _y, const interval& _z)
-          : x(_x), y(_y), z(_z)
-        {
-            pad_to_minimums();
-        }
+          : x(_x), y(_y), z(_z) {}
```
hollasch, line 610: This is a mistake. Any input interval could be zero-sized. For example, passing an axis-aligned polygon would give you a singular interval in that dimension.

books/RayTracingTheNextWeek.html
```diff
@@ -607,19 +607,14 @@
...
 
         aabb(const point3& a, const point3& b) {
             // Treat the two points a and b as extrema for the bounding box, so we don't require a
             // particular minimum/maximum coordinate order.
             x = interval(fmin(a[0],b[0]), fmax(a[0],b[0]));
             y = interval(fmin(a[1],b[1]), fmax(a[1],b[1]));
             z = interval(fmin(a[2],b[2]), fmax(a[2],b[2]));
-
-            pad_to_minimums();
```
hollasch, line 622: If a = b, then we will need to call `pad_to_minimums()`.

books/RayTracingTheNextWeek.html
```diff
@@ -2486,36 +2473,57 @@
 
 <div class='together'>
 Quads are flat, so their axis-aligned bounding box will have zero thickness in one dimension if the
-quad lies in the XY, YZ, or ZX plane. This can lead to numerical problems with ray intersection, but
+quad lies in the XY, YZ, or XZ plane. This can lead to numerical problems with ray intersection, but
```
hollasch: Please keep as "ZX" (right-hand rule).

books/RayTracingTheNextWeek.html
hollasch: There's nothing that guarantees that the input intervals are non-singular, nor that a ≠ b. Agreed with @wlbksy.

hollasch (changes requested):

hollasch:
Oh good grief. I missed the timing aspect of your fix. Will review again.

books/RayTracingTheNextWeek.html
hollasch: Me saying it's a mistake is a mistake.

books/RayTracingTheNextWeek.html
hollasch: Disregard.

books/RayTracingTheNextWeek.html
hollasch: Ugh. I'm batting 0 for 1000 today. Yes, since the other constructors ensure that all `aabb`s have non-zero dimensions, we are safe when composing such `aabb` objects.

I agree with @armansito's original reasoning and change.

hollasch:
Ok, @armansito's changes look good, except for changing "ZX" to "XZ". I'll just add a quick commit and merge.

hollasch (approved):

books/RayTracingTheNextWeek.html
wlbksy: > Ugh. I'm batting 0 for 1000 today. Yes, since the other constructors ensure that all `aabb`s have non-zero dimensions, we are safe when composing such `aabb` objects.
> 
> I agree with @armansito's original reasoning and change.

Yeah. Original is sound. But since we two all gets confused here at first glimpse, maybe adding the `pad_to_minimums()` call here is not too overkill, and ease reader's mind.

Issue #1370, opened 2024-01-27
[Book 3] Section 2.3 Stratified Samples cause color loss


bit704:
In camera.h:
```c++
for (int j = 0; j < image_height; ++j) {
      std::clog << "\rScanlines remaining: " << (image_height - j) << ' ' << std::flush;
      for (int i = 0; i < image_width; ++i) {
          color pixel_color(0,0,0);
          for (int s_j = 0; s_j < sqrt_spp; ++s_j) {
              for (int s_i = 0; s_i < sqrt_spp; ++s_i) {
                  ray r = get_ray(i, j, s_i, s_j);
                  pixel_color += ray_color(r, max_depth, world);
              }
          }
          write_color(std::cout, pixel_color, samples_per_pixel);
      }
  }
```
We can see that `ray_color` have been called `sqrt_spp`\*`sqrt_spp` times, which is small than `samples_per_pixel` when the latter isn't a perfect square.
But in `write_color`, `scale = 1.0 / samples_per_pixel`, so color loss happens.
Should use `scale = 1.0 / (sqrt_spp * sqrt_spp)`.
`write_color` in color.h:
```c++
void write_color(std::ostream &out, color pixel_color, int samples_per_pixel) {
    auto r = pixel_color.x();
    auto g = pixel_color.y();
    auto b = pixel_color.z();

    // Divide the color by the number of samples.
    auto scale = 1.0 / samples_per_pixel;
    r *= scale;
    g *= scale;
    b *= scale;

    // Apply a linear to gamma transform for gamma 2
    r = linear_to_gamma(r);
    g = linear_to_gamma(g);
    b = linear_to_gamma(b);

    // Write the translated [0,255] value of each color component.
    static const interval intensity(0.000, 0.999);
    out << static_cast<int>(256 * intensity.clamp(r)) << ' '
        << static_cast<int>(256 * intensity.clamp(g)) << ' '
        << static_cast<int>(256 * intensity.clamp(b)) << '\n';
}
```

hollasch:
Excellent catch!

We should in fact pass `scale` to the `write_color()` function instead of `samples_per_pixel`, so we don't have to calculate the reciprocal over and over again as well.

Sadly, we should probably make this change retroactive for the `write_color` function and render loop throughout the books. Sigh.

hollasch:
I'm putting this on the alpha.2 milestone since it will affect all three books.

bit704:
[...]

PR #1405, opened 2024-02-23
[All Books] Use 1 / (sqrt_spp * sqrt_spp) for sample scale. 

rupsis:
Addresses #1370 

From the naked eye, the color difference isn't noticeable:

| 1.0 / samples_per_pixel  | 1.0 / (sqrt_spp * sqrt_spp) |
| ------------- | ------------- |
| IOWDefualtColor  | IOWNoColorLoss  |

Differencing the two images, we see that there is in fact a _slight_ increase in value:
value_difference

Highlighted difference:
highlighted_difference

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -1984,9 +1984,9 @@ pixel and average the resulting light (color) values together.
 
 First we'll update the `write_color()` function to account for the number of samples we use: we need
 to find the average across all of the samples that we take. To do this, we'll add the full color
-from each iteration, and then finish with a single division (by the number of samples) at the end,
-before writing out the color. To ensure that the color components of the final result remain within
-the proper $[0,1]$ bounds, we'll add and use a small helper function: `interval::clamp(x)`.
+from each iteration, and then finish with a single division (by the number of samples in the unit square)
+at the end, before writing out the color. To ensure that the color components of the final result remain
+within the proper $[0,1]$ bounds, we'll add and use a small helper function: `interval::clamp(x)`.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class interval {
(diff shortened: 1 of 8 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -2011,19 +2011,18 @@
...
 
-        // Divide the color by the number of samples.
-        auto scale = 1.0 / samples_per_pixel;
-        r *= scale;
-        g *= scale;
-        b *= scale;
+        // Divide the color by the sampling scale.
+        r *= spp_scale;
+        g *= spp_scale;
+        b *= spp_scale;
```
rupsis, line 2025: Without a scale definition, the comment "Divide the color by the sampling scale." 
seems a _bit_ misleading.

books/RayTracingInOneWeekend.html
```diff
@@ -2072,8 +2073,21 @@
         }
         ...
       private:
-        ...
+        int    image_height;    // Rendered image height
+        point3 center;          // Camera center
+        point3 pixel00_loc;     // Location of pixel 0, 0
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
+        int    sqrt_spp;        // Square Root of random samples for each pixel
+        double spp_scale;       // Scale value for the random samples for each pixel
```
rupsis, line 2081: Following the naming used in Book 3, this should be `recip_sqrt_spp_sqrd`, but I find that cumbersome. Not completely convinced on this name however.

src/TheRestOfYourLife/color.h
```diff
@@ -35,13 +35,12 @@ void write_color(std::ostream& out, const color& pixel_color, int samples_per_pi
...
-    r *= scale;
-    g *= scale;
-    b *= scale;
+    // Divide the color by the sampling scale.
+    r *= spp_scale;
+    g *= spp_scale;
+    b *= spp_scale;
 
-    // Apply a linear to gamma transform for gamma 2
+    // Apply the linear to gamma transform.
```
rupsis, line 43: Doesn't appear that `linear_to_gamma` changes between books, i assume we'd want consistency in code comments? 

Perhaps better addressed in it's own issue.

books/RayTracingInOneWeekend.html
armansito: How about simply `sample_scale`? Maybe how this value gets computed doesn't really need to be conveyed by the variable name besides it being a scale that we apply to samples (as the comment explains).

Perhaps the specific logic for computing it could be explained by an inline code comment and/or the book text surrounding the listing?

hollasch:
Yuck. This change continues to grow the blob of private state variables computing in `camera::initialize()` and then used in `camera::render()`. Your change continues the pattern I started, which is good for the scope of this change, but I'm really regretting the artificial boundary between these two functions. All of this stuff should just be part of the `camera::render()` function. I'll create a new issue to track this hopeful refactor and toss it on the backlog.

Although a lot of code uses setters and getters instead of exposing public variables (which we deliberately chose to simplify the code). If we went with that approach (`camera::set_pixel_sample_count(int desiredCount)`, then we'd have to have the glob of private variables. Perhaps another thing to consider.

hollasch:
You know what? Now that I'm thinking about it, the multi-sampling and subsequent normalization is really the job of the camera, and the `write_color()` function should not be in the business of managing it. If you don't mind, I'll tag on a commit to show you what I'm thinking.

hollasch:
I just realized that you implemented stratified sampling for all three books — that's a mistake. Stratified color sampling is introduced in book 3, section 2.3.

rupsis:
@hollasch 
> I just realized that you implemented stratified sampling for all three books — that's a mistake. Stratified color sampling is introduced in book 3, section 2.3.

My mistake, from your comment on the issue it seemed like it was a change to all of the books: 
> Sadly, we should probably make this change retroactive for the write_color function and render loop throughout the books. Sigh.

So in this case, we should just update Book 3?

hollasch:
I'm working on a revised PR based on your work. Unfortunately, I've also stumbled on a bug in the `camera::pixel_sample_square()` function. Will post soon for your comment.

rupsis:
@hollasch  Sounds good! 
Curious to know what the bug is.

rupsis:
The work in #1417 refactors, and fulfills the requirements. Closing.

PR #1416, opened 2024-03-04
Clarify the new BVH constructor behavior

hollasch:


rupsis (approved):
Slight suggestion:

```
// There's a C++ subtlety here. This constructor (taking a hittable list without span
// indices) creates an implicit copy of the hittable list, which we will modify. The
// lifetime of the copied list only extends until this constructor exits. That's OK
// because all we need to persist is the resulting bounding volume hierarchy.
 ```
          
"of this copied list" -> "of the copied list".  `This` is already referring to the constructor. 

Last sentence is a bit of a run-on. Splitting it.

PR #1417, opened 2024-03-05
Remove pixel sampling knowledge from write_color

hollasch:
The prior version was passed the number of samples for the summed pixel color, which it then had to divide out in order to derive the average sample color.

This change simplifies the write_color function to take only a color, requiring the caller to compute any desired final color.

In the service of this, the `camera` class now caches the inverse of the number of pixel samples, so it can compute the average sample color before passing the result to the now simplified write_color() function.

In addition, I've factored out some steps in the final color value computation to make each phase more discrete and easier to grok.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -424,10 +424,17 @@ that writes a single pixel's color out to the standard output stream.
     using color = vec3;
 
     void write_color(std::ostream& out, const color& pixel_color) {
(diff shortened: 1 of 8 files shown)


hollasch:
As a side note, after this work is done, I plan to rework & refactor the `camera::pixel_sample_square()` function, as I believe it's incorrect.

armansito (approved):
The changes overall look good to me.

rupsis (approved):
Changes make sense, and simplify the code.

PR #1418, opened 2024-03-06
correct call to reflectance()

TheThief:
reflectance() takes `double ref_idx` (which is `ir` in the outer scope), not `refraction_ratio`. Confirmed with the formula for Schlick's approximation (https://en.wikipedia.org/wiki/Schlick%27s_approximation) which uses two indexes of refraction, of which one is hard coded to `1` in the function here.

A future todo would be to keep track of the "ir of the current medium" in the ray and use that instead of a hard coded `1` in the reflectance formula.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -3447,7 +3447,7 @@ material:
             vec3 direction;
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            if (cannot_refract || reflectance(cos_theta, refraction_ratio) > random_double())
+            if (cannot_refract || reflectance(cos_theta, ir) > random_double())
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
                 direction = reflect(unit_direction, rec.normal);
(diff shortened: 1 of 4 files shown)


TheThief:
I previously mentioned this here: https://github.com/RayTracing/raytracing.github.io/issues/1278#issuecomment-1788597280 but it took me a little while to get around to making the pull request

TheThief:
Note: you'd think this would end up calculating reflectance wrong when leaving the sphere or with the inverted sphere, thinking that maybe it should be `ref_idx-1` instead of `1-ref_idx` (with the two ir's the other way round in the Schlick formula) but as it immediately squares it these actually end up equivalent and it doesn't matter.

hollasch (changes requested):
This change lacks an update to book 3.

Also, please update CHANGES.md.

hollasch:
If you would like, add an entry for yourself in books/acknowledgments.md.html.

hollasch:
I've gone ahead and made the changes. See PR #1426.

hollasch:
@TheThief — let me know if you'd like an entry in the acknowledgments.

TheThief:
> @TheThief — let me know if you'd like an entry in the acknowledgments.

@hollasch If you would be so kind, you could credit me in the Acknowledgements as "Gareth Martin"

I'm sorry I didn't get the changes made myself, I've been busy

TheThief:
I see your link to #1420 and I think this fix breaks that idea - the Schlick formula specifically needs to know the two IRs, not the ratio, so representing the inner bubble of air as 1/ir isn't helpful in general as we need 1 and the ir for the Schlick formula. It's possible to detect an ir<1.0 and invert it - but that only works in specifically the case of air inside another medium, not e.g. _water_ inside the glass sphere.

PR #1421, opened 2024-03-09
Refactor pixel subsampling

hollasch:
- Simplify pixel sampling to make the sampling functions simpler and agnostic about their use as pixel sampling functions. Now they sample an idealized unit square or disk and are ignorant about image components.
- pixel_sample_xxx() functions renamed to sample_xx().
- Rename pixel_sample_scale to pixel_samples_scale.
- Reworded some comments for clarity.
- I really wish I could have come up with a better rename for `sqrt_spp` and `recip_sqrt_spp`, but I've been unable to think of anything better.

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -358,8 +358,8 @@ location.
         ...
         ray get_ray(int i, int j, int s_i, int s_j) const {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-            // Get a randomly-sampled camera ray for the pixel at location i,j, originating from
-            // the camera defocus disk, and randomly sampled around the pixel location.
+            // Construct a camera ray originating from the defocus disk and directed at a randomly
+            // sampled point around the pixel location i, j for stratified sample square s_i, s_j.
 
             auto offset = sample_square_stratified(s_i, s_j);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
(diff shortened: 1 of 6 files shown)


books/RayTracingTheRestOfYourLife.html
```diff
@@ -348,7 +348,7 @@
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
             sqrt_spp = int(sqrt(samples_per_pixel));
-            pixel_sample_scale = 1.0 / (sqrt_spp * sqrt_spp);
+            pixel_samples_scale = 1.0 / (sqrt_spp * sqrt_spp);
```
armansito, line 351: A part of me wants to rewrite this line as `pixel_samples_scale = recip_sqrt_spp * recip_sqrt_spp;` but I'm not sure if the opaqueness is worth it.

books/RayTracingTheRestOfYourLife.html
```diff
@@ -357,14 +357,15 @@
         }
         ...
         ray get_ray(int i, int j, int s_i, int s_j) const {
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
             // Get a randomly-sampled camera ray for the pixel at location i,j, originating from
```
armansito: This isn't directly related to this PR but the "randomly sampled" part here feels a little redundant, how about rephrasing this sentence to something like:

```
// Build a camera ray originating from the camera defocus disk and directed at
// a randomly sampled point around the pixel at location i,j.
```

armansito (approved):
Looks good. I left a couple of suggestions that I don't feel particularly strong about, so feel free to ignore them.

> - I really wish I could have come up with a better rename for sqrt_spp and recip_sqrt_spp, but I've been unable to think of anything better.

I think these names are fine.

books/RayTracingTheRestOfYourLife.html
hollasch: Heh, I had the same thought. It was the opaqueness that made me decide against it. I'll keep it as is.

books/RayTracingTheRestOfYourLife.html
hollasch: Will do.

PR #1422, opened 2024-03-10
Interval-based AABB update / Attempt #2

### LINKED ISSUE #927: AABB hit and NaNs & Infinities
_(Disclaimer: my implementation of this project is in Rust, and my experience with C++ is limited. I am linking to reference docs just to make sure I have verified my own understanding - not to imply that the authors of this book & maintainers of this repo wouldn't know! Feel free to point out any inaccuracies and mistakes - I'm here to learn. Thanks!)_

Looking at `dev-major` branch, to be most up to date.

Given the function

https://github.com/RayTracing/raytracing.github.io/blob/62ce2e6a1f08b4c103808e513dc1c8fc6a188776/src/common/aabb.h#L61-L73

and an example ray with an origin of `[0.0, 0.0, -1.0]` and direction `[0.0, 0.0, 1.0]` - or anything similar for that matter, where any component of the direction vector is zero - you will end up getting a `NaN` from the division by `r.direction()[a]`.

On a quick look at the [reference](https://en.cppreference.com/w/c/numeric/math/fmin), `fmin(NaN, other)` and `fmax(NaN, other)` calls will return `other`, unless both operands are `NaN`, in which case a `NaN` is returned. This will mean that `t0` and `t1` will get set to `NaN`. Additionally, `ray_t.min` and `ray_t.max` will get set to what they already are, and the comparison of `ray_t.max <= ray_t.min` will be falsy so the loop continues to the next iteration.

This could potentially lead to unintended results, with false positives on whether the AABB can be hit with the ray (todo: are false negatives possible?)

_Sidenote: also pondering whether this could potentially be one explaining factor for the open question of why one implementation is 10x faster compared to the other one - or not, this is just a random guess._

---

Note that even the other implementation below it is susceptible:

https://github.com/RayTracing/raytracing.github.io/blob/62ce2e6a1f08b4c103808e513dc1c8fc6a188776/src/common/aabb.h#L78-L88

In this function `invD` will be `NaN` due to the division by zero. The later [multiplications](https://en.cppreference.com/w/c/language/operator_arithmetic#Multiplication) will make `t0` and `t1` set to `NaN`. The `invD < 0` comparison will [always be falsy with a NaN](https://en.cppreference.com/w/c/language/operator_comparison) so the values won't get swapped. The `fmin` and `fmax` function calls will return `other`, the resulting `ray_t.max <= ray_t.min` call will always be falsy so the loop continues.

---

There's also the version that currently exists in the published book:

```cpp
    for (int a = 0; a < 3; a++) {
        auto invD = 1.0f / r.direction()[a];
        auto t0 = (min()[a] - r.origin()[a]) * invD;
        auto t1 = (max()[a] - r.origin()[a]) * invD;
        if (invD < 0.0f)
            std::swap(t0, t1);
        t_min = t0 > t_min ? t0 : t_min;
        t_max = t1 < t_max ? t1 : t_max;
        if (t_max <= t_min)
            return false;
    }
    return true;
```

This should be identical to the one above.

hollasch:
In my environment, this slows down debug builds about 2%, and speeds up
release builds about 9%. More importantly, I believe the code is easier
to understand using interval arithmetic.

In addition, I discovered that in many cases using `fmin()` and `fmax()`
are performance bombs. Likely this is because these functions have
special handling for NaNs. Instead, I switched to using ternary
expressions like `a < b ? a : b`, which had a large performance impact.
For the `aabb::hit()` function, this improved performance of the new
interval code from 100% worse to about 10% better.

Added a new `span()` function that returns an interval of two doubles,
regardless of their order.

Added new `interval::is_empty()` function that also returns true when
either of the bounds is a NaN.

Added new `interval::intersect()` function.

This should resolve #927

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Change Log -- Ray Tracing in One Weekend
 
 # v4.0.0-alpha.2 (In Progress)
 
-This will likely change to the official v4.0.0 release, but we _might_ put out another alpha before
+This will likely change to the official v4.0.0 release, but we might put out another alpha before
 then.
 
 ### Common
@@ -29,8 +29,11 @@ then.
   - Change - Reworked the AABB chapter (#1236)
   - New - add section on alternative 2D primitives such as triangle, ellipse and annulus (#1204,
           #1205)
-  - Change - changed bvh construction (removed const qualifer for objects vector) so sorting is done 
+  - Change - changed bvh construction (removed const qualifer for objects vector) so sorting is done
              in place and copying of vector is avoided, better bvh build performance (#1388, #1391)
+  - Change - AABB code now uses interval arithmetic for the `hit()` function. In my testing, this
+             yields about a 2% slowdown in debug builds, and a 9% speedup in release builds.
+             (#927, #1270)
 
 ### The Rest of Your Life
 
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -615,9 +615,9 @@ Now we have everything we need to implement the new AABB class.
             // Treat the two points a and b as extrema for the bounding box, so we don't require a
             // particular minimum/maximum coordinate order.
 
-            x = span(a[0],b[0]);
-            y = span(a[1],b[1]);
-            z = span(a[2],b[2]);
+            x = (a[0] <= b[0]) ? interval(a[0], b[0]) : interval(b[0], a[0]);
+            y = (a[1] <= b[1]) ? interval(a[1], b[1]) : interval(b[1], a[1]);
+            z = (a[2] <= b[2]) ? interval(a[2], b[2]) : interval(b[2], a[2]);
         }
 
         const interval& axis_interval(int n) const {
@@ -634,12 +634,18 @@ Now we have everything we need to implement the new AABB class.
                 const interval& ax = axis_interval(axis);
                 const double adinv = 1.0 / ray_dir[axis];
 
-                auto s = (ax.min - ray_orig[axis]) * adinv;
-                auto t = (ax.max - ray_orig[axis]) * adinv;
+                auto t0 = (ax.min - ray_orig[axis]) * adinv;
+                auto t1 = (ax.max - ray_orig[axis]) * adinv;
 
-                ray_t = ray_t.intersect(span(s,t));
+                if (t0 < t1) {
+                    if (t0 > ray_t.min) ray_t.min = t0;
+                    if (t1 < ray_t.max) ray_t.max = t1;
+                } else {
+                    if (t1 > ray_t.min) ray_t.min = t1;
+                    if (t0 < ray_t.max) ray_t.max = t0;
+                }
 
-                if (ray_t.is_empty())
+                if (ray_t.max <= ray_t.min)
                     return false;
             }
             return true;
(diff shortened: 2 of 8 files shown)


hollasch:
@armansito — for a sanity check, I'd appreciate a perf comparison for both debug and release builds. Perf measurements from anyone else welcomed as well. I'd suggest running the default scene for TheNextWeek.

hollasch:
Ref: #1270

rupsis:
Figured I'd give a stab at a performance comparison. 
Method: wrapping default final scene in a simple [chrono](https://stackoverflow.com/a/27739925) start/stop method. 

**Note:** I'm using Clang on OSX, using [Clangs optimization flags](https://clang.llvm.org/docs/CommandGuide/clang.html#code-generation-options)

| Method  | Branch | Results |
| ------------- | ------------- | ------------- |
| cmake ` —debug` flag (no other flags)  | Dev latest  |  Render Duration = **378s** (378798ms)|
| cmake ` —release` flag (no other flags)  | Dev latest  |  Render Duration = **379s** (379069ms)|
| cmake ` —debug` flag (no other flags)  | interval-aabb-2  |  Render Duration = **426s** (426921ms) |
| cmake ` —release` flag (no other flags)  | interval-aabb-2   | Render Duration = **424s** (424911ms) |
| unoptimized `-O0` flag  | Dev latest |  Render Duration = **359s** (359946ms)|
| optimized `-O3` flag  | Dev latest  |  Render Duration = **77s** (77526ms)|
| unoptimized `-O0` flag  | interval-aabb-2   |  Render Duration = **500s** (500757ms)|
| optimized `-O3` flag  | interval-aabb-2  |  Render Duration = **66s** (66970ms)|

 
For the Cmake `--debug` vs `--release` flags, I'm seeing no difference, and a hit in performance (which is inline with what @armansito mentioned). For the Clang flags, I'm seeing about a 33% slow down on unoptimized code, and about a 15.4% lift in optimized code. 

At the _very_ least, I'm going to always add `add_compile_options(-O3)`, a 4x performance boost for free!

armansito:
I ran a comparison on my 2021 (ARM) M1 Max MacBook Pro. In an optimized build I see a minor slow down (by about a second which could be statistical noise but seems to be reproducible) and a larger slow down on a debug build.

I produced the release build with `cmake -B build/release -DCMAKE_BUILD_TYPE=Release`, which inserts `-O3` when compiling with Clang. I produced the debug build with `cmake -B build/debug -DCMAKE_BUILD_TYPE=Debug` which leaves the code unoptimized. I measured the wall time by running `time ./build/release/theNextWeek > release.ppm`, etc: 

| Method  | Branch | Results |
| ------------- | ------------- | ------------- |
| CMAKE_BUILD_TYPE=Release (-O3) | Dev latest  |  39.86 |
| CMAKE_BUILD_TYPE=Release (-O3) | interval-aabb-2  |  40.87 |
| CMAKE_BUILD_TYPE=Debug | Dev latest  |  318.7 |
| CMAKE_BUILD_TYPE=Debug | interval-aabb-2  |  353.46 |

I'll try to replicate this on Windows and Ubuntu with different compilers and see what I get.

I think to truly understand the performance impact it might make sense to both a) Capture a profile and see where the cycles get spent; and b) Compare the assembly/IR produced by different compilers using a tool like godbolt. I'm speculating here but some of what you said around fmin/fmax vs ternary operators may vary depending on the instructions those expressions compile down to.

armansito:
Results on Windows, comparing compiling with `--config Release` and `--config Debug`. (The machine is a AMD Ryzen Threadripper 3970X from 2019):

| Method  | Branch | Results |
| ------------- | ------------- | ------------- |
| Release (/O2 /Ob2 /DNDEBUG) | Dev latest  |  64.52 |
| Release (/O2 /Ob2 /DNDEBUG) | interval-aabb-2  | 64.83 |
| Debug (/Ob0 /Od /RTC1) | Dev latest  | 519.54 |
| Debug (/Ob0 /Od /RTC1) | interval-aabb-2  | 524 |

On both computers the release build times are very close to each other and within a 1-2 second variance I get across runs. Debug gets slower as before. I accidentally built the `release` branch at first (instead of `dev`) and that took 941.11 seconds with a Debug build, so fwiw the dev branch is still overall faster than the older version of the code.

I'll try Linux on the same machine next.

armansito:
Same AMD machine on Ubuntu using gcc:

| Method  | Branch | Results |
| ------------- | ------------- | ------------- |
| Release (-O3 -DNDEBUG) | Dev latest  |  47.17 |
| Release (-O3 -DNDEBUG) | interval-aabb-2  | 46 |
| Debug (-g) | Dev latest  | 319.89 |
| Debug (-g) | interval-aabb-2  | 347.19 |

hollasch:
> I accidentally built the release branch at first (instead of dev) and that took 941.11 seconds with a Debug build, so fwiw the dev branch is still overall faster than the older version of the code.

That's good to know — I hadn't considered that. Still I feel like I've held onto this much longer than I should have.

hollasch:
I believe we should just wrap this up and go with this:

```
    bool hit(const ray& r, interval ray_t) const {
        const point3& ray_orig = r.origin();
        const vec3&   ray_dir  = r.direction();

        for (int axis = 0; axis < 3; axis++) {
            const interval& ax = axis_interval(axis);
            const double adinv = 1.0 / ray_dir[axis];

            auto s = (ax.min - ray_orig[axis]) * adinv;
            auto t = (ax.max - ray_orig[axis]) * adinv;

            #if 0
                ray_t = ray_t.intersect(span(s,t));
            #elif 0
                auto si = (s < t) ? interval(s,t) : interval(t,s);
                ray_t = ray_t.intersect(si);
            #else
                if (s < t) {
                    if (s > ray_t.min) ray_t.min = s;
                    if (t < ray_t.max) ray_t.max = t;
                } else {
                    if (t > ray_t.min) ray_t.min = t;
                    if (s < ray_t.max) ray_t.max = s;
                }
            #endif

            #if 1
                if (ray_t.is_empty())
                    return false;
            #else
                if (ray_t.max < ray_t.min)
                    return false;
            #endif
        }
        return true;
    }
```

This appears to be a good speedup across debug & release on Windows, and likely on Linux and Mac as well. I think it's this line:

```
ray_t = ray_t.intersect(span(s,t));
```

that's the killer.

armansito:
I ran the measurements on my M1 Max MBP, replacing `ray_t = ray_t.intersect(span(s,t));` with the if statements. In my case this slowed down the optimized build by 3 seconds on average (I re-compiled ran the different versions multiple times to make sure I wasn't dreaming). That said, I don't think we should dwell on this:  we're well into micro-optimization territory and there's clearly enough variance between the compilers and CPUs that I think this is an exercise better left to the reader.

TBH I prefer the explicit inlined logic (such as the if-statement version) to the `ray_t = ray_t.intersect(span(s,t));` mainly because I find extra abstractions not that pedagogically useful. The extra indirection means that the reader has to jump around files to understand the logic. I think it's generally easier to learn an algorithm (like the AABB intersection logic) if it's presented as one self-contained snippet.

Anyhow, let me know what you think. I defer to you on the best way to proceed. I'll follow up with a review of the PR content and do not want to block this on further performance discussions.

hollasch:
@armansito — Seems like we're in agreement. I'm reworking the PR for (hopefully final) review.

hollasch:
Final reversions done. I will rebase, clean up, and create a final PR for this.

hollasch:
→ Superseded by #1440.

Issue #1358, opened 2024-01-16
[Book 2] List 39 introduces u,v,w, then re-introduces them again in Listing 44


rupsis:
Listing 39 (in Dev)  introduces 3 variables (u,v,w)

![image](https://github.com/RayTracing/raytracing.github.io/assets/9646385/5fdca654-4590-4fa2-8d71-2f2de5bb9a99)

(In Listing 40 it's improved with Hermitian Smoothing)

in Listing 44, we swap over to use Random Vectors, and the highlighting suggest we add u,v,w back in again:
![image](https://github.com/RayTracing/raytracing.github.io/assets/9646385/163553a2-e088-4f3a-b5cc-681a0a616f16)

Showing the removal of the Hermitian Smoothing _might_ make more visual sense:
![image](https://github.com/RayTracing/raytracing.github.io/assets/9646385/2a841d04-077f-46f2-8817-1e9a963acc2c)

Anyway, not really an issue. It tripped me up for a second, and I figured it was worth mentioning. Happy to make this change :)

hollasch:
I like the removal idea — go for it.

rupsis:
Fixed in https://github.com/RayTracing/raytracing.github.io/commit/8646911a5b7a7562e113002fc7b3222934396798

PR #1425, opened 2024-03-10
Fix section describing total internal reflection

### LINKED ISSUE #900: Book 1 Section 10.3 Dielectric Sphere In Images Appears Identical Despite Total Internal Reflection Changes
An image of 2 glass spheres is presented immediately prior to Section 10.3, and then in Section 10.3, Total Internal Reflection, it starts off by saying:

That doesn't look right. <proceeds to describe total internal reflection>

and then a new image is presented, which (randomly) changes the center sphere to blue and removes the fuzz on the right sphere.

The presentation flows like, "look at this image, ok it doesn't look right, then ok here's a new thing to consider (total internal reflection), and now look at the new image."

But that doesn't appear to be what was intended. Instead, we're introduced to total internal reflection, which doesn't even seem to appear in either image (as the left sphere appears the same in both images). The obvious thing to look for in the second image is some kind of improvement in the left sphere, after the discussion of TIR, but instead there's just those random 2 unrelated changes mentioned above.

The first image is captioned with "Glass sphere that always refracts", while the second says "Glass sphere that sometimes refracts", and yet the left sphere in both images appears identical. I don't understand what the author intended to convey here.

hollasch:
With a sphere of index of refraction greater than the surrounding medium, there is no incoming ray that can be chosen to show total internal reflection. An incoming ray will be bent to a smaller angle relative to the surface normal, and then bent back to the original angle on exit.

To illustrate total reflection, this change renders a bubble of air inside water. Glancing rays are then reflected off the surface instead of refracted.

In the service of this change, I've also moved the center sphere back a bit from the left and right spheres so you can see the affected side more clearly. This will invalidate all of the other three-sphere images, which will be fixed up in #1358 (Alpha.2 milestone verify Book 1 progression).

Resolves #900


books/RayTracingInOneWeekend.html
```diff
@@ -3302,9 +3303,11 @@
 
 Total Internal Reflection
 --------------------------
-That definitely doesn't look right. One troublesome practical issue is that when the ray is in the
-material with the higher refractive index, there is no real solution to Snell’s law, and thus there
-is no refraction possible. If we refer back to Snell's law and the derivation of $\sin\theta'$:
+One troublesome practical issue with refraction is that there are ray angles for which no solution
+is possible using Snell's law. This happens when the ray goes into a medium of lower index of
```
armansito: Slight rephrasing of the second sentence:

_When the ray enters a medium of lower index of refraction at a sufficiently glancing angle, it can refract with an angle greater than 90&deg;._

books/RayTracingInOneWeekend.html
```diff
@@ -3400,25 +3405,42 @@
...
-parameters:
+Attenuation is always 1 -- the glass surface absorbs nothing.
 
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ Highlight
+If we render the prior scene with the new `dielectric::scatter()` function, we see … no change. Huh?
+
+Well, it turns out that given a sphere of material with an index of refraction greater than air,
+there's no incident angle that will yield total internal reflection -- neither at the ray-sphere
+entrance point nor at the ray exit. This is due to the geometry of spheres, as a grazing incoming
+ray will always be bent to a smaller angle, and then bent back to the original angle on exit.
```
armansito, line 3414: I feel that a figure demonstrating the path of a ray and the angles of reflection/refraction (such as the one in #900) would go a long way here.

armansito (approved):
Approved with some minor suggestions. My suggestion about a figure doesn't need to get resolved in this PR.

books/RayTracingInOneWeekend.html
hollasch: _Sigh_. https://psgraphics.blogspot.com/2020/11/how-to-write-paper-when-you-suck-at.html

I have a huge embarrassing weakness — I am much more comfortable writing lots of text than I am generating illustrations and other graphics. Ironic, but there it is.

I think you're absolutely right. Created a new issue #1444 — feel free to jump on this if you wish.

armansito: Sounds good, I think I can take a crack at it. Thank you for the link to the blog post, it's a great read.

books/RayTracingInOneWeekend.html

Issue #1420, opened 2024-03-06
Implement bubbles with ratios of indices of refraction and eliminate negative radius spheres


hollasch:
Currently, we implement bubbles (or transparent spheres with non-zero thickness) by specifying an inner sphere of negative radius. It's a hack that works, but creates several geometric issues that then need to be special-cased. Here are some issues dealing with this:

- #270
- #532
- #733
- #899
- #977
- #1130
- #1278

However, there's a much cleaner way to address this. Instead of having `dielectric` take an index of refraction for the single construction parameter, it should take the ratio of the inner index of refraction over the outer index of refraction. Since air is considered to have an index of refraction of 1.00, the resulting value is the same most materials, where the index of refraction is greater then 1.00. However, this means that bubbles (say of air inside glass) can be specified with a ratio of 1.00 over the index of refraction of the object material (that is, modeling a bubble of air, instead of a bubble of negative glass).

Tests show that the results are identical. To model a hollow glass sphere of non-zero thickness (implicitly filled with air), current code defines an inside sphere as having a negative radius, with an index of refraction of glass. If we change this to have an inside sphere of positive (but smaller) radius, with an index of refraction of 0.67 (= 1 / IOR_glass), then we get identical results. If we follow this approach, then we can eliminate the need to handle or special-case spheres of negative radii.

Before, with negative-radius bubble, IOR of glass:
![x_720](https://github.com/RayTracing/raytracing.github.io/assets/313871/06683ede-e3a8-44ab-a885-d3e887e37253)

After, with positive-radius bubble, IOR of 1/glass:
![y_720](https://github.com/RayTracing/raytracing.github.io/assets/313871/3edeb5cc-e866-4752-9a14-11444c84e1da)

(Note that Vassillen Chizhov (@vchizhov) has already made this suggestion parenthetically here: https://github.com/RayTracing/raytracing.github.io/issues/270#issuecomment-561524329.)

If this work is done, I suggest we either take the absolute value of sphere radii in the constructor, or throw an exception for negative radii or something. Following that, remove the negative-radius hack from the text.

hollasch:
I should also point out that this change would allow you to specify bubbles of arbitrary material, such as a bubble of air inside a bubble of water inside a bubble of diamond inside a bubble of quartz inside a sphere of glass in the air.

[...]

PR #1426, opened 2024-03-11
Correct call to reflectance()

hollasch:
The function `reflectance()` takes `double ref_idx` (which is `ir` in the outer scope), not `refraction_ratio`.

Confirmed with the formula for Schlick's approximation (https://en.wikipedia.org/wiki/Schlick%27s_approximation) which uses two indexes of refraction, of which one is hard coded to `1` in the function here.

A future todo would be to keep track of the "ir of the current medium" in the ray and use that instead of a hard coded `1` in the reflectance formula.

Originally due to https://github.com/TheThief.


hollasch:
Original PR from @TheThief at #1418.

hollasch:
After review, it turns out that this fix isn't needed. For details, see #1420.

PR #1427, opened 2024-03-11
Normalize constructor parameter names

hollasch:
In order to avoid confusion between constructor parameter names and member variables, we either used alternate names for the parameters or prefixed the parameter names with an underscore. For example:

    class foo {
      public:
        foo(int _i, bool semaphore, short bar) : i(_i), flag(semaphore)
        {
            baz = bar;
        }
       private:
         int i;
         bool flag;
         short baz;
    };

However, this is unnecessary. Constructor initializer lists are unambiguous, as the assigned item always refers to the class member, and the assigned value is always from the parameter (unless prefixed with "this->").

Inside the constructor body, parameter names override member variables, and "this->" can be used to indicate the class member in the case of name collision.

Thus, the above code can be rewritten as this:

    class foo {
      public:
        foo(int i, bool flag, short baz) : i(i), flag(flag)
        {
            this->baz = baz;
        }
       private:
         int i;
         bool flag;
         short baz;
    };

I've taken advantage of this to clarify names used in constructors throughout the codebase. Mostly for constructor parameter names, sometimes for member variable names.

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -2844,18 +2844,18 @@ class:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ Highlight
     class lambertian : public material {
       public:
-        lambertian(const color& c) : c(c) {}
+        lambertian(const color& albedo) : albedo(albedo) {}
 
         bool scatter(const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered)
         const override {
             auto scatter_direction = rec.normal + random_unit_vector();
             scattered = ray(rec.p, scatter_direction);
-            attenuation = c;
+            attenuation = albedo;
             return true;
         }
 
       private:
-        color c;
+        color albedo;
     };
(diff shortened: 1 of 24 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -2844,18 +2844,18 @@
...
             auto scatter_direction = rec.normal + random_unit_vector();
             scattered = ray(rec.p, scatter_direction);
-            attenuation = albedo;
+            attenuation = c;
             return true;
         }
 
       private:
-        color albedo;
+        color c;
```
rupsis: My preference would be `col` (or `colour` if we want to make people mad). 
One less single character variable to remember.

x, y, z, Q, u, v, etc are all used in formulas.

rupsis (approved):
1 Subjective opinion. Otherwise the changes make sense, and go a long way to unify the code.

hollasch:
The problem is that `col` more typically refers to a column. And yes, `c` often refers to a char. `color` is out because it's a type name. (Personally, I prefer leading capitals on my type names, but that's not the convention here.

"color" actually isn't a great name, because it makes the common mistake of naming something after its kind/type, instead of what it is specifically. I'll revisit.

books/RayTracingInOneWeekend.html
hollasch: After discussion in Slack, reverted back to "albedo". See also #1430.

PR #1428, opened 2024-03-11
No negative radius spheres

### LINKED ISSUE #1420: Implement bubbles with ratios of indices of refraction and eliminate negative radius spheres

hollasch:
**[This change is stacked on top of the outstanding PR to normalize constructor parameter names. Ignore that commit, as I'll rebase this onto dev when that PR is merged.]**

Implement bubbles with IOR, not negative radii

We were using negative radii on spheres to invert the sense of front vs
back faces of spheres, and using this to get an inverted index of
refraction (IOR) for bubbles. This method was used to model a hollow
glass sphere.

Over the past couple of years we've encountered a number of bugs induced
by negative radii spheres (see the list in the description for issue
1420).

Instead, this change interprets the dielectric parameter not as an IOR
relative to surrounding air, but instead as the ratio of the IOR of the
object material over the IOR of the surrounding medium. Interpreting the
parameter this way lets you model a sphere of air enclosed in glass --
just what we need to model a hollow glass sphere purely with the proper
IORs, and without resorting to inverted geometry with negative radii
spheres.

We also end up using this method to properly demonstrate total
internal/external reflection in another section.

Resolves #1420

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -3503,14 +3503,16 @@ through the air inside. Then it will continue on, hit the inside surface of the 
 refract back, then hit the inside surface of the outer sphere, and finally refract and exit back
 into the scene atmosphere.
 
-The outer sphere is just modeled with a standard glass sphere, with an index of refraction of around
-1.50. The inner sphere is a bit different, because _its_ index of refraction should be relative to
-the "glass atmosphere" of the surrounding outer sphere. This is actually simple to specify, as our
-index-of-refraction parameter to the dielectric material can be interpreted as the ratio of the
-index of refraction of the object over the index of refraction of the enclosing medium.
+The outer sphere is just modeled with a standard glass sphere, with a refractive index of around
+1.50 (modeling a refraction from the outside air into glass). The inner sphere is a bit different
+because _its_ refractive index should be relative to the "glass atmosphere" of the surrounding outer
+sphere, thus modeling a transition from glass into the inner air.
 
-In this case, the inner sphere would have an index of refraction of air (the inner sphere material)
-over the index of refraction of glass (the enclosing medium), or $1.00/1.50 = 0.67$.
+This is actually simple to specify, as the `ref_index` parameter to the dielectric material can be
+interpreted as the _ratio_ of the refractive index of the object divided by the refractive index of
+the enclosing medium. In this case, the inner sphere would have an refractive index of air (the
+inner sphere material) over the index of refraction of glass (the enclosing medium), or $1.00/1.50 =
+0.67$.
 
 Here's the code:
 
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -1114,7 +1114,7 @@ And here’s the sphere:
 
     class sphere : public hittable {
       public:
-        sphere(const point3& center, double radius) : center(center), radius(radius) {}
+        sphere(const point3& center, double radius) : center(center), radius(fmax(0,radius)) {}
 
         bool hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec) const override {
             vec3 oc = center - r.origin();
@@ -2795,7 +2795,7 @@ To achieve this, `hit_record` needs to be told the material that is assigned to 
       public:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
         sphere(const point3& center, double radius, shared_ptr<material> mat)
-          : center(center), radius(radius), mat(mat) {}
+          : center(center), radius(fmax(0,radius)), mat(mat) {}
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
 
         bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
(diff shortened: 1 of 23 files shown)


books/RayTracingInOneWeekend.html
```diff
@@ -3477,50 +3477,47 @@
...
+inside it. If you think about the path of a ray going through such an object, it will hit the outer
+sphere, refract, hit the inner sphere (assuming we do hit it), refract a second time, and travel
+through the air inside. Then it will continue on, hit the inside surface of the inner sphere,
+refract back, then hit the inside surface of the outer sphere, and finally refract and exit back
+into the scene atmosphere.
 
-However, properly handling negative radii can be tricky. Recall the line from `sphere::hit()` in
-listing [sphere-material] that calculates the outward normal:
+The outer sphere is just modeled with a standard glass sphere, with an index of refraction of around
+1.50. The inner sphere is a bit different, because _its_ index of refraction will now be relative to
```
armansito: Some optional small suggestions for this paragraph in **_bold_**:

The outer sphere is just modeled with a standard glass sphere, with an index of refraction of around 1.50 **_relative to the surrounding air_**. The inner sphere is a bit different because _its_ index of refraction **_should be_** relative to the **_surrounding glass outer sphere and model a transition back to air_**. This is actually simple ...

armansito (approved):
I really like this change. Approved with some optional suggestions.

books/RayTracingInOneWeekend.html
hollasch: I'll tweak the wording.

hollasch: Hmmm, I barely tweaked it:
> The outer sphere is just modeled with a standard glass sphere, with an index of refraction of around
1.50. The inner sphere is a bit different, because _its_ index of refraction should be relative to
the "glass atmosphere" of the **surrounding** outer sphere. This is actually simple to specify, as our

I don't want to add "relative to the surrounding air" since it's really a surrounding vacuum, (air = 1.000293), and most refractive indices are given with an implicit understanding that they're already relative to vacuum. Also, for the second change it's actually transitioning back into glass.

Either way, I'm going to add a bit more explanation earlier where we introduce refraction.

armansito: I agree, vacuum is more correct. I only suggested air because it was used elsewhere in the book. My main point was to highlight that the refractive index is always relative to the surrounding medium (while a reader may assume its _absolute_ meaning, i.e. relative to vacuum). I think mentioning "relative to the surrounding air/vacuum" in the first sentence sets things up for the second sentence but please use whichever phrasing you prefer.

> ... Also, for the second change it's actually transitioning back into glass.

Doesn't a transition from the inner sphere's exterior (as defined by the outward facing normals) to its interior represent a transition from glass to air/vacuum (or whatever interior/exterior medium pair that happens to have the same refraction ratio)? That's really what I meant but I may have misunderstood something.

> Either way, I'm going to add a bit more explanation earlier where we introduce refraction.

Sounds good. As I said I think the main concern was to highlight that the refractive index is always relative. Even the absolute one is relative but readers may make assumptions about the definition of IOR and get confused about the meaning of `dialectric::ref_index`.

hollasch: See the next commit and let me know what you think.

> Doesn't a transition from the inner sphere's exterior (as defined by the outward facing normals) to its interior represent a transition from glass to air/vacuum (or whatever interior/exterior medium pair that happens to have the same refraction ratio)? That's really what I meant but I may have misunderstood something.

Ah, yes, I misunderstood your suggested reading. I think it was the "back to air" that confused me. I'd rather explain this as a transition into a third medium that just happens in this example to be the same as the first medium. Anyway, I think the paragraph immediately preceding this explains this all in detail, and we don't need to repeat it in this paragraph.

hollasch: This?

> The outer sphere is just modeled with a standard glass sphere, with an index of refraction of around 1.50 relative to the surrounding air, **modeling a refraction from air into glass**. The inner sphere is a bit different because its index of refraction should be relative to the surrounding glass outer sphere and model a transition **from glass into air**. This is actually simple ...

hollasch: Kind of wish we had three different materials to keep things straight, but the model itself would then be rather needlessly complicated. Perhaps use the terms "outside air" and "inside air"? `¯\_(ツ)_/¯`

hollasch:
Ugh. Too many PRs piled up — now this one has a bunch of conflicts. I'll have to create a second one for this, rebased.

hollasch:
Superseded by #1435.

Issue #1432, opened 2024-03-12
Clang --debug vs --release flags have no performance benefit 


rupsis:
While testing out #1422, I noticed we get a huge performance boost on OSX/Clang while using the `-O3` flag. 
However, the Cmake `--debug` vs `--release` flags seem to have no impact on performance. 

We should add a conditional check against Clang in the `CMakeList.txt` to set the `-O3` [optimization flag](https://clang.llvm.org/docs/CommandGuide/clang.html#code-generation-options).

```
if (CMAKE_BUILD_TYPE STREQUAL "Release")
    add_compile_options(-O3)
endif()
```

Happy to make these changes :D 

_Originally posted by @rupsis in https://github.com/RayTracing/raytracing.github.io/issues/1422#issuecomment-1989587869_

armansito:
Yes, I think we should add this for Clang. If you have the time for it I would certainly appreciate it :)

PR #1434, opened 2024-03-12
Update README with --release / --debug examples

### LINKED ISSUE #1433: Cmake flags in README Building and Running is a bit vague
In the readme, under "Building and Running" the section:
```
On Windows, you can build either debug (the default) or release (the optimized version). 
To specify this, use the --config <debug|release> option.
``` 

suggests (to me) that these are _Windows specific_ flags, rather than general Cmake options. 

I also am a fan of examples, so adding in one for the `--target`, and the `--config` would be helpful. But I understand the desire to not clutter the README. 

happy to make these changes.

rupsis:
Resolves #1433

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,14 @@ You can specify the target with the `--target <program>` option, where the progr
 `inOneWeekend`, `theNextWeek`, `theRestOfYourLife`, or any of the demonstration programs. By default
 (with no `--target` option), CMake will build all targets.
 
-On Windows, you can build either `debug` (the default) or `release` (the optimized version). To
+    $ cmake --build build --target inOneWeekend
+
+You can build either `debug` (the default) or `release` (the optimized version). To
 specify this, use the `--config <debug|release>` option.
+
+    $ cmake --build build --config release
+
+We recommend building the `release` version before the final render for fastest results.
 
 ### CMake GUI on Windows
 You may choose to use the CMake GUI when building on windows.
--- a/README.md
+++ b/README.md
@@ -129,12 +129,15 @@ GitHub pages serves up https://raytracing.github.io/.
 
 Building and Running
 ---------------------
-Copies of the source are provided for you to check your work and compare against. If you wish to build
-the provided source, this project uses CMake. To build, go to the root of the project directory and
-run the following commands to create the debug version of every executable:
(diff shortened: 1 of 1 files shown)


armansito (approved):
I verified that these flags work on macOS. Do you think it makes sense to annotate `cmake -B build` as the first thing the reader must run to create the directory? Something like:

```shell
$ cmake -B build        # configure the project and create the 'build' directory
$ cmake --build build   # compile  
```

Mentioning this since the nuance wasn't obvious to me as someone who doesn't use cmake frequently (and I didn't know I could rerun the first line to fix my build after my configuration got messed up for some reason).

armansito:
> I verified that these flags work on macOS. 

To clarify, I meant that cmake accepts these config options on macOS without complaining. However, they don't have any effect as described in #1432.

rupsis:
> Do you think it makes sense to annotate cmake -B build as the first thing the reader must run to create the directory?

Yes. In my opinion, the more Cmake help the better. My brain has this uncanny ability to forget anything I've learned about Cmake (regex too for that matter...).

armansito:
OK, I think the original directions were in fact specific to Windows. I think we should keep those as such and for Unix/macOS we suggest the config-time `CMAKE_BUILD_TYPE` setting (e.g. `cmake -B build/Release -DCMAKE_BUILD_TYPE=Release`). The end result is the same with both approaches (i.e. you end up with a `build/Release` directory containing the artifacts), with the only difference that the Release/Debug config is set at build time on Windows and at config time on macOS and Linux.

If there are no objections to that, I'll go ahead and submit a change to the README.

**NOTE**: On Windows, the default `cmake -B build` invocation sets `CMAKE_CXX_FLAGS_RELEASE` and `CMAKE_CXX_FLAGS_DEBUG`. The former (Release) is set to `/O2 /Ob2 /DNDEBUG` and the latter (Debug) is set to `/Ob0 /Od /RTC1`.

hollasch:
> If there are no objections to that, I'll go ahead and submit a change to the README.

Yeah, sounds great.

I think that the real answer is that these differences are due to the discovered C++ compiler. What we're calling "Windows" here is probably really "Visual Studio". But we're getting more and more into the weeds here, and I think calling it Windows is fine.

PR #1435, opened 2024-03-12
Implement bubbles with IOR, not negative radii

### LINKED ISSUE #1420: Implement bubbles with ratios of indices of refraction and eliminate negative radius spheres

hollasch:
We were using negative radii on spheres to invert the sense of front vs back faces of spheres, and using this to get an inverted index of refraction (IOR) for bubbles. This method was used to model a hollow glass sphere.

Over the past couple of years we've encountered a number of bugs induced by negative radii spheres (see the list in the description for issue 1420).

Instead, this change interprets the dielectric parameter not as an IOR relative to surrounding air, but instead as the ratio of the IOR of the object material over the IOR of the surrounding medium. Interpreting the parameter this way lets you model a sphere of air enclosed in glass -- just what we need to model a hollow glass sphere purely with the proper IORs, and without resorting to inverted geometry with negative radii spheres.

We also end up using this method to properly demonstrate total internal/external reflection in another section.

Resolves #1420

### DIFF
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -3516,8 +3516,8 @@ into the scene atmosphere.
 
 The outer sphere is just modeled with a standard glass sphere, with a refractive index of around
 1.50 (modeling a refraction from the outside air into glass). The inner sphere is a bit different
-because _its_ refractive index should be relative to the "glass atmosphere" of the surrounding outer
-sphere, thus modeling a transition from glass into the inner air.
+because _its_ refractive index should be relative to the material of the surrounding outer sphere,
+thus modeling a transition from glass into the inner air.
 
 This is actually simple to specify, as the `refraction_index` parameter to the dielectric material
 can be interpreted as the _ratio_ of the refractive index of the object divided by the refractive
--- a/books/RayTracingInOneWeekend.html
+++ b/books/RayTracingInOneWeekend.html
@@ -1114,7 +1114,7 @@ And here’s the sphere:
 
     class sphere : public hittable {
       public:
-        sphere(const point3& center, double radius) : center(center), radius(radius) {}
+        sphere(const point3& center, double radius) : center(center), radius(fmax(0,radius)) {}
 
         bool hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec) const override {
             vec3 oc = center - r.origin();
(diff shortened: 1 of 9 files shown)


hollasch:
Supersedes #1428, rebased on top of latest dev, incorporating suggestions from @armansito.

books/RayTracingInOneWeekend.html
```diff
@@ -3485,50 +3507,49 @@
...
 
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
-    vec3 outward_normal = (rec.p - center).unit_vector();
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [improper-invert-sphere-normal]:
-        Problematic normal calculation for spheres with negative radii
-    ]
+The outer sphere is just modeled with a standard glass sphere, with a refractive index of around
+1.50 (modeling a refraction from the outside air into glass). The inner sphere is a bit different
+because _its_ refractive index should be relative to the "glass atmosphere" of the surrounding outer
```
armansito: I think I prefer "glass interior" to "atmosphere". While I appreciate the analogy I think it's potentially confusing, given the mixture of terms around "air", "vacuum", "enclosing medium" and the gaseous implications of the term atmosphere

armansito (approved):
Looks great. Just one nitpick.

books/RayTracingInOneWeekend.html
hollasch: That's fair. I've changed it to

> The inner sphere is a bit different because _its_ refractive index should be relative to the **material** of the surrounding outer sphere, ...

PR #1440, opened 2024-03-19
Refactor AABB class

### LINKED ISSUE #927: AABB hit and NaNs & Infinities

hollasch:
Discovered that sometimes `std::fmin()` and `std::fmax()` can be performance bombs. Likely this is because these functions have special handling for NaNs. Instead, I switched to using ternary expressions like `a < b ? a : b`, which had a large performance impact. This particularly impacts the `aabb::hit()` function.

Update book for latest AABB changes:

- Add clarifying comment for corner cases of ray-aabb intersection.
- Fix some listings.
- Update code listings.
- Deprecate section "An Optimized AABB Hit Method"

Resolves #927

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -574,8 +574,10 @@ given amount:
     class interval {
       public:
         ...
-        double size() const {
-            return max - min;
+        double clamp(double x) const {
+            if (x < min) return min;
+            if (x > max) return max;
+            return x;
         }
 
 
@@ -586,7 +588,7 @@ given amount:
         }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
 
-        ...
+        static const interval empty, universe;
     };
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     [Listing [interval-expand]: <kbd>[interval.h]</kbd> interval::expand() method]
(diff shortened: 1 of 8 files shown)


hollasch:
_phew_

hollasch:
Supersedes #1422

books/RayTracingTheNextWeek.html
```diff
@@ -614,25 +616,37 @@
         aabb(const point3& a, const point3& b) {
             // Treat the two points a and b as extrema for the bounding box, so we don't require a
             // particular minimum/maximum coordinate order.
-            x = interval(fmin(a[0],b[0]), fmax(a[0],b[0]));
-            y = interval(fmin(a[1],b[1]), fmax(a[1],b[1]));
-            z = interval(fmin(a[2],b[2]), fmax(a[2],b[2]));
+
+            x = (a[0] <= b[0]) ? interval(a[0], b[0]) : interval(b[0], a[0]);
```
armansito, line 620: Is a negative interval ever valid? If not, would it make sense to move this logic to the `interval(double, double)` constructor and sort the arguments there, the same way `interval(const interval&, const interval&)` handles this for the interval-from-intervals case?

armansito (approved):
Looks good. Just one question regarding one of the interval constructors.

books/RayTracingTheNextWeek.html
hollasch: If you mean an interval where $min > max$, then yes, this is how empty intervals are represented. They naturally fall out of and interoperate with intersection and union results using purely arithmetic operations.

Using the representation of $null = [+∞,-∞]$ and $universe = [-∞,+∞]$, then intersection is simply max of the mins and the min of the maxes, while union is the min of the mins and the max of the maxes. The special $null$ and $universe$ values also flow through these arithmetic operations without requiring examination and multiple special-case branches.

The case where you _don't_ want an inverted interval is what the new `span()` function was for, reordering the values as needed to get a positive interval. But with the interval AABB logic falling out, we're back to not needing such a `span()` function.

armansito: I should have been a bit more clear: I understand that the special case of $null = [+∞,-∞]$ is constructed this way, using the default `interval` constructor. I was mainly wondering if it's expected to construct such an empty interval using the `interval(double, double)` constructor as well, and whether that should sort its input.

Either way I'm fine with the changes as they are.

hollasch: I see. Yes, instead of

```
const interval interval::empty = interval(+infinity, -infinity);
```

we could have instead used

```
const interval interval::empty    = interval();
```

But imagine the following two implementations of `intersect()` and `union()`:

```
interval intersect(interval a, interval b) {
    return interval(fmax(a.min, b.min), fmin(a.max, b.max));
}

interval union(interval a, interval b) {
    return interval(fmin(a.min, b.min), fmax(a.max, b.max));
}
```

These two functions work with null, universe and half-open intervals in either direction from any point. These would require a flexible `(double, double)` constructor that does not reinterpret its arguments.

You could certainly argue that for our current uses this is a theoretical objection, but I didn't want to hamper the class from future operations in case they may need in the future.

(BTW, `union` may not be the best name for the second function, but it definitely is an operation that's often needed, which bounds two input intervals in the same way that an interval bounds to input scalars.)

armansito: OK, I see the utility of having a plain assignment constructor.

hollasch: It just occurred to me that since we have the simplistic style of public data members, your question makes more sense. The "front door" via constructor could be restrictive, but since we have direct access to the min/max values, we could be free of up-front tidying of values, and instead create & mutate values to get inverted intervals. So not a crazy suggestion.

(I'm soooo glad I'm not on something like the C++ committee, where I could rat-hole on such esoteric considerations for decades.)

PR #1459, opened 2024-03-24
Fix undefined lights parameter for cam.render()

### LINKED ISSUE #1389: Book 3 Listing 5 ("Stratifying the samples inside pixels") passes undefined `lights` to `cam.render()`
The render call should just be

    cam.render(world);

hollasch:
Resolves #1389

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -298,7 +298,7 @@ location.
 
         cam.defocus_angle = 0;
 
-        cam.render(world, lights);
+        cam.render(world);
     }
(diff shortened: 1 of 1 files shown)


books/RayTracingTheRestOfYourLife.html
```diff
@@ -298,7 +298,7 @@
 
         cam.defocus_angle = 0;
```
rupsis, line 299: Not sure if this _really_ matters, but in the source code `cam.defocus_angle` is `0.0` (rather than just `0`)

same with `vfov` (40.0 instead of 40). 

I can make a separate issue, or you an quickly address them here.

rupsis (approved):

books/RayTracingTheRestOfYourLife.html
hollasch: I'll just do a quick fix after this.

Issue #1321, opened 2023-10-28
Book 3.6.2: Need for clarification about the resulting render when using an imperfect PDF


Shridhar2602:
In section 6.2, "Using a Uniform PDF Instead of a Perfect Match",  there's a need for clarification to avoid confusion.
The section currently states -

> This will still converge on the correct answer, as all we've done is change the PDF, but since the PDF is now less of a perfect match for the real distribution, it will take longer to converge. Which, for the same number of samples means a noisier image:
> ....
>You should get a very similar result to before, only with slightly more noise, it may be hard to see.

In my implementation, I am getting a much brighter image after changing the PDF to uniform, which as per my knowledge is the expected behavior. While the section does mention increased noise, it doesn't adequately address the potential issue of a brighter or dimmer rendered image. This omission can be frustrating for beginners, who would expect to see a slightly noisy image but would instead get a drastically different render after changing the pdf. 

Or am I mistaken and going wrong somewhere?

gau-nernst:
I observed the same problem. From my understanding, the problem is related to #933. Not sure why it was fixed in v3 but v4 has the wrong implementation. Basically, the sampling PDF and scattered ray should be linked. We cannot change the sampling PDF (`pdf` in the code) without changing how the scattered ray is sampled.

dimitry-ishenko:
Related to #1302

Issue #1302, opened 2023-10-23
Book 3.6.2: Missing actual change to the probability function 


TheThief:
It's not very clearly explained IMO, but the pdf value you're multiplying by is the value from the ideal pdf you want to use for the surface, and the one you divide by is the value from the pdf you've _actually used_ to generate the rays (which can e.g. be biased towards lights for importance sampling).

But in 6.2 as it stands at the time of writing (4.0.0-alpha.1 I believe) you're only changing the value you're dividing by to a uniform value, but still _generating_ with the cosine distribution inside the material! So they don't match, and instead of correcting for the distribution being used it's actually altering it to a double cosine reflectance by accident.

The change you make in 6.3 to change to generate rays using a uniform hemisphere should be in 6.2 as well, to get the effect you're claiming of giving the same result but taking longer to converge (aka more noise).

Also, a lot of the code past this point is using the material function that returns a scatter ray in an out parameter but then throws that away and generates another one it actually uses immediately afterwards, which is inefficient

hollasch:
From (now closed #1272):
> In 6.2 Using a Uniform PDF Instead of a Perfect Match, the scatter ray is sampled in Lambert Distribution in rec.mat->scatter(), but the pdf value is set to 1/(2*pi). I think it is against the Monte Carlo Theory. Can It still converge to the correct answer? As the book said  This will still converge on the correct answer. I'm confused about this.

dimitry-ishenko:
FWIW, I got different results between 6.1 and 6.2.

6.1
![img-3 6 1](https://github.com/RayTracing/raytracing.github.io/assets/2858291/e897866e-9bf8-49ed-8a96-0594a5cbd53b)

6.2
![img-3 6 2](https://github.com/RayTracing/raytracing.github.io/assets/2858291/12f4da5e-2614-4f7a-acc0-373e4e9c3040)

This seems to make sense, actually. In 6.1 we multiply and divide by the same value of the PDF, which nullify each other. In 6.2 brighter rays are penalized less, relative to the darker ones since all are divided by the same amount.

I think this sentence in 6.2:

> We will sample using a uniform PDF about the hemisphere, so we'll set the denominator to 1/2𝜋. This will still converge on the correct answer ...

goes against what was said earlier in the book. For example, when talking about directing rays towards the light in 3.3:

[...]

PR #1462, opened 2024-03-27
Progression 3 updates

hollasch:
A slew of changes from the book 3 progression up to listing 28. Listing 28 leaves us in a broken state, covered in issues #1302, #1321.


hollasch:
Here's a reminder when going through these potato chip changes that on GitHub while reviewing a commit, `n`/`p` will advance to the next/previous commit in the PR.

hollasch:
Oh, by the way, the corresponding book 3 progression for this batch is from `progression/3/2024-0325`.

books/RayTracingTheRestOfYourLife.html
```diff
@@ -110,23 +110,24 @@
...
             if (x*x + y*y < 1)
                 inside_circle++;
         }
-        std::cout << std::fixed << std::setprecision(12);
+
         std::cout << "Estimate of Pi = " << (4.0 * inside_circle) / N << '\n';
     }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [estpi-1]: <kbd>[pi.cc]</kbd> Estimating π]
+    [Listing [estpi-1]: <kbd>[pi.cc]</kbd> Estimating π -- version one]
```
rupsis: For consistency of symbols, shouldn't this be `$\pi$` instead of π?

books/RayTracingTheRestOfYourLife.html
```diff
@@ -143,28 +144,35 @@
...
         }
+
+
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ delete
+        std::cout << "Estimate of Pi = " << (4.0 * inside_circle) / N << '\n';
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    [Listing [estpi-2]: <kbd>[pi.cc]</kbd> Estimating π, v2]
+    [Listing [estpi-2]: <kbd>[pi.cc]</kbd> Estimating π -- version two]
```
rupsis: Same question here.

books/RayTracingTheRestOfYourLife.html
```diff
@@ -356,9 +371,10 @@
             center = lookfrom;
             ...
         }
-        ...
-        ray get_ray(int i, int j, int s_i, int s_j) const {
+
+
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
+        ray get_ray(int i, int j, int s_i, int s_j) const {
```
rupsis, line 377: Should this line get highlighted? The function signature doesn't change.

books/RayTracingTheRestOfYourLife.html
```diff
@@ -457,27 +473,19 @@
 
 <div class='together'>
 Our work above is equally valid as a means to solve for $pi$ as it is a means to solve for the area
```
rupsis: Shouldn't this be `$\pi$` instead of `$pi$` ?

rupsis (approved):
A couple questions, but overall changes make sense and look good.

books/RayTracingTheRestOfYourLife.html
hollasch: I limit code styles and mathematical typesetting in the captions. If you think we should reconsider this, validate that Markdeep is cool with this (also use single quotes instead of HTML `<kbd>` tags), and create a new issue for the backlog.

books/RayTracingTheRestOfYourLife.html
hollasch: Yup!

books/RayTracingTheRestOfYourLife.html
hollasch: This is a definite change (the addition of stratified indices). The prior signature (book 2) was:

```
        ray get_ray(int i, int j) const {
```

books/RayTracingTheRestOfYourLife.html
hollasch: Looks like using back quotes gets the background-shaded and italic code style in the caption, which is no good. However, I believe this would be an easy adjustment to our CSS. MathJax looks fine. I'm fine with doing this work.

rupsis: Seems to work well with just 
```
$\pi$
```
(github seems to render valid LaTeX, sorry for weird formatting)

image

hollasch: Heh, we have captions with exponents that get rendered with superscripts, but likely due to Markdeep, and not MathJax. I'll have to revisit all captions (image and listing).

I'm going to let the filename sit for now with `<kbd>` tags, and will (sigh) go ahead and update the captions, since we're using math notation that's not properly rendered with either Markdeep or MathJax.

books/RayTracingTheRestOfYourLife.html
rupsis: Ah, I was just going off of the diff. All good then!

books/RayTracingTheRestOfYourLife.html
hollasch: Ah, we're already using MathJax in captions in book 2...

hollasch: Changes made.

PR #1469, opened 2024-03-28
Use .x(), .y() and .z() instead of axis numbers

dimitry-ishenko:
Improve readability and better match formulas in the book.

### DIFF
--- a/src/TheNextWeek/hittable.h
+++ b/src/TheNextWeek/hittable.h
@@ -91,17 +91,15 @@ class rotate_y : public hittable {
         point3 min( infinity,  infinity,  infinity);
         point3 max(-infinity, -infinity, -infinity);
 
-        for (int i = 0; i < 2; i++) {
-            for (int j = 0; j < 2; j++) {
-                for (int k = 0; k < 2; k++) {
-                    auto x = i*bbox.x.max + (1-i)*bbox.x.min;
-                    auto y = j*bbox.y.max + (1-j)*bbox.y.min;
-                    auto z = k*bbox.z.max + (1-k)*bbox.z.min;
+        for (auto x : { bbox.x.min, bbox.x.max }) {
+            for (auto y : { bbox.y.min, bbox.y.max }) {
+                for (auto z : { bbox.z.min, bbox.z.max }) {
 
-                    auto newx =  cos_theta*x + sin_theta*z;
-                    auto newz = -sin_theta*x + cos_theta*z;
-
-                    vec3 tester(newx, y, newz);
+                    vec3 tester {
+                         cos_theta*x + sin_theta*z,
+                         y,
+                        -sin_theta*x + cos_theta*z
+                    };
 
                     for (int c = 0; c < 3; c++) {
                         min[c] = fmin(min[c], tester[c]);
--- a/src/TheNextWeek/hittable.h
+++ b/src/TheNextWeek/hittable.h
@@ -116,14 +116,17 @@ class rotate_y : public hittable {
 
     bool hit(const ray& r, interval ray_t, hit_record& rec) const override {
         // Change the ray from world space to object space
-        auto origin = r.origin();
-        auto direction = r.direction();
+        point3 origin {
+            cos_theta*r.origin().x() - sin_theta*r.origin().z(),
+            r.origin().y(),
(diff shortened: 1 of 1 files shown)


hollasch:
Sorry, but we are not accepting unplanned changes right now as book 2 is in alpha.2 release preparation, and after that we are focusing on book 3 final changes for v4.0.0. Please see CONTRIBUTING.md for our contribution guidelines. (I plan to add a blocking information to our PR pipeline on GitHub to refer to our contribution guidelines, but have been focusing on our release at the moment; sorry about that.)

For the first change, I don't find that to be meaningfully more clear than what is currently in the text.

For the second change, we avoid a lot of modern C++ idioms in order to keep the code easily understandable for non-C++ programmers. Like you, I would definitely use ranges and many other modern C++ conveniences in my own ray tracing code.

That said, I suppose I haven't looked closely at this code — I agree with you that it _definitely_ needs a refactoring, particularly in light of our new `interval` class. I also find the current code awkward with 0/1 multiplication used to select values.

Please create an issue for this and I'll add it to our post v4.0.0 backlog. Thank you!

dimitry-ishenko:
@hollasch while I understand you not wanting to merge this because you have other priorities, I completely disagree with your assessment.

> For the first change, I don't find that to be meaningfully more clear than what is currently in the text.

Throughout the book we use _x_, _y_ and _z_ axes and not 0, 1, 2. Not sure how you don't see it as more clear... Plus, who's to say that the axes are ordered _x_, _y_, _z_? At the very least maybe add an enum, eg:
```c++
enum axis { X = 0, Y = 1, Z = 2 };
```
and use that instead of hard-coding axis numbers.

> For the second change, we avoid a lot of modern C++ idioms in order to keep the code easily understandable for non-C++ programmers. Like you, I would definitely use ranges and many other modern C++ conveniences in my own ray tracing code.

Modern C++ idiom?... Range-based for-loops have been part of C++ since 2011. It's 2024. You would rather use some arcane math to select different struct members (which by the way is not even explained anywhere) instead of using "modern" C++ idiom, which exists in most programming languages nowadays?

While I understand this is a course about ray tracing and not programming, you are teaching people bad programming practices.

armansito:
I wouldn't go so far as to say that the book is teaching bad programming practices (I disagree with that notion) but I do think that the way @dimitry-ishenko restructured the loop is more readable than what's in the current text (specifically the `auto x = i*bbox.x.max + (1-i)*bbox.x.min;` bit that selects between min and max based on the index).

I think this is a change worth reconsidering post-release.

dimitry-ishenko:
> I wouldn't go so far as to say that the book is teaching bad programming practices (I disagree with that notion)

Apologies, by _"bad"_ I didn't mean _"immoral or evil"_. My definition was more along the lines of _"not achieving an adequate standard"_. Maybe, _"poor"_ would have been a better word.

All in all, this is a great course, which I've learnt a lot from. And, I appreciate your choice of C++ for it. Unfortunately, there are a few cringe-worthy places IMHO (use of raw pointers in the `perlin` class also comes to mind). And, seeing this is a first encounter with C++ for many people coming here, I think it behooves to teach them good programming habits "from the cradle" so to speak. 😄 

Also, I don't quite understand @hollasch's desire to stay away from _"modern"_ C++ features. Many of these features were introduced to put C++ on par with other modern programming languages, and help avoid many common pitfalls. Plus, some of these features _are_ already being used in the course (shared pointers, `auto` keyword, etc.). So, why not others?

rupsis:
> I think it behooves to teach them good programming habits

This is largely subjective. Wouldn't teaching "good programming habits" be from first principle? Once you understand first principles, then you can leverage advanced features of the language. 

But I agree with @armansito, this bit of code could use some more love.

hollasch:
Yeah, that's why I asked for an issue to track this.

dimitry-ishenko:
> This is largely subjective. Wouldn't teaching "good programming habits" be from first principle? Once you understand first principles, then you can leverage advanced features of the language.

@rupsis it's not subjective. It's common sense. People come here for learn first principles about ray tracing. If they are not familiar with C++, they end up learning first principles about C++ as well. Unless you want them to go and do a crash course on C++ before coming here, they will learn those principles as laid out in this course.

PR #1481, opened 2024-03-29
More updates to book 1

hollasch:


books/RayTracingInOneWeekend.html
```diff
@@ -1456,8 +1455,8 @@
 
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ delete
-    using std::shared_ptr;
     using std::make_shared;
+    using std::shared_ptr;
```
rupsis, line 1459: These don't need to be flipped. When introduced in `The hittable_list class` (Listing 21) 
they are ordered: 
```
using std::shared_ptr;
using std::make_shared;
```

books/RayTracingInOneWeekend.html
```diff
@@ -2139,6 +2146,9 @@
         double pixel_samples_scale;  // Color scale factor for a sum of pixel samples
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
         point3 center;               // Camera center
+        point3 pixel00_loc;          // Location of pixel 0, 0
+        vec3   pixel_delta_u;        // Offset to pixel to the right
+        vec3   pixel_delta_v;        // Offset to pixel below
```
rupsis, line 2151: For my own edification, what is the ruling on (if there is one) contextual anchor points vs code collapse? 

For instance, `pixel_samples_scale` is added (and highlighted) after `image_height`. So we know were in the code to add it. 

Is there a benefit to adding the rest private definitions here? 

if we do decide to add them, we need to remove the `...` between these and `void initialize() {` as there isn't any code there.

books/RayTracingInOneWeekend.html
```diff
@@ -3113,6 +3120,21 @@
 
 </div>
 
+Now we'll update the `sphere` constructor to take a material.
```
rupsis: Nit: "Now we'll update the `sphere` constructor to take a material pointer." might be more inline with the `TODO` comment.

rupsis (approved):

books/RayTracingInOneWeekend.html
hollasch: Hmm, some of these reorderings must have been dropped. Since adding a new using statement for fabs here, the random order looked more annoying, so I'll grow this in sorted order. Will update.

books/RayTracingInOneWeekend.html
hollasch: It's subjective. Here the ellipsis was just hiding three lines of code context, which when updating the code wasn't really worth saving just two lines. It was just in response to the speedbump I hit while updating.

I'll elide the now unnecessary ellipsis.

books/RayTracingInOneWeekend.html

PR #1501, opened 2024-04-06
Fix green texmap artifacts on earth render

### LINKED ISSUE #1485: [Book 2] Artifacts on texture mapped sphere in recent progression updates
The cover image uploaded in #1468 shows green blotchy artifacts on parts of the texture mapped sphere that aren't present in earthmap.jpg (and weren't present in the image before). The discoloration is most prominent over the Sahara and Arabian deserts:

Screenshot 2024-03-29 at 9 47 02 PM

hollasch:
Turns out the problem has to do with the conversion to gamma-corrected intensities from the [0,1] values read in from the STB image library.

I thought the range would be [0,1), but instead it's [0,1]. Thus, intensity values of 1.0 were mapped to 256, which then converted to 0 for unsigned 8-bit chars. Thankfully, I typo'ed the conversion in the latest book 2 progression, and multiplied intensities by _257_, which made the problem worse, and enough to to hit me.

This fix makes the conversion bullet-proof, and zero or negative intensities are now mapped to 0, and values greater than or equal to 1.0 are now mapped to 255.

Resolves #1485

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -1797,6 +1797,14 @@ header into a folder called `external`. Adjust according to your directory struc
             return high - 1;
         }
 
+        static unsigned char float_to_byte(float value) {
+            if (value <= 0.0)
+                return 0;
+            if (1.0 <= value)
+                return 255;
+            return static_cast< unsigned char >(256.0 * value);
+        }
+
         void convert_to_bytes() {
             // Convert the linear floating point pixel data to bytes, storing the resulting byte
             // data in the `bdata` member.
@@ -1810,7 +1818,7 @@ header into a folder called `external`. Adjust according to your directory struc
             auto *bptr = bdata;
             auto *fptr = fdata;
             for (auto i=0; i < total_bytes; i++, fptr++, bptr++)
-                *bptr = static_cast< unsigned char >(*fptr * 256.0);
+                *bptr = float_to_byte(*fptr);
         }
     };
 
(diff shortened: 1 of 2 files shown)


books/RayTracingTheNextWeek.html
```diff
@@ -1810,7 +1818,7 @@
             auto *bptr = bdata;
             auto *fptr = fdata;
             for (auto i=0; i < total_bytes; i++, fptr++, bptr++)
-                *bptr = static_cast< unsigned char >(*fptr * 256.0);
+                *bptr = float_to_byte(*fptr);
```
rupsis, line 1821: instead of making a new method, couldn't we just use the clamp defined above?

```
*bptr = static_cast< unsigned char >(clamp(256.0 * value, 0, 255));
```

hollasch: Hmmm. The existing `clamp()` function clamps 32-bit integers. So I could first convert to `int`, though I'm not sure what values less than -2.147483648e9 would convert to, nor what values greater than +2.147483647e9 would convert to. These are susceptible to the same over/underflow results that the 8-bit values displayed. I do know that HDR values are supported, so we're not restricted to [0,1) or [0,1]. Thus, I'd have to perform some clamping _before_ safely passing values to the `int` clamping function, and then once more to clamp _those_ values to the unsigned 8-bit clamp, to scavenge the input signed `32-bit` values.

In the end, I think it's simpler and easier to just clamp the floating-point values to [0,1] (via the `if` statements), and then manage the conversion of _that_ range to [0,255] (through the division into 256 segments and subsequent implicit $\lfloor{x}\rfloor$ function).

hollasch:
Thankfully, this whole file is just plopped down like a pseudo black box, without really having to explain any specific part of it.

rupsis (approved):

PR #1503, opened 2024-04-06
Image updates for book 1

hollasch:


armansito (approved):

images/img-1.23-book1-final.jpg
rupsis, line 1: Was this re-rendered? it has zero difference from the previous image. If we want to keep this image the same, that's fine by me. But considering everything apart from the 3 center spheres have _some_ element of randomness, a re-render wouldn't produce the same result. Unless I'm missing something?

hollasch: Yeah, good point. I'll rebase this to drop the final render, and then merge it in.

hollasch:
Rebased to drop the unchanged final image, committed to `dev` in 5188fb36ba63ce2bc273d0749c45a0f137a71b4d.

PR #1508, opened 2024-04-07
Fix bad hittable_list::hit() signature

### LINKED ISSUE #1507: [dev] Book 2, Listing 14: stale `hittable_list::hit` signature
The listing shows `bool hit(const ray& r, double ray_tmin, double ray_tmax, hit_record& rec) const override` but it should show `bool hit(const ray& r, interval ray_t, hit_record& rec) const override` (a change introduced in the book 1).

hollasch:
Resolves #1507


armansito (approved):
Thanks!

Issue #1312, opened 2023-10-27
Listing 3.29 lacks proper change highlighting


hollasch:
Benny Tsang writes:
> (Book 3) Section 8.3, the addition of the argument `double& pdf` in scatter() is not highlighted in Listing 29. Also, the corresponding changes in other materials are also not shown.

hollasch:
For "corresponding changes in other materials", see #1019.

rupsis:
Resolved in #1514

PR #1514, opened 2024-04-08
Adding in scatter method signature change for all materials

rupsis:
Addresses https://github.com/RayTracing/raytracing.github.io/issues/1312

### DIFF
--- a/books/RayTracingTheRestOfYourLife.html
+++ b/books/RayTracingTheRestOfYourLife.html
@@ -2348,6 +2348,57 @@ We can rewrite our Lambertian material using this to get:
 
         ...
     };
+
+    class metal : public material {
+        public:
+            ...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
+        bool scatter(
+            const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered, double& pdf
+        ) const override {
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+             ...
+        }
+    }; 
+
+    class dielectric : public material {
+        public:
+            ...
+
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
+        bool scatter(
+            const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered, double& pdf
+        ) const override {
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
+             ...
+        }
+    }; 
+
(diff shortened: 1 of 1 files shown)


rupsis:
So I'm a bit out of my depth here. This is my first time progressing though Book 3, and this is all new to me. 

Obviously with the `virtual bool scatter` method signature change in Listing 28, the rest of the materials need to update their `scatter` signature as well. Thats easy enough, and what the changes currently are. What I'm not sure about, is the `material::scatter()` call in `camera::get_color`. 

Since we're taking a pdf reference, and assigning a pdf value based on the material it would make sense to pass in a `pdf` from the camera and use that:
```c++
    // camera.h
    color ray_color(const ray &r, int depth, const hittable &world) const {
    ...
    
    double pdf = 0.0;

    if (!rec.mat->scatter(r, rec, attenuation, scattered, pdf)) {
      return color_from_emission;
    }

    double scattering_pdf = rec.mat->scattering_pdf(r, rec, scattered);
    // double pdf = scattering_pdf;

    color color_from_scatter = (attenuation * scattering_pdf *
                                ray_color(scattered, depth - 1, world)) /
                               pdf;

    return color_from_emission + color_from_scatter;
    }
```

This results in a ton of fireflies:
image6_pdf


Now, if we just use a throwaway variable:
```c++
    // camera.h
    color ray_color(const ray &r, int depth, const hittable &world) const {
    
    double throwaway = 0.0;

    if (!rec.mat->scatter(r, rec, attenuation, scattered, throwaway)) {
      return color_from_emission;
    }

    double scattering_pdf = rec.mat->scattering_pdf(r, rec, scattered);
    double pdf = scattering_pdf;
    ...
    }
```

We seem to get similar results to the current Image 6:
image6_throwaway

Do we know what the correct update to `camera::get_color` is suppose to be?

armansito:
I haven't debugged this myself but I suspect from what you're describing that the pdf returned by `material::scatter` and `matterial::scattering_pdf` for the same scattering direction are not equal (as they were expected to be from the narrative). Can you double check how those two functions compute their pdfs?

I have a more meta question: what is the `pdf` returned from `material::scatter` supposed to represent? Isn't that the same as the scattering PDF? If it is, then is it still necessary to have a separate `scattering_pdf` function?

hollasch:
@trevordblack needs to weigh in on this.

hollasch (approved):

hollasch:
This PR doesn't resolve the underlying issue, but is a definite step in the right direction.

rupsis:
> I haven't debugged this myself but I suspect from what you're describing that the pdf returned by `material::scatter` and `matterial::scattering_pdf` for the same scattering direction are not equal (as they were expected to be from the narrative). Can you double check how those two functions compute their pdfs?

@armansito 
Setting the `scattering_pdf` to the `pdf` returned from the lambertian material (orthonormal basis scatter) produces this: 

image6_same_pdf

Which is certainly an improvement over the fireflies!  

> I have a more meta question: what is the pdf returned from material::scatter supposed to represent? Isn't that the same as the scattering PDF? If it is, then is it still necessary to have a separate scattering_pdf function?

For the orthonormal basis example, it would make more sense to remove the `pdf` parameter from the `scatter` function, and update the `scattering_pdf` to use `onb`. Means we wouldn't have to update the `scatter` method for all of the other materials. 

Something else interesting in **Listing 30**. We declare a new `double pdf;` in `ray_color`, then it get's set by the emission check:
```
 if (!rec.mat->scatter(r, rec, attenuation, scattered, pdf))
            return color_from_emission;
```

But _then_ we just override it before it gets used:
```
 pdf = distance_squared / (light_cosine * light_area)
```

hollasch:
@rupsis — this commit resolves #1312. Please convert this to a true PR and let's at least get this in and close that issue.

hollasch (approved):

PR #1520, opened 2024-04-13
Use explicit std:: namespace almost everywhere

### LINKED ISSUE #1487: Add missing std:: namespace
There are a number of small functions throughout the course that are referred to in global namespace, for example:

* `fmin`, `fmax`
* `sqrt`
* `fabs`
* `sin`, `cos`, `tan`
* `rand`

Two of them -- `sqrt` and `fabs` -- are pulled into global namespace with the `using` directive inside `rtweekend.h`, but others -- aren't. Two possible solutions:

1. Prefix all functions with `std::` namespace (preferred).
2. Pull missing functions into global namespace (poor practice).

hollasch:
When referring to C++ standard library functions, use the explicit `std::` namespace qualifier. The only exception is that we bring `shared_ptr` and `make_shared` into the global namespace in order to keep the code brief, since we have a 96-character line limit in our source code.

Currently, this covers the following identifiers:

    acos        fixed          numeric_limits    sqrt
    atan2       floor          ostream           string
    cerr        flush          rand              tan
    clog        fmax           setprecision      vector
    cos         fmin           shared_ptr
    cout        log            sin
    fabs        make_shared    sort

Resolves #1487


books/RayTracingTheRestOfYourLife.html
```diff
@@ -650,12 +650,12 @@
...
 actually a lot less work than the Monte Carlo. That might be true in the case where the function is
 $f(x) = x^2$, but there exist many functions where it might be simpler to solve for the Monte Carlo
-than for the integration, like $f(x) = sin^5(x)$.
+than for the integration, like $f(x) = \sin^5(x)$.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
         for (int i = 0; i < N; i++) {
             auto x = random_double(a, b);
-            sum += pow(sin(x), 5.0);
+            sum += pow(std::sin(x), 5.0);
```
rupsis: Should `pow` from <cmath> also use explicit std::?

rupsis (approved):

books/RayTracingTheRestOfYourLife.html
hollasch: Good spot; I'll add it.

PR #1545, opened 2024-04-23
Updates from book 3 complete progression

### LINKED ISSUE #988: Book 3.10.1: `camera::ray_color()` switches from `double pdf` to `double pdf_val` halfway through
Pick one of the above and make it consistent.

### LINKED ISSUE #1317: Listing 3.40: Update code listing change highlighting
Benny Tsang writes:
> - (Book 3) Listing 40: `ray_color`'s function signature should also be highlighted since we have added hittable `lights`. Also `light_ptr` in the `hittable_pdf` constructor should be `lights`.

hollasch:
Also includes some minor fixes.

Note that this shows the current state of some buggy renders, which render the entire scene as black, except for the single square light source.

Resolves #988
Resolves #1317


src/TheRestOfYourLife/pdf.h
```diff
@@ -1,15 +1,5 @@
...
-//==============================================================================================
-// Originally written in 2016 by Peter Shirley <ptrshrl@gmail.com>
-//
-// To the extent possible under law, the author(s) have dedicated all copyright and related and
-// neighboring rights to this software to the public domain worldwide. This software is
-// distributed without any warranty.
-//
-// You should have received a copy (see file COPYING.txt) of the CC0 Public Domain Dedication
-// along with this software. If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
-//==============================================================================================
```
hollasch: Fixed in next commit.

books/RayTracingTheRestOfYourLife.html
```diff
@@ -2351,59 +2349,88 @@
...
             ...
-            
+
+            ray scattered;
+            color attenuation;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
-        bool scatter(
-            const ray& r_in, const hit_record& rec, color& attenuation, ray& scattered, double& pdf
-        ) const override {
+            double pdf_value; // TODO: What are we supposed to do with the returned PDF?
```
rupsis, line 2415: Is this TODO for the reader?

hollasch: No, things are bad enough at this point that I wanted to make sure all instances of this problematic code are flagged, so we don't miss a copy before shipping. It's also why I've updated images with obviously bad renders.

rupsis (approved):
Slowly getting better!

PR #1551, opened 2024-04-26
Standardize our use of begin/end iterators

hollasch:
We try to keep our use of C++ dead simple for non-C++ programmers to read easily, but there are cases where we need to yield a bit. (For example, using shared pointers to keep memory management simple.)

In order to use the std::sort() function, we need to provide or use begin and end iterators. With the recent change to our `estimate_halfway.cc` program, we had different ways of using such iterators.

The BVH code used the member functions `begin()` and `end()`, while the `estimate_halfway` program used argument promotion from an array and the array plus an integer offset.

To standardize both and reduce variance, this change uses the currently-recommended practice of using `std::begin(thing)` and `std::end(thing)`.

For now, I'm avoiding spending any text explaining to the non-C++ reader what these are, in the hopes that the meaning can be easily deduced.

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -989,7 +989,7 @@ we haven't yet defined.
                 left = objects[start];
                 right = objects[start+1];
             } else {
-                std::sort(objects.begin() + start, objects.begin() + end, comparator);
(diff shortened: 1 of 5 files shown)


rupsis:
We need the corresponding book changes as well?

hollasch:
Gah!

rupsis (approved):

PR #1552, opened 2024-04-27
Update RayTracingInOneWeekend.html to better reflect the image

jenshz:
The code as written does not generate the picture in the book - this confused me, and had me thinking I made some kind of error in the code. I do not know which values were used to generate the original picture, but these seem close.


jenshz:
This does not reflect the right aperture. I will have a look at the values one more time.

PR #1561, opened 2024-05-20
[Book2] Add missing word in comment of Listing 53

### LINKED ISSUE #1560: [Book2] Typo in Listing 53
```cpp
// Determine the hit point lies within the planar shape using its plane coordinates.
```
misses an `if`, as it probably should be 
```cpp
// Determine if the hit point lies within the planar shape using its plane coordinates.
```

FineFindus:
Fixes a typo in Book 2, Listing 53, by adding a missing `if` in the comment.
Resolves #1560


hollasch:
Thanks!

hollasch (approved):

PR #1602, opened 2024-08-16
Fix broken highlighting on some code listings

### LINKED ISSUE #1600: Book 1: Highlighting missing due to mis-matched casing
I ran across a listing in the first book (41, in section 8.1, FWIW), where I expected highlighting, but it was missing.
In the source a string of tildes followed by "C++ Highlight" appears to mark a section as class "Highlight" (upper-case _h_).
However, the proper class is "highlight" (lower-case _h_), so "C++ highlight" would seem to be the correct annotation.
This only seems to be the case in the first book, and it contains 16 instances that should be changed.
I am fine creating the PR if desired.

whydoubt:
Resolves #1600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Change Log / Ray Tracing in One Weekend
 
 ### In One Weekend
   - Fix    -- Fixed usage of the term "unit cube" for a cube of diameter two (#1555)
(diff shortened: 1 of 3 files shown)


hollasch (approved):

hollasch:
Thanks Jeff!

PR #1609, opened 2024-08-19
Include hittable.h from material.h

### LINKED ISSUE #1608: material.h should include hittable.h
Currently in `material.h` we have a forward declaration of `class hit_record`. However, this line is unnecessary because in `main.cc`, we include `hittable.h` (where `hit_record` is defined) before we include `material.h`. Further, the class declaration is insufficient, as there are multiple member dereferences throughout `material.h`, which requires the `hit_record` definition.

The solution to this is to include `hittable.h` in the beginning of `material.h`. The current situation likely came about as I wrestled with the potential circular dependency where material classes needed access to `hit_record`, and `hit_record` has a shared pointer to a `material` instance. The current `class material` declaration at the top of `hittable.h` should be sufficient to address this.

hollasch:
The prior version used a forward declaration of `hit_record`, which was insufficient since hittable.h lines referenced members of `hit_record`, and was unnecessary because hittable.h was included before material.h in practice.

For this specific code, the inclusion is unnecessary, but is proper form.

Resolves #1608


hollasch:
@whydoubt — Jeff, I've added you on this review in case you're interested in helping out with PR reviews. Let me know if you are, and I'll add you to the RTW reviewers' group.

whydoubt (approved):

PR #1619, opened 2024-08-23
Clarify difference of average vs expected value

### LINKED ISSUE #1535: Book 3.3.1: Average vs Expected value
The two terms—_average_ and _expected value_—are used somewhat interchangeably in the book (mainly chapters 3.1 and 3.2) even though they are, strictly speaking, two different things. For example, if I have 6-sided die, my _expected value_ is always:

$$E[X] = 1 \cdot \frac 1 6 + 2 \cdot \frac 1 6 + 3 \cdot \frac 1 6 + 4 \cdot \frac 1 6 + 5 \cdot \frac 1 6 + 6 \cdot \frac 1 6 = 3.5$$

But, my _average_ can be anything between 1 and 6. If I roll the die 10 times and get this sequence: 1 1 4 3 6 4 2 6 2 1, my _average_ is:

$$\bar X = \frac {1 + 1 + 4 + 3 + 6 + 4 + 2 + 6 + 2 + 1} {10} = \frac {30} {10} = 3$$

The only thing that links the two values is the [Law of Large Numbers](https://en.wikipedia.org/wiki/Law_of_large_numbers), which states that _average_ converges to the _expected value_ as the number of samples increases.

Maybe some mention of the law can be made in the book and a few formulas tweaked where we equal the two.

hollasch:
Resolves #1535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Change Log / Ray Tracing in One Weekend
 ### The Rest of Your Life
   - Fix    -- Fix typo of "arbitrary" (#1589)
   - New    -- Added a bit more explanation of Buffon's needle problem (#1529)
+  - Fix    -- Clarify the distinction between average and expected value (#1535)
 
 
 ----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


hollasch:
@dimitry-ishenko — let me know what you think.

dimitry-ishenko:
Thanks @hollasch. Excellent explanation.

PR #1620, opened 2024-08-23
Correct range of scatter angle theta in sec 3.3.5

### LINKED ISSUE #1331: Book 3.5.3: pScatter(theta_o < 0) = 0 is incorrect
In chapter 5.3. The scattering PDF, when we integrate over the hemisphere, it is written that $pScatter(\theta_o<0) = 0$. I believe the correct one should be $pScatter(\pi/2 < \theta_o < \pi) = 0$ (note that $\theta_o \in [0, \pi]$). The equation below that is correct. Since $pScatter(\pi/2 < \theta_o < \pi) = 0$, we only need to integrate $\theta_o$ from 0 to $\pi/2$.

Furthermore, to help with understanding, should we add the variable name in the integral sign? i.e. replace $\int_0^{2\pi} \int_0^{\pi/2}$ with $\int_{\phi=0}^{2\pi} \int_{\theta=0}^{\pi/2}$.

hollasch:
Resolves #1331


hollasch:
@gau-nernst

Issue #1621, opened 2024-08-25
Book 2: improve center_vec name


whydoubt:
As we know, naming variables is one of the hardest things...

In listing 2.3, `center_vec` is used to indicate the change
in position over an interval of time.  I don't feel the name
conveys that well.  Also, there is a well-established name
for that concept: velocity.  So I would propose `velocity`
as a replacement.  Note that this would also need to be
changed in listings 2.10 and 2.11.

Also in listing 2.11... Some underscores appeared where
they weren't before: `center_vec = _center2 - _center1;`
Looks like it may have been a leftover from an earlier
revision of the book.

whydoubt:
A step further could be to pass velocity instead of center2... cutting out the intermediary.
I haven't yet checked if that would complicate anything further down in the text though.
I could put together a PR to this effect if the idea seems good.

whydoubt:
If you only want to consider the change in position, without the hard implication that velocity is constant, `displacement` could be another alternative for `center_vec`.

PR #1623, opened 2024-08-26
DRAFT: idea for moving_sphere refactor

hollasch:
Idea from issue #1621


whydoubt (commented):
Wasn't so sure at first, but using a ray to represent the motion is growing on me.
Being able to use the provided interpolation is certainly handy.   :+1:

hollasch:
Cool. I'll finish the rest of this for future PR.

PR #1625, opened 2024-08-27
Reword spherical coords in book 3, chap 4, para 3

### LINKED ISSUE #1536: Book 3.4: Mapping unit sphere to 2D space
3rd paragraph in chapter 4 is not very clear IMHO:

> Every direction in 3D space has an associated point on the unit sphere and can be generated by solving for the vector that travels from the origin to that associated point. You can think of choosing a random direction as choosing a random point in a constrained two dimensional plane: the plane created by mapping the unit sphere to Cartesian coordinates.

My understanding of this is:

In the [spherical coordinate system](https://en.wikipedia.org/wiki/Spherical_coordinate_system) each point on a sphere can be represented by $r$, $\theta$ and $\phi$ (radial distance, polar angle and azimuth). Since $r = 1$ for a unit-sphere, we only need to map the two angles in a constrained two-dimensional space, where $\theta \in [0, \pi]$ and $\phi \in [0, 2\pi]$.

hollasch:
Resolves #1536


hollasch:
@dimitry-ishenko

hollasch:
Looking at this again, I think I need more changes to this chapter, as the changes above don't flow well with following paragraphs. Closing out this PR until I address this.

PR #1627, opened 2024-08-28
Refactor/moving sphere

### LINKED ISSUE #1621: Book 2: improve center_vec name

hollasch:
Resolves #1621

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -176,56 +176,41 @@ Adding Moving Spheres
 ----------------------
 Now to create a moving object. I’ll update the sphere class so that its center moves linearly from
 `center1` at time=0 to `center2` at time=1. (It continues on indefinitely outside that time
-interval, so it really can be sampled at any time.)
+interval, so it really can be sampled at any time.) We'll do this by replacing the 3D center point
+with a 3D ray that describes the original position at time=0 and the displacement to the end
+position at time=1.
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class sphere : public hittable {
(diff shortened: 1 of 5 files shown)


hollasch:
Might take this one in for v4.0.1, otherwise I'll merge for v4.0.2.

books/RayTracingTheNextWeek.html
```diff
@@ -176,70 +176,67 @@
...
 
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++
     class sphere : public hittable {
       public:
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ C++ highlight
         // Stationary Sphere
-        sphere(const point3& center, double radius, shared_ptr<material> mat)
-          : center1(center), radius(std::fmax(0,radius)), mat(mat), is_moving(false) {}
+        sphere(const point3& static_center, double radius, shared_ptr<material> mat)
+          : center(static_center, vec3(0,0,0)), radius(std::fmax(0,radius)), mat(mat) {}
```
whydoubt, line 189: Could create the null vector with just `vec3()`, but explicit components are good too.

whydoubt (approved):
I'm fine either way on initializing the null vector.

PR #1629, opened 2024-08-29
Headers assume implicit rtweekend.h include

### LINKED ISSUE #1628: Inclusion of rtweekend.h still unclear
1.6.7 states "All main program files will include rtweekend.h first, so most other header files (where the bulk of our code will reside) can assume these definitions are already available."

I read that as: `main.cc` includes it up top, so we don't have to include it elsewhere.  However, the source code is inconsistent with that premise, but the source code is also inconsistent with itself.  Just looking at the other files mentioned in section 1.6.7 as needing changes:
- `color.h`, `sphere.h`, `vec3.h` in the `src` folder do NOT include `rtweekend.h`
- `hittable.h`, `hittable_list.h` in the `src` folder DO include `rtweekend.h`

Additionally, `camera.h` and `material.h`, which are introduced later, both include `rtweekend.h` in both the book and the `src` folder.

This effects of this carry over to the other books.

hollasch:
Resolves #1628


hollasch:
@hollasch — **Fix commit comment typo on merge**

hollasch:
Forgot changelog. Also since I want to fix the original typo, I'm going to toss this PR and replace it with a corrected one.

Issue #1668, opened 2025-01-04
Non skip pdf in lights HittableList


niccolot:
Hello, if i add to the lights list an object that is not a metal or dielectric the image produced is like

![pdfmaterial](https://github.com/user-attachments/assets/75c1878f-d34c-4475-a417-8241110e11c9)

i' ve obtained this by putting 

```c++
// in the original main is glass
auto nonglass = make_shared<lambertian>(color(0.8, 0.2, 0.2));
world.add(make_shared<sphere>(point3(190,90,190), 90, nonglass));

// Light Sources
hittable_list lights;
auto m = shared_ptr<material>();
lights.add(make_shared<quad>(point3(343,554,332), vec3(-130,0,0), vec3(0,0,-105), m));
lights.add(make_shared<sphere>(point3(190, 90, 190), 90, m));
```
in the main.cpp of therestofyourlife part.

I may not have understood correctly this part of the code and thus it may not be an actual issue but still i wanted to ask this since also a material like a lambertian is still emitting some reflected light

niccolot:
I' ve looked into it a bit and found out that what's causing the problem is the sphere part of the HittablePDF list belonging to lights:

```c++
// main.cpp
HittableList lights;
lights.add(std::make_shared<Quad>(Vec3(213, 554, 227), Vec3(130, 0, 0), Vec3(0, 0, 105), empty_mat));
lights.add(std::make_shared<Sphere>(Vec3(190, 90, 190), 90, empty_mat));
```

if i select just the sphere part

```c++
// camera.cpp
Color color_from_scatter = Color();
    for (const auto& ray_t : srec.scattered_rays) {
        if (ray_t.skip_pdf || ray_t.pdf == nullptr) {
            color_from_scatter += ray_t.attenuation * ray_color(ray_t.skip_pdf_ray, depth-1, world, lights);
        } else {
            auto light_ptr = std::make_shared<HittablePDF>(lights, rec.point()); // lights is the problematic part
            MixturePDF p(light_ptr, ray_t.pdf); 
            Ray scattered = Ray(rec.point(), p.generate(), r.time());
            double pdf_value = p.value(scattered.direction(), r.direction(), -w);
            double scattering_pdf = rec.material()->scattering_pdf(r, rec, scattered, -w);
            Color sample_color = ray_color(scattered, depth-1, world, lights);
                
            color_from_scatter += (ray_t.attenuation * scattering_pdf * sample_color) / pdf_value;
        }
    }
[...]

PR #1674, opened 2025-02-06
now nonskippdf materials can be added to lights hittable list

niccolot:
Hello, i' ve modified the pdf_value function for the sphere and the random hittable_list method to include controls for NaNs and zero division, now even materials like a lambertian can be added to the lights hittable list. To see the details and the changes refer to the relative issue [](https://github.com/RayTracing/raytracing.github.io/issues/1668)(https://github.com/RayTracing/raytracing.github.io/issues/1668), i hope this could be usefull.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 
 build/
 /*.ppm
+
+.vscode/
--- a/src/TheRestOfYourLife/hittable_list.h
+++ b/src/TheRestOfYourLife/hittable_list.h
@@ -59,9 +59,14 @@ class hittable_list : public hittable {
         return sum;
     }
 
-    vec3 random(const point3& origin) const override {
-        auto int_size = int(objects.size());
-        return objects[random_int(0, int_size-1)]->random(origin);
+       vec3 random(const point3& origin) const override {
+        int int_size = int(objects.size());
+        auto vec = objects[random_int(0, int_size-1)]->random(origin);
+        while (vec.is_nan()) {
+            vec = objects[random_int(0, int_size-1)]->random(origin);
+        }
+
+        return vec;
     }
 
   private:
--- a/src/TheRestOfYourLife/main.cc
+++ b/src/TheRestOfYourLife/main.cc
@@ -44,7 +44,8 @@ int main() {
 
     // Glass Sphere
     auto glass = make_shared<dielectric>(1.5);
-    world.add(make_shared<sphere>(point3(190,90,190), 90, glass));
+    auto nonglass = make_shared<lambertian>(color(0.8, 0.2, 0.4));
+    world.add(make_shared<sphere>(point3(190,90,190), 90, nonglass));
 
     // Light Sources
     auto empty_material = shared_ptr<material>();
@@ -56,9 +57,9 @@ int main() {
     camera cam;
 
     cam.aspect_ratio      = 1.0;
-    cam.image_width       = 600;
-    cam.samples_per_pixel = 100;
-    cam.max_depth         = 50;
+    cam.image_width       = 400;
+    cam.samples_per_pixel = 1000;
+    cam.max_depth         = 10;
     cam.background        = color(0,0,0);
 
     cam.vfov     = 40;
--- a/src/TheRestOfYourLife/sphere.h
+++ b/src/TheRestOfYourLife/sphere.h
@@ -76,11 +76,14 @@ class sphere : public hittable {
         if (!this->hit(ray(origin, direction), interval(0.001, infinity), rec))
             return 0;
 
-        auto dist_squared = (center.at(0) - origin).length_squared();
-        auto cos_theta_max = std::sqrt(1 - radius*radius/dist_squared);
+        auto cos_theta_max = sqrt(1 - radius*radius/(center.at(0) - origin).length_squared());
         auto solid_angle = 2*pi*(1-cos_theta_max);
 
-        return  1 / solid_angle;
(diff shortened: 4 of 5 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@
 
 build/
 /*.ppm
+
+.vscode/
```
hollasch, line 6: This is an artifact of your own personal environment, so it belongs in your personal global .gitconfig file (along with other things like editor, tool, tagging, OS and other artifacts that come from your general environment and toolsets).

src/TheRestOfYourLife/main.cc
```diff
@@ -44,7 +44,8 @@ int main() {
 
     // Glass Sphere
     auto glass = make_shared<dielectric>(1.5);
-    world.add(make_shared<sphere>(point3(190,90,190), 90, glass));
+    auto nonglass = make_shared<lambertian>(color(0.8, 0.2, 0.4));
+    world.add(make_shared<sphere>(point3(190,90,190), 90, nonglass));
```
hollasch, line 48: This seems to be a testing artifact that made it into your PR.

src/TheRestOfYourLife/main.cc
```diff
@@ -56,9 +57,9 @@ int main() {
     camera cam;
 
     cam.aspect_ratio      = 1.0;
-    cam.image_width       = 600;
-    cam.samples_per_pixel = 100;
-    cam.max_depth         = 50;
+    cam.image_width       = 400;
+    cam.samples_per_pixel = 1000;
+    cam.max_depth         = 10;
```
hollasch, line 62: More testing artifacts.

src/TheRestOfYourLife/sphere.h
```diff
@@ -76,11 +76,14 @@ class sphere : public hittable {
         if (!this->hit(ray(origin, direction), interval(0.001, infinity), rec))
             return 0;
 
-        auto dist_squared = (center.at(0) - origin).length_squared();
-        auto cos_theta_max = std::sqrt(1 - radius*radius/dist_squared);
+        auto cos_theta_max = sqrt(1 - radius*radius/(center.at(0) - origin).length_squared());
```
hollasch, line 79: I don't understand the motivation for this change. It doesn't seem related to the PR purpose.

src/TheRestOfYourLife/hittable_list.h
```diff
@@ -59,9 +59,14 @@ class hittable_list : public hittable {
         return sum;
     }
 
-    vec3 random(const point3& origin) const override {
-        auto int_size = int(objects.size());
-        return objects[random_int(0, int_size-1)]->random(origin);
+       vec3 random(const point3& origin) const override {
+        int int_size = int(objects.size());
```
hollasch, line 63: Why the change to make this a double declaration?

src/TheRestOfYourLife/hittable_list.h
```diff
@@ -59,9 +59,14 @@ class hittable_list : public hittable {
         return sum;
     }
 
-    vec3 random(const point3& origin) const override {
-        auto int_size = int(objects.size());
-        return objects[random_int(0, int_size-1)]->random(origin);
+       vec3 random(const point3& origin) const override {
```
hollasch, line 62: Indentation error.

src/TheRestOfYourLife/hittable_list.h
```diff
@@ -59,9 +59,14 @@ class hittable_list : public hittable {
...
+        int int_size = int(objects.size());
+        auto vec = objects[random_int(0, int_size-1)]->random(origin);
+        while (vec.is_nan()) {
+            vec = objects[random_int(0, int_size-1)]->random(origin);
+        }
```
hollasch, line 67: This pattern calls for a do-while loop to eliminate the code duplication.

src/TheRestOfYourLife/sphere.h
```diff
@@ -76,11 +76,14 @@ class sphere : public hittable {
...
         auto solid_angle = 2*pi*(1-cos_theta_max);
 
-        return  1 / solid_angle;
+        if (std::isnan(solid_angle) || std::fabs(solid_angle) < 1e-3) {
```
hollasch, line 82: What is the reasoning for such a large minimum value of `solid_angle`? `1e-3` eliminates over half of all representable finite floating-point numbers.

src/TheRestOfYourLife/hittable_list.h
hollasch: More generally, a vector that contains a NaN value has issues before this code executes, so we still need to find the root cause for why this happens and fix it at the source. Addressing it here is too late.

src/TheRestOfYourLife/sphere.h
hollasch: Why would `solid_angle` be NaN here? The only way I can think of is if `cos_theta_max` is NaN. Why would that be?

→ `sqrt(1 - radius*radius/(center.at(0) - origin).length_squared())` must be NaN
→ `1 - radius*radius/(center.at(0) - origin).length_squared()` must be negative
→ `radius*radius/(center.at(0) - origin).length_squared()` must be greater than 1
→ `radius*radius/(center.at(0) - origin)` must be greater than 1
→ `radius*radius` must be less than `center.at(0) - origin`

Seems like this needs further digging to figure out why this is an issue.

hollasch (changes requested):

hollasch:
No response on pull request comments; closing.

PR #1682, opened 2025-02-19
Clarify random_double(min,max) implementation

### LINKED ISSUE #1681: `static` random distribution does not update the intervals
### Discussed in https://github.com/RayTracing/raytracing.github.io/discussions/1680

<div type='discussions-op-text'>

<sup>Originally posted by **niccolot** February 18, 2025</sup>
I' ve just realized that one of the suggested way to generate random numbers, namely

```cpp
inline double random_double(double min = 0, double max = 1) {
    /**
     * @brief returns a double in the range [min, max)
     */
    static std::uniform_real_distribution<double> distribution(min, max); // should not be static
    static std::mt19937 generator;

    return distribution(generator);
}
```

produces numbers only in the first (min,max) range it is instanciated with, e.g.

```cpp
int main()
{
    for (int i=0; i<5; ++i) {
        std::cout << random_double() << std::endl;
    }
    std::cout << "\n";
    for (int i=0; i<5; ++i) {
        std::cout << random_double(1,2) << std::endl;
    }
}
```

prints 

```
0.135477
0.835009
0.968868
0.221034
0.308167

0.547221
0.188382
0.992881
0.996461
0.967695
```
I think is kind of a coincidence that everything works regardless, the random_double() is always used with (0,1), the only problem could arise in 

```cpp
Vec3 HittableList::random(const Vec3& origin) const {
    auto int_size = int(objects.size());
    auto vec = objects[random_int(0, int_size-1)]->random(origin);

// rest
```
but i think it would not produce any particularly visible artifact.</div>

hollasch:
Resolves #1681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Change Log / Ray Tracing in One Weekend
 
 ### Common
   - Fix    -- Fixed some dangling references to `random_in_unit_sphere()` (#1637)
+  - Fix    -- Clarify `uniform_real_distribution` usage for `random_double()` (#1680)
 
 ### In One Weekend
   - Fix    -- Fix equation for refracted rays of non-unit length (#1644)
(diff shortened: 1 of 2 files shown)


hollasch:
@niccolot

hollasch:
Note: this addresses #1681, not #1680 as originally stated.

dimitry-ishenko (changes requested):
@hollasch as gone over in #1681 the functions should be marked as `inline` to avoid problems with multiple inclusions.

Issue #1482, opened 2024-03-29
[Book 2, Chapter 8.2] Use range-based for-loop


dimitry-ishenko:
In Listing 70, instead of using math tricks to permute between min/max members of `bbox`, use range-based for-loop that was introduced in C++11. This improves readability and allows people to understand what is actually being done in the loop.

```diff
         point3 min( infinity,  infinity,  infinity);
         point3 max(-infinity, -infinity, -infinity);
 
-        for (int i = 0; i < 2; i++) {
-            for (int j = 0; j < 2; j++) {
-                for (int k = 0; k < 2; k++) {
-                    auto x = i*bbox.x.max + (1-i)*bbox.x.min;
-                    auto y = j*bbox.y.max + (1-j)*bbox.y.min;
-                    auto z = k*bbox.z.max + (1-k)*bbox.z.min;
-
-                    auto newx =  cos_theta*x + sin_theta*z;
-                    auto newz = -sin_theta*x + cos_theta*z;
-
-                    vec3 tester(newx, y, newz);
+        for (auto x : { bbox.x.min, bbox.x.max }) {
+            for (auto y : { bbox.y.min, bbox.y.max }) {
+                for (auto z : { bbox.z.min, bbox.z.max }) {
+
+                    vec3 tester {
+                         cos_theta*x + sin_theta*z,
+                         y,
+                        -sin_theta*x + cos_theta*z
+                    };
 
                     for (int c = 0; c < 3; c++) {
                         min[c] = fmin(min[c], tester[c]);```

hollasch:
Thanks Dimitry.

hollasch:
Ref #819

hollasch:
Note the echo of this approach in `perlin::perlin_interp()`.

PR #1696, opened 2025-04-02
Optimize the aabb::rotate_y function.

### LINKED ISSUE #1640: Book 2 8.2 `rotate_y` not necessary to loop through y-value
In `rotate_y`'s constructor, you loop through all three axes to get the maximal bounding box containing the rotated object, but since the rotation is around the y-axis, the y-interval will not change. So looping through y-values can be omitted.

WiHceeg:
y-interval will not change, the triple loop can be optimized into a double loop.
Resolves #1640

### DIFF
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -3866,29 +3866,23 @@ earlier.
             double sin_theta = std::sin(radians);
             double cos_theta = std::cos(radians);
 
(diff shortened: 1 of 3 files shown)


hollasch:
See #1482, where this is a component of a larger refactoring already underway.

hollasch:
@WiHceeg —
Just a note that I'm going to revisit this change, and might just take it or adopt it for the planned work. I'll preserve your commit if possible.

Issue #1622, opened 2024-08-25
Book3 -> Importance Sampling -> error ICD function


liudeyuan2021:
* error ICD:
![Screenshot from 2024-08-25 11-43-09](https://github.com/user-attachments/assets/de869d5b-8ab7-4e0f-b344-3d277a95e2c4)

* correct ICD:
![Screenshot from 2024-08-25 11-43-51](https://github.com/user-attachments/assets/895f4818-3845-484f-99ca-7e78a4bb7771)

PR #1705, opened 2025-05-11
Fix ICD formula for solved PDF in Importance Sampling (#1622)

spayne:
1. This pull request fixes the ICD math.  What was published was $8 d^\frac{1}{3}$.  What we want is: $2 d^\frac{1}{3}$.
2. Note that when generating the estimate, the ICD gets cancelled out anyway.  The code ```sum += x*x/pdf(x)``` is $\sum \frac{icd^2}{pdf(icd)} = \sum \frac{icd^2}{\frac{3}{8}icd^2} = \frac{8}{3}$.  The text doesn't mention this cancellation but does bring up the ICD again when talking about the PDF.   I think it's wiser to stick to the PDF so I updated this reference as well.
4. I also updated CHANGELOG.md and the integrate_x_sq.cc


spayne:
Requesting review from @RayTracing/reviewers

PR #1734, opened 2026-03-01
Simplify box() face construction

### LINKED ISSUE #1733: Book 2.8: Optional simplification of box() face construction
The ` box()`  helper constructs the six faces of the box by explicitly recombining coordinates, for example:

`point3(min.x(), min.y(), max.z())`

However, since the function already computes the axis-aligned edge vectors dx, dy, and dz, the six faces can alternatively be constructed symmetrically using only min, max, and these edge vectors:

```
sides->add(make_shared<quad>(min, dx, dy, mat)); // back
sides->add(make_shared<quad>(min, dx, dz, mat)); // bottom
sides->add(make_shared<quad>(min, dy, dz, mat)); // left
sides->add(make_shared<quad>(max, -dx, -dy, mat)); // front
sides->add(make_shared<quad>(max, -dx, -dz, mat)); // top
sides->add(make_shared<quad>(max, -dy, -dz, mat)); // right
```

xiaodoliu:
Resolves #1733

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Change Log / Ray Tracing in One Weekend
 
 ### The Next Week
   - Fix    -- Remove premature source line for call to `get_sphere_uv` (#1701)
+  - Change -- Simplify `box()` face construction to use `min` and `max` points directly
+              without recombining coordinates (#1733)
 
 ### The Rest of Your Life
 
--- a/books/RayTracingTheNextWeek.html
+++ b/books/RayTracingTheNextWeek.html
@@ -3520,12 +3520,12 @@ create a function that returns a box, by creating a `hittable_list` of six recta
         auto dy = vec3(0, max.y() - min.y(), 0);
         auto dz = vec3(0, 0, max.z() - min.z());
 
-        sides->add(make_shared<quad>(point3(min.x(), min.y(), max.z()),  dx,  dy, mat)); // front
-        sides->add(make_shared<quad>(point3(max.x(), min.y(), max.z()), -dz,  dy, mat)); // right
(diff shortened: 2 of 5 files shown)


hollasch (changes requested):
Just realized that there's no need for the vector negations. My mental model was to "anchor" the cross product for the normal vector to the given point (`min` or `max`). But cross products aren't "from" any point at all, so we can just choose the simplest cross product that yields the face normal. Put even more simply: ${-\vec{A}} \times {-\vec{B}} = \vec{A} \times \vec{B}$.

hollasch:
Note that I'm traveling and may not get back to this until March 12.

xiaodoliu:
> Just realized that there's no need for the vector negations. My mental model was to "anchor" the cross product for the normal vector to the given point (`min` or `max`). But cross products aren't "from" any point at all, so we can just choose the simplest cross product that yields the face normal. Put even more simply: − A → × − B →= A → × B → .



> Just realized that there's no need for the vector negations. My mental model was to "anchor" the cross product for the normal vector to the given point (`min` or `max`). But cross products aren't "from" any point at all, so we can just choose the simplest cross product that yields the face normal. Put even more simply: − A → × − B →= A → × B → .

Although the face normal is the same after flipping both vectors, it changes the vertex positions.

For example, quad(Q, −u, −v) gives vertices (Q, Q−u, Q−v, Q−u−v), while quad(Q, u, v) gives (Q, Q+u, Q+v, Q+u+v). The quad is no longer in the same position unless the anchor is offset to Q−u−v (i.e., quad(Q−u−v, u, v)).

So when constructing faces from the `max` point, we still need the vector negations.

hollasch (approved):

hollasch:
Copilot added some interesting feedback with regard to the texture UV orientation changes. Ultimately, we don't really have an example that requires honoring the original face orientation, so I think we'll just ignore this for now.

